PR #31, opened 2022-10-26
fix input-group padding when there is no addon

monK87:
**Changes**
fix input-group padding when there is no addon given (it's optional)

- delete branch

### DIFF
--- a/src/input-group/input-group.style.tsx
+++ b/src/input-group/input-group.style.tsx
@@ -89,7 +89,7 @@ export const MainInput = styled(Input)<MainInputStyleProps>`
         box-shadow: none;
     }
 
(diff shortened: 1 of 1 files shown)


src/input-group/input-group.style.tsx
```diff
@@ -81,28 +81,35 @@ export const MainInput = styled(Input)<MainInputStyleProps>`
...
     border: none;
+
     :focus,
     :active {
         outline: none;
         border: none;
         box-shadow: none;
     }
 
+    padding: 0.2rem 1rem 0.3rem 1rem; // Input text appears lower hence the higher lower padding
```
keithtxw, line 92: We can use `padding:  0.2rem 1rem 0.3rem;` for this. instead

keithtxw (approved):

PR #35, opened 2022-11-03
change the font size of input subtitle

zxian97:
**change the font size of input subtitle**
change the font size of input subtitle

- delete branch


src/form/form-label.style.tsx
```diff
@@ -38,4 +38,5 @@ export const ErrorMessage = styled(Text.H6)`
 export const Subtitle = styled(Text.XSmall)<LabelProps>`
```
keithtxw, line 38: why dont use `Text.BodySmall` then? that is 1rem. Unless u want to keep the line height and letter spacing of `XSmall`?

keithtxw (changes requested):
To clarify

keithtxw (approved):

PR #43, opened 2022-11-15
Improve breadcrumb fade specification

keithtxw:
**Changes**
Improve fade specification by allowing user to specify the left and right colour set. Currently only allow specifying a single colour set that applies to both sides. This is helpful in cases where the background is not a single colour.

- [delete] branch

**Changelog entry**

- Allow specifying different fade colors for `Breadcrumb`

**Additional information**

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/MOL-11243)

### DIFF
--- a/src/breadcrumb/breadcrumb.tsx
+++ b/src/breadcrumb/breadcrumb.tsx
@@ -10,7 +10,7 @@ import {
     PreviousLink,
     Wrapper,
 } from "./breadcrumb.style";
-import { BreadcrumbProps } from "./types";
+import { BreadcrumbProps, FadeColorSet } from "./types";
 
 export const Breadcrumb = ({
     links,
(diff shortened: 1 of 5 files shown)


src/breadcrumb/breadcrumb.tsx
```diff
@@ -137,13 +137,36 @@ export const Breadcrumb = ({
     };
 
     const renderFade = () => {
+        let fadeColorSet: FadeColorSet;
+
+        if (Array.isArray(fadeColor) && fadeColor.length > 0) {
```
qroll, line 142: if the check is nested like this, then there's no need to cast FadeColorSet afterwards
```
         if (Array.isArray(fadeColor)) {
            if (fadeColor.length) {
               // ...
            }
```

src/breadcrumb/types.ts
```diff
@@ -1,8 +1,13 @@
...
 
+export interface FadeColorSet {
+    left?: string[] | undefined;
+    right?: string[] | undefined;
+}
+
 export interface BreadcrumbProps {
     links: React.AnchorHTMLAttributes<HTMLAnchorElement>[];
-    fadeColor?: string[] | undefined;
+    fadeColor?: string[] | FadeColorSet | undefined;
```
qroll, line 10: alternatively we can make the typing "enforce" at least one element `[string, ...string[]]`

src/breadcrumb/breadcrumb.tsx
keithtxw: Haha nested ifs are ugly. :P

src/breadcrumb/types.ts
keithtxw: You mean to counter not having to check for length?

PR #44, opened 2022-11-16
Add unit number input

zxian97:
**Changes**
add unit number input. The component is referencing on the `date-input component`

- delete branch

**Changelog entry**

- [WARNING] had make some change in the style of `date-input component`

**Additional information**

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/MOL-10991)


src/unit-number/unit-number-input.tsx
```diff
@@ -0,0 +1,390 @@
...
+        const targetName = event.target.name as FieldType;
+
+        /**
+         * NOTE: Unfortunately, the maxLength is not respected for
+         * input type=number. Will need to add safeguards.
+         * Reference: https://stackoverflow.com/a/18510925
+         *
+         * Ignore inputs that exceed the respective max length
+         */
+
```
qroll, line 213: is this logic still needed? noticed that input types are now `text`

qroll (changes requested):

src/unit-number/unit-number-input.tsx
```diff
@@ -0,0 +1,390 @@
...
+            "Backspace",
+            "Delete",
+            "ArrowLeft",
+            "ArrowRight",
+        ];
+
+        // form validation for inputs is not supported on ios,
+        // hence manual checks needed to prevent special characters input
+        const isPermittedCharacter =
+            /[0-9A-Za-z]/.test(event.key) ||
```
qroll, line 149: was able to enter special characters on Android browser, event.key was `Unidentified`

src/unit-number/unit-number-input.tsx
```diff
@@ -211,18 +164,17 @@ export const UnitNumberInput = ({
          * Ignore inputs that exceed the respective max length
          */
 
-        const value = event.target.value.toLocaleUpperCase();
+        let value = event.target.value.toLocaleUpperCase();
+        if (!/.*[0-9A-Za-z]$/.test(value)) {
+            value = value.replace(/.$/, "");
```
qroll, line 169: this regex will allow special characters at the front

for alphanumeric checks and replacement, try
```suggestion
       if (/[^0-9A-Za-z]/.test(value)) {
            value = value.replace(/[^0-9A-Za-z]/, "");
```

src/unit-number/unit-number-input.tsx
qroll: can clean up this comment if the length check logic is no longer required 🙂

qroll (changes requested):
looks good functionally, just need to update regex and resolve other minor code style issues

src/unit-number/unit-number-input.tsx
```diff
@@ -0,0 +1,341 @@
...
+    UnitInput,
+    UnitNumberDivider,
+} from "./unit-number-input.style";
+import { AddOnContainer } from "../input-group/input-group.style";
+import { DateInputProps } from "../date-input/types";
+
+type FieldType = "floor" | "unit" | "none";
+type ValueFieldTypes = Exclude<FieldType, "none">;
+
+export interface UnitNumberInputProps extends DateInputProps {
```
qroll, line 16: would prefer not to extend from `DateInputProps`, especially since those properties have date input specific comments

src/unit-number/unit-number-input.tsx
```diff
@@ -0,0 +1,341 @@
...
+    };
+
+    const performOnBlurHandler = () => {
+        if (onBlur) {
+            const values: Record<ValueFieldTypes, string> = {
+                floor: floorValueStateRef.current,
+                unit: unitValueStateRef.current,
+            };
+
+            const returnValue = getFormattedValue(values);
```
qroll, line 238: sometimes returns `1-01` or `01-1` instead of the expected `01-01`, do we need to run `formatInput` here?

to reproduce, type floor=1, unit=1, then click outside of the component

qroll (approved):

PR #47, opened 2022-11-21
[MOL-11592][EK] Extracted the Bubble used in Popover as a stand-alone component.

eugkong:
**Changes**
Extracted the Bubble used in Popover as a stand-alone component, so that it can be reused.
[MOL-11592 requires the addition of a bubble (similar to the one in Popover) in the map modal (location picker) in mol-frontend-engine.]

- delete branch

**Additional information**
Things to consider:
- What should be the default styling (padding, font size, etc.)? Currently following the bubble for the map modal.
- Should the component have a different name (e.g. Card)?
- Any additional functionality needed for this Bubble component?
- Will add a commit to increase the patch version and update the changelog after all changes are finalized.

### DIFF
--- a/src/card/card.style.tsx
+++ b/src/card/card.style.tsx
@@ -1,0 +1,12 @@
+import { Color } from "src/color";
+import styled from "styled-components";
+
+// =============================================================================
+// STYLING
+// =============================================================================
+export const StyledCard = styled.div`
+    border-radius: 8px;
+    background: ${Color.Neutral[8]};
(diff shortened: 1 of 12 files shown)


qroll:
- renaming to Card is fine, Bubble doesn’t make sense outside of the Popover context
- having a reasonable default padding is fine imo (realistically we’ll end up constantly overriding it, because most components have specific margin/padding requirements)
- should keep to the default text style

qroll (approved):

PR #50, opened 2022-11-23
[MOL-11718][CCW|SW] add key and export type

ryhuz:
**Changes**
export smart app banner Content type
add key to star img components

not an urgent change. can go out with the next release

### DIFF
--- a/src/smart-app-banner/smart-app-banner.tsx
+++ b/src/smart-app-banner/smart-app-banner.tsx
@@ -36,13 +36,13 @@ function SmartAppBannerComponent(
     const {
         className,
         show,
-        link,
+        href,
         content,
         offset = 0,
         icon = APP_ICON,
-        isAnimated = false,
-        onBannerDismiss,
-        onBannerPress,
+        animated = false,
+        onDismiss,
+        onClick,
     } = props;
 
     const { title, message, buttonLabel, buttonAriaLabel, numberOfStars } =
@@ -55,9 +55,9 @@ function SmartAppBannerComponent(
         e: React.MouseEvent<HTMLDivElement | HTMLButtonElement>
     ) => {
         e.stopPropagation();
-        window.open(link, "_blank", "noreferrer");
-        onBannerDismiss();
-        onBannerPress?.();
+        window.open(href, "_blank", "noreferrer");
+        onDismiss();
+        onClick?.();
     };
 
     // =============================================================================
@@ -96,12 +96,12 @@ function SmartAppBannerComponent(
             {show && (
                 <SmartAppBannerContainer
                     ref={ref}
-                    $isAnimated={isAnimated}
+                    $isAnimated={animated}
                     $offset={offset}
                     className={className}
                 >
                     <DismissContainer
-                        onClick={onBannerDismiss}
+                        onClick={onDismiss}
                         id={`${ID}-dismiss`}
                         data-testid={`${ID}-dismiss-container`}
                     >
--- a/src/smart-app-banner/smart-app-banner.tsx
+++ b/src/smart-app-banner/smart-app-banner.tsx
@@ -12,24 +12,7 @@ import {
     TextContainer,
     Title,
 } from "./smart-app-banner.styles";
-
-interface SmartAppBannerProps {
-    show: boolean;
-    link: string;
-    content: SmartAppBannerContent;
-    offset?: number | undefined;
-    className?: string | undefined;
-    onBannerDismiss: () => void;
-    onBannerPress?: () => void | undefined;
-}
-
-export interface SmartAppBannerContent {
-    title: string;
-    buttonLabel: string;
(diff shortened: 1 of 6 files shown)


src/smart-app-banner/smart-app-banner.tsx
```diff
@@ -23,7 +23,7 @@ interface SmartAppBannerProps {
     onBannerPress?: () => void | undefined;
 }
 
-interface Content {
+export interface Content {
```
qroll, line 26: suggest renaming to something like `SmartAppBannerContent` to avoid naming collision

qroll:
can help to update typing for `numberOfStars` to be `numberOfStars?: number | undefined;`?

src/smart-app-banner/smart-app-banner.tsx
```diff
@@ -16,14 +16,14 @@ import {
 interface SmartAppBannerProps {
```
keithtxw, line 16: These interfaces should be in the `types.ts` file instead.

keithtxw: File is also missing all the necessary headers to demarcate the event handlers, helper functions and render functions

keithtxw (changes requested):
- Interfaces to be in types.ts file
- File is missing necessary header as part of working agreements in `CONTRIBUTING.md`

src/smart-app-banner/smart-app-banner.tsx
keithtxw: To rename as `SmartAppBannerContentProps`

src/smart-app-banner/types.ts
```diff
@@ -0,0 +1,17 @@
+export interface SmartAppBannerProps {
+    show: boolean;
+    link: string;
+    content: SmartAppBannerContentProps;
+    offset?: number | undefined;
+    className?: string | undefined;
+    onBannerDismiss: () => void;
```
keithtxw, line 7: actually a question, is this required to be mandatory?

src/smart-app-banner/types.ts
```diff
@@ -0,0 +1,17 @@
...
+    onBannerPress?: () => void | undefined;
+}
+
+export interface SmartAppBannerContentProps {
+    title: string;
+    buttonLabel: string;
+    buttonAriaLabel: string;
```
keithtxw, line 14: Are these the only 2 props required for the button?

src/smart-app-banner/types.ts
ryhuz: yes, because there is no default set. we need to pass a callback which sets `show` to false

src/smart-app-banner/types.ts
ryhuz: yes, button will have the same behaviour as onBannerPress

keithtxw (changes requested):

stories/smart-app-banner/props-table.tsx
```diff
@@ -110,6 +110,13 @@ const DATA: ApiTableSectionProps[] = [
                     "The description of the banner displayed below the title",
                 propTypes: ["string"],
             },
```
keithtxw, line 112: to also update the `Content` prop name

src/smart-app-banner/types.ts
```diff
@@ -0,0 +1,19 @@
+export interface SmartAppBannerProps {
+    show: boolean;
+    link: string;
```
keithtxw, line 3: Maybe `href` is a better name for this since it is just the destination

keithtxw (changes requested):

src/smart-app-banner/types.ts
```diff
@@ -0,0 +1,19 @@
...
+    content: SmartAppBannerContentProps;
+    offset?: number | undefined;
+    icon?: string | undefined;
+    isAnimated?: boolean | undefined;
```
keithtxw, line 7: Can i recommend to rename to `animated`

keithtxw (changes requested):

src/smart-app-banner/types.ts
ryhuz: updated

src/smart-app-banner/types.ts
ryhuz: updated

src/smart-app-banner/types.ts
```diff
@@ -0,0 +1,19 @@
...
+    className?: string | undefined;
+    onBannerDismiss: () => void;
+    onBannerPress?: () => void | undefined;
```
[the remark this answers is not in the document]
ryhuz: updated

src/smart-app-banner/types.ts
ryhuz: updated

keithtxw (approved):

PR #52, opened 2022-11-28
Migration enhancements

qroll:
**Changes**
- add on to migration docs
- attempt to fix navbar generic getting inferred as `unknown` (form input group also affected)
- export navbar mobile height
- add `flex-container` type for layout
  - common pain point was having to specifically add `flex-direction: column` to the nested container in `Layout.Content`
-  attempt to address type error with `extractedValue` for `onSelectOption` prop
   - TS sometimes has trouble reconciling `T | V` (V cannot be assigned to T as it doesn't extend T), so just type as V instead
- delete branch

**Changelog entry**

- Enhance migration docs
- [WARNING] Augment `React.forwardRef` to allow generics. This overrides the default `ForwardRefExoticComponent` return type for DS components
- Expose `Navbar` mobile height
- Add `flex-container` type for layout
- Resolve type error with `onSelectOption` prop when `extractedValue` cannot be assigned to `option`

**Additional information**

- Open to suggestions for a better way to include the react augmented typings
- There's still a limitation with `onSelectOption`: if valueExtractor is not passed in, then extractedValue will be unknown. Have tried a couple of things including function overload but the results were unsatisfactory. Since extractedValue is not useful without valueExtractor, I think this is an acceptable tradeoff?


weili-govtech:
the augmentation looks fine to me, if you prefer to apply to `NavBar` only, you can also cast it as a function with generics

src/navbar/navbar.styles.tsx
```diff
@@ -38,7 +38,7 @@ export const Nav = styled.nav<StyleProps>`
     transition: ${Transition.Base};
 
     ${MediaQuery.MaxWidth.tablet} {
-        height: 3.5rem;
+        height: ${() => NAVBAR_MOBILE_HEIGHT}rem;
```
keithtxw, line 41: why does this need to be a function?

keithtxw (changes requested):

src/navbar/navbar.styles.tsx
qroll: hmm, this was originally meant to resolve a type error, but turns out the actual error is from using the constant before its declaration

will amend

src/shared/dropdown-list/dropdown-list.tsx
```diff
@@ -139,8 +139,8 @@ export const DropdownList = <T, V>({
     // =============================================================================
     // HELPER FUNCTIONS
     // =============================================================================
-    const getValue = (item: T): V | T => {
-        return valueExtractor ? valueExtractor(item) : item;
+    const getValue = (item: T): V => {
+        return valueExtractor ? valueExtractor(item) : (item as unknown as V);
```
keithtxw, line 143: Why do we use a double assertion here? rather than do the original?

qroll: Trying to treat the extracted value as `V` always. Since `V` could be any arbitrary type, Typescript doesn't accept `T as V`, so casting to unknown works around that

keithtxw (approved):

PR #53, opened 2022-11-29
[MOL-11381][ZX] add date input readonly style

zxian97:
**Changes**
add readonly status for date input

- delete branch

**Changelog entry**

- change for the handle key down function as it does not work in android browser. [replace this checking function at the 
   onchange handle ]
- add readonly status
- change the placeholder function

**Additional information**

- You may refer to this [MOL-11381](https://jira.ship.gov.sg/browse/MOL-11381)

### DIFF
--- a/src/date-input/date-input.tsx
+++ b/src/date-input/date-input.tsx
@@ -253,8 +253,12 @@ export const DateInput = ({
 
         if (onChangeRaw) {
             const valuesArr = [
-                ...(field === "day" ? [changeValue] : [dayValue]),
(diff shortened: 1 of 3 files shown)


src/date-input/date-input.tsx
```diff
@@ -156,64 +123,8 @@ export const DateInput = ({
     // =============================================================================
     // EVENT HANDLERS
     // =============================================================================
-    const handleKeyDown = (event: KeyboardEvent) => {
```
keithtxw, line 159: Will just checking for the format be enough?

PR #56, opened 2022-12-01
 [MOL-10991][ZX] correction keydown of unit number

zxian97:
**Changes**
add the keydown function and change on the style

- delete branch

**Additional information**

- You may refer to this [MOL-10991](https://jira.ship.gov.sg/browse/MOL-10991)

### DIFF
--- a/src/unit-number/unit-number-input.tsx
+++ b/src/unit-number/unit-number-input.tsx
@@ -154,10 +154,9 @@ export const UnitNumberInput = ({
     const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
         const targetName = event.target.name as FieldType;
 
-        let value = event.target.value.toLocaleUpperCase();
-        if (/[^0-9A-Za-z]/.test(value)) {
-            value = value.replace(/[^0-9A-Za-z]/, "");
-        }
+        const value = event.target.value
+            .toLocaleUpperCase()
+            .replace(/[^0-9A-Za-z]/, "");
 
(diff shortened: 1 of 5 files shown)


src/unit-number/unit-number-input.tsx
```diff
@@ -308,6 +323,7 @@ export const UnitNumberInput = ({
                     onFocus={handleFocus}
                     onBlur={handleBlur}
                     onChange={handleChange}
+                    onKeyDown={handleKeyDown}
                     disabled={disabled}
                     readOnly={readOnly}
                     type="text"
```
keithtxw, line 329: missing the `inputMode` otherwise the keyboard will have non-numeric input options

src/unit-number/unit-number-input.tsx
```diff
@@ -231,8 +246,8 @@ export const UnitNumberInput = ({
 
         if (onBlurRaw) {
             const valuesArr = [
-                floorValueStateRef.current,
-                unitValueStateRef.current,
+                formatInput(floorValueStateRef.current),
```
keithtxw, line 249: move `formatInput` to the correct section. It should not be with the event handlers

keithtxw (changes requested):
- Also add a section header `RENDER FUNCTIONS` in `unit-number-input.tsx` 
- I would actually prefer the folder to be `unit-number-input` rather than `unit-number`. But if u are consuming it already then never mind.

keithtxw (approved):

PR #64, opened 2022-12-13
Fix unit number input format clash

keithtxw:
**Changes**
Fix unintended formatting of input value as the user types

Also:

- Standardise the return values for `onChange` and `onChangeRaw` to be unpadded for single digit values
- Also improve some stories' documentation

- [delete] branch

**Changelog entry**

- `DateInput`
     - Fix unintended padding of single digit values
     - Standardise the return values for `onChange` and `onChangeRaw` to be unpadded for single digit values

### DIFF
--- a/src/unit-number/unit-number-input.tsx
+++ b/src/unit-number/unit-number-input.tsx
@@ -22,7 +22,7 @@ export const UnitNumberInput = ({
     onChangeRaw,
     onBlurRaw,
     readOnly,
-    placeholder,
+    placeholder = "00-8888",
     ...otherProps
 }: UnitNumberInputProps) => {
     // =============================================================================
@@ -191,11 +191,12 @@ export const UnitNumberInput = ({
     const updateValues = (value: string | undefined) => {
         if (value === INVALID_VALUE) {
             return;
-        }
-
-        if (!(value === undefined || value.length === 0)) {
+        } else if (value === undefined || value.length === 0) {
+            setFloorValue("");
+            setUnitValue("");
+        } else {
             const valueArr = value.split("-");
-            if (valueArr.length !== 0) {
+            if (valueArr.length === 2) {
                 // Valid value
                 const floor = valueArr[0];
                 const unit = valueArr[1];
@@ -206,9 +207,6 @@ export const UnitNumberInput = ({
                     currentFocus === "unit" ? unit : formatInput(unit)
                 );
             }
-        } else {
-            setFloorValue("");
-            setUnitValue("");
         }
     };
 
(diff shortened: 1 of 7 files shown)


src/unit-number/unit-number-input.tsx
```diff
@@ -99,7 +99,7 @@ export const UnitNumberInput = ({
     }, [floorValue]);
 
     useEffect(() => {
-        formatDisplayValues(value);
+        updateValues(value);
```
keithtxw, line 102: Rename for clarity

src/unit-number/unit-number-input.tsx
```diff
@@ -188,69 +188,70 @@ export const UnitNumberInput = ({
...
-                ...(field === "floor"
-                    ? [formatInput(changeValue)]
-                    : [floorValue]),
-                ...(field === "unit"
-                    ? [formatInput(changeValue)]
-                    : [unitValue]),
-            ];
-
-            onChangeRaw(valuesArr);
+            onChangeRaw([values.floor, values.unit]);
```
keithtxw, line 234: We'll not format the values onChange unless the input has been blurred

stories/form/form-unit-number-input/props-table.tsx
```diff
@@ -43,6 +43,11 @@ const DATA: ApiTableSectionProps[] = [
                 description: "The unique identifier of the component",
                 propTypes: ["string"],
             },
+            {
+                name: "placeholder",
+                description: "The placeholder of the component",
```
qroll, line 48: should we keep the original note on the placeholder format here? passing in any other string that doesn't match `<floor>-<unit>` format may have unexpected results

src/unit-number/unit-number-input.tsx
qroll: just want to clarify: does input here mean the individual floor/unit input elements? not the entire component?

so this would be the expected behaviour?
1. type 1 in `floor` -> `onChange("Invalid Unit Number")`
2. switch to unit input -> floor gets padded visually, but no change/blur handlers triggered
3. type 1 in `unit` -> `onChange("01-1")`

qroll (changes requested):

src/unit-number/unit-number-input.tsx
keithtxw: Yes the input I am referring to is the entire component. 

And yes the behaviour you listed is expected.

stories/form/form-unit-number-input/props-table.tsx
keithtxw: It's meant to be free and easy. But i'll add a recommendation instead

src/unit-number/unit-number-input.tsx
```diff
@@ -188,69 +188,70 @@ export const UnitNumberInput = ({
...
-        if (value === undefined || value === "") {
-            setFloorValue("");
-            setUnitValue("");
-        } else {
+    const updateValues = (value: string | undefined) => {
+        if (value === INVALID_VALUE) {
+            return;
+        }
+
+        if (!(value === undefined || value.length === 0)) {
```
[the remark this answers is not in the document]
keithtxw: the negation is still required for the initial and removal of all values. But i'll swop it around for clarity

qroll (approved):

PR #65, opened 2022-12-14
[MISC][KT] Amend max-width values to align with SGDS

keithtxw:
**Changes**
Amend max widths for various viewports as part of alignment with SGDS

- [delete] branch

**Changelog entry**

- Update `Layout.Container` max widths for various viewports as part of alignment with SGDS


Wilker0725 (approved):
LGTM

PR #68, opened 2022-12-22
[MISC][RL] Fix unintended undefined function return types

qroll:
**Changes**
current function typings allow undefined to be returned, which isn't intended

- Update typings
  - Remove undefined return type for functions
  - Add undefined for optional properties that were missed out
  - Remove some unused DateInput props
- Fix error that occurs when Footer links are not specified but `showDownloadAddon` is true
- delete branch

**Changelog entry**

- Remove undefined return type for functions in component props

### DIFF
--- a/src/date-input/types.ts
+++ b/src/date-input/types.ts
@@ -1,31 +1,15 @@
-export interface DateInputTarget {
-    value: string;
-    type?: string | undefined; // input type
-    name?: string | undefined;
-    id?: string | undefined;
-}
-
-/* This mimics the React.SyntheticEvent */
-export interface DateInputEvent {
-    bubbles: boolean;
-    target: DateInputTarget;
-}
-
 export interface DateInputProps extends React.AriaAttributes {
     // Standard HTML Attributes
     className?: string | undefined;
     id?: string | undefined;
-    style?: React.CSSProperties | undefined;
-    tabIndex?: number | undefined;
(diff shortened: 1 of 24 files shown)


src/date-input/types.ts
```diff
@@ -18,7 +18,7 @@ export interface DateInputProps extends React.AriaAttributes {
     style?: React.CSSProperties | undefined;
     tabIndex?: number | undefined;
     "data-testid"?: string | undefined;
-    readOnly?: boolean;
+    readOnly?: boolean | undefined;
```
keithtxw, line 21: I think this no longer needs to be added since we're saying we'll omit them

src/date-input/types.ts
```diff
@@ -28,16 +28,16 @@ export interface DateInputProps extends React.AriaAttributes {
     name?: string | undefined;
     error?: boolean | undefined;
     disabled?: boolean | undefined;
-    onChange?: (value: string) => void | undefined;
+    onChange?: ((value: string) => void) | undefined;
```
keithtxw, line 31: since we will omit the `undefined`, we no longer need the parenthesis

keithtxw (changes requested):
I think we'll remove all the undefined ones. I'll take some time to look through to see which one truly needs.

keithtxw (approved):

PR #72, opened 2022-12-23
[BOOKINGSG-3770][AS] Phone Number Input

afiq0209:
**Changes**
New Phone Number Input Component

- [delete] branch

**Changelog entry**

- [WARNING] Dropdown-list: Add new secondaryLabelExtractor for secondaryLabel and add clearable option for search field

**Additional information**

- You may refer to this [BOOKINGSG-3770](https://jira.ship.gov.sg/browse/BOOKINGSG-3770)


src/shared/dropdown-list/types.ts
```diff
@@ -15,6 +15,7 @@ export interface DropdownDisplayProps<T, V> {
         item: T,
         args: ListItemRenderArgs
     ) => JSX.Element | undefined;
+    secondaryLabelExtractor?: (item: T) => string | undefined;
```
keithtxw, line 18: I would prefer if we add on to the `listExtractor` instead. To be something like that
```
listExtractor?: ((item: T) => string | ListItemDisplayProps) | undefined;
```
and 

```
interface ListItemDisplayProps {
    title: string;
    secondaryLabel?: string | undefined
}
```

keithtxw: then in our Dropdown list we'll check, if it is a string, it means the user just wants title. Then we'll render the string into the title. But if it is an object, we'll render the title and if secondaryLabel is specified, we render it as well. 

Anyway I realised those functions typings is lacking a parenthesis. Might need you to help me add for this file. Because e.g.
`valueExtractor?: (item: T) => V | undefined;`
it actually meant can pass in a function or pass undefined. But if specify like that, it means, must pass function that has return type of either `V` or `undefined`

So it should be 
`valueExtractor?: ((item: T) => V) | undefined;`

src/shared/dropdown-list/dropdown-list.tsx
```diff
@@ -307,6 +336,11 @@ export const DropdownList = <T, V>({
                                     hasExceededContainer(item)
                                         ? renderTruncatedText(item)
                                         : getOptionLabel(item)}
+                                    {secondaryLabelExtractor && (
+                                        <div className="secondary-label">
```
keithtxw, line 340: Should use a span instead of div. And also to create a styled component instead of using `className`

src/shared/dropdown-list/dropdown-list.styles.tsx
```diff
@@ -125,6 +125,11 @@ export const Label = styled.div<LabelProps>`
     text-align: left;
     line-height: 1.375rem;
     overflow: hidden;
+    .secondary-label {
```
keithtxw, line 128: create as a styled component instead of doing this

keithtxw (changes requested):

src/phone-number-input/phone-number-input-helper.tsx
```diff
@@ -0,0 +1,85 @@
...
+    if (!predefinedMask) {
+        return prefix + "".padEnd(countryCode.length, ".") + " " + defaultMask;
+    } else {
+        return (
+            prefix + "".padEnd(countryCode.length, ".") + " " + predefinedMask
+        );
+    }
+};
+
+export const getCountries = [].concat(
```
keithtxw, line 26: can i recommend using a namespace export format

```
export namespace PhoneNumberInputHelper {
   export const getCountries { ... };
   export const formatNumber { ... };
}
```

src/phone-number-input/data/country-code.ts
```diff
@@ -0,0 +1,262 @@
...
+// =============================================================================
+// Regions:
+// ['america', 'europe', 'asia', 'oceania', 'africa']
+// =============================================================================
+// Sub-regions:
+// ['north-america', 'south-america', 'central-america', 'carribean',
+//  'eu-union', 'ex-ussr', 'ex-yugos', 'baltic', 'middle-east', 'north-africa']
+// =============================================================================
+
+const countries = [
```
keithtxw, line 19: We try to avoid default exports. So exporting it as `CountryData` might be favourable

src/phone-number-input/phone-number-input.styles.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { TextStyle } from "../text/text-style";
+import { Transition } from "../transition";
+
+interface StyleProps {
+    disabled?: boolean;
+    $expanded?: boolean;
+}
+
+// =============================================================================
+// STYLING
```
keithtxw, line 18: These styles look exactly the same as the `input-group-list-addon.style.tsx`. May i suggest moving that to `shared` instead and reuse them.

keithtxw: alternatively, use from input group

src/phone-number-input/phone-number-input.tsx
```diff
@@ -0,0 +1,255 @@
+import React, { useEffect, useRef, useState } from "react";
```
keithtxw, line 1: Would it be possible to create a wrapper that uses the `InputGroup` instead?

keithtxw (changes requested):

src/phone-number-input/phone-number-input.tsx
afiq0209: I think it's possible but will need to use or enhance the addon list input group. but that introduce confusion as that for more general usage & this phone number specific usage. might need proper discussion on the implementation and proper testing as that might impact existing usage. Could this be part of future improvement?

keithtxw: Curious what kind of customisations do you foresee? I thought the existing stuff would have sufficed. 

My idea was to use the `InputGroup` wholesale and for your case, pass the:
- calling codes data as prop of `options`, and customising the `listExtractor` to format the options into the `ListItemDisplayProps` shape for the display
- Then for the value padding, you can just use the `value` prop from the `InputGroup`
- The `clearable` feature can be added onto the `Input` component directly. I think it will definitely benefit other users. Can you also rename this to `allowClear` instead

Let me know if i missed anything or over simplified. We can have a discussion

src/shared/dropdown-list/dropdown-list.tsx
```diff
@@ -252,6 +268,12 @@ export const DropdownList = <T, V>({
         if (onSearch) onSearch();
     };
 
+    const handleOnClear = () => {
```
keithtxw, line 271: build feature to `Input`

src/phone-number-input/types.ts
```diff
@@ -0,0 +1,9 @@
+import { FormInputProps } from "src/form/types";
+import { InputProps } from "../input/types";
+import { ListAddon } from "src/input-group";
+
+export interface PhoneNumberInputProps<T, V>
+    extends InputProps,
+        FormInputProps {
+    countrySelector?: ListAddon<T, V>;
```
keithtxw, line 8: I might have forgotten to mention previously. I believe this component is meant to provide an exhaustive list required. I think we need not expose the ListAddon props. I also realise it might not be a good idea to plainly inherit the InputProps. As such here are my proposal for the props:

```
export interface PhoneNumberInputValue {
    country?: string | undefined;
    number?: string | undefined;
}
export interface PhoneNumberInputProps
    extends React.HTMLAttributes<HTMLInputElement> {
    value?: PhoneNumberInputValue | undefined;
    onChange?: (value: PhoneNumberInputValue) => void | undefined;
    error?: boolean | undefined;
    /** Allows clearing of input. Specifying true will render a cross icon button */
    allowClear?: boolean | undefined;
    onClear?: () => void | undefined;

    // Dropdown option props
    optionSearchPlaceholder?: string | undefined;
    /** Specifying will render a search bar in the dropdown */
    enableSearch?: boolean | undefined;
    /** Called when the country code options is dismissed */
    onHideOptions?: () => void | undefined;
     /** Called when the country code options is shown */
    onShowOptions?: () => void | undefined;

    // Input props
    disabled?: boolean | undefined;
    name?: string | undefined;
    pattern?: string | undefined;
    placeholder?: string | undefined;
    readOnly?: boolean | undefined;
}
```

Your valueExtractor and listExtractor and all should reside within this component instead of having to pass from outside.

keithtxw: The custom `onChange` will be fired whenever the user enters/clear something in the input or selects an option in the dropdown

keithtxw (changes requested):

src/phone-number-input/types.ts
keithtxw: I also feel can just standardise the display to all have the secondary label.

src/phone-number-input/phone-number-input-helper.tsx
```diff
@@ -0,0 +1,87 @@
+import { CountryData } from "./data/country-code";
```
keithtxw, line 1: Can i get you to write unit test for this helper file

keithtxw (changes requested):

keithtxw (approved):

PR #73, opened 2023-01-13
[CHECKBOX][JL] Shift tabbable state to parent container

jeraldlyh:
**Changes**
- Change aria-state to parent div container instead of child input element
- delete branch

### DIFF
--- a/src/checkbox/checkbox.tsx
+++ b/src/checkbox/checkbox.tsx
@@ -29,12 +29,21 @@ export const Checkbox = ({
             | React.KeyboardEvent<HTMLDivElement>
     ) => {
         if (!disabled) {
+            const keyboardEvent =
+                event as React.KeyboardEvent<HTMLInputElement>;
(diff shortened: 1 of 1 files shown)


src/checkbox/checkbox.tsx
```diff
@@ -49,13 +49,19 @@ export const Checkbox = ({
...
+            tabIndex={disabled ? -1 : 0}
         >
             <Input
-                type="checkbox"
+                id="checkbox-input"
                 data-testid="checkbox-input"
+                aria-hidden="true"
+                type="checkbox"
+                tabIndex={-1}
                 onChange={handleOnCheck}
```
keithtxw, line 63: you will need special keypress handling since the tab focus is now no longer on the actual input. I am unable to select the input using the keyboard.

keithtxw (changes requested):

keithtxw (approved):

PR #74, opened 2023-01-13
[MOL-11383][ZX] add readonly style for inputselect

zxianteo:
**Changes**
add readonly style for input select (dropdown)

- delete branch

**Additional information**

- You may refer to this [MOL-11383](https://jira.ship.gov.sg/browse/MOL-11383)

### DIFF
--- a/src/input-select/input-select.tsx
+++ b/src/input-select/input-select.tsx
@@ -61,7 +61,7 @@ export const InputSelect = <T, V>({
     const handleSelectorClick = (event: React.MouseEvent) => {
         event.preventDefault();
 
-        if (!disabled) {
+        if (!disabled && !otherProps.readOnly) {
             setShowOptions(!showOptions);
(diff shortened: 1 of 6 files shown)


src/input-select/input-select.tsx
```diff
@@ -61,7 +61,7 @@ export const InputSelect = <T, V>({
     const handleSelectorClick = (event: React.MouseEvent) => {
         event.preventDefault();
 
-        if (!disabled) {
+        if (!disabled && !otherProps.readOnly) {
```
keithtxw, line 64: I somehow find this condition abit awkward. There is no issue and it works but maybe we can be abit more explicit by reversing the condition
```
if (disabled || otherProps.readOnly) {
    return;
}
```

keithtxw (changes requested):
Minor change to reverse the condition

keithtxw (approved):

PR #84, opened 2023-02-03
Input date calendar

Wilker0725:
**Changes**
Description of changes...

- delete branch

**Changelog entry**

- [BREAKING | WARNING] Description of change in `DatePicker Calendar`

**Additional information**

- You may refer to this [ticket](https://www.figma.com/file/us90jOpWBahsqK2VAluPvD/BookingSG-Design-System?node-id=3360%3A23096&t=KL4vwV8Fmg0hKXlF-0)


src/date-input/date-input.tsx
```diff
@@ -24,21 +27,32 @@ export const DateInput = ({
...
-}: DateInputProps) => {
+}: DateInputProps & DateInputRangeProps) => {
     // =============================================================================
     // CONST, STATE, REF
     // =============================================================================
     const [dayValue, _setDayValue] = useState<string>("");
     const [monthValue, _setMonthValue] = useState<string>("");
     const [yearValue, _setYearValue] = useState<string>("");
     const [currentFocus, _setCurrentFocus] = useState<FieldType>("none");
+    const [calendarManualInput, setCalendarManualInput] = useState<string>("");
```
keithtxw, line 49: I have been looking through and find that using a string requires splitting and checking, which is laborious. Also curious as to the issue of assigning to the existing `dayValue` etc states?

Is it that any selection of within the calendar is not confirmed until you press ok? If so, keeping it as an object/separate variables might be easier

src/date-input/date-input.tsx
```diff
@@ -369,6 +680,18 @@ export const DateInput = ({
                     maxLength={2}
                     value={monthValue}
                     ref={monthInputRef}
+                    {...(hoverValue !== undefined && {
```
keithtxw, line 683: i find this hard to understand. Could you consider a helper function to do this instead? Then assign an undefined to the prop rather than conditionally adding.

src/date-picker/types.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export type DatePickerType = "start" | "range";
+
+export type InputContainerType = DatePickerType | "none";
+
+export type YearMonthTransition = TransitionValue | object;
+
+export type VariantDay =
+    | "default"
+    | "nextMonth"
```
keithtxw, line 11: convention is kebab-case, so it'll be `next-month`

src/date-picker/types.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export interface DatePickerProps extends React.AriaAttributes {
+    "data-testid"?: string;
+    id?: string | undefined;
+    disabled?: boolean;
+
+    // Input-specific Attributes
+    error?: boolean;
+    readOnly?: boolean;
+    disabledDates?: string[]; // [YYYY-MM-DD]
```
keithtxw, line 50: add inline documentation instead and some brief explanation would be helpful
```
/** specifies the list of disabled dates in the YYYY-MM-DD format */
```

src/date-picker/types.ts
```diff
@@ -0,0 +1,86 @@
...
+    rangeValue?: string; // YYYY-MM-DD
+    type: DatePickerType;
+
+    onChange?: ((value: string) => void) | undefined;
+}
+
+export interface GenerateDayStatusValue {
+    variant: VariantDay;
+    dateStartWithYear: string;
+    isDisabled: boolean;
```
keithtxw, line 62: `disabled`

src/date-picker/types.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export type VariantYear =
+    | "default"
+    | "currentYear"
+    | "nextDecaded"
+    | "selectedYear";
+
+export type View = "Day" | "Month" | "Year";
+
+export type FocusToTypes = {
```
keithtxw, line 30: I think we should use `interface` for consistency

src/date-picker/date-picker.tsx
```diff
@@ -0,0 +1,1311 @@
+import React, {
```
keithtxw, line 1: This component is too big already. Can you consider pulling the calendar out instead?

src/date-picker/date-picker.tsx
```diff
@@ -0,0 +1,1311 @@
...
+        const startClasses = handleSelectStart(date);
+        const hoverClasses = handleHovering(date);
+
+        /* handle selected between date */
+        if (
+            start.length &&
+            range.length &&
+            dayjs(date).isBetween(start, range, "day", "[]")
+        ) {
+            classes.push("selected-between pre-selected-between");
```
keithtxw, line 390: I'm not comfortable about interchanging usage of class names for styled components. Would usage of props not suffice?

keithtxw (changes requested):
Perhaps we should spend some time to resolve fundamental concerns such as

- possibly extracting the calendar component and make it extensible to the desktop variant too
- type naming

Wilker0725:
Work map flow:

- [ ] Clean up the date-input.tsx (Copy from master)
- [ ] Create move main input fn to new Component StandAloneInput
- [ ] Remove all date-picker files
- [ ] Remove date-range-input files
- [ ] onChange return value as string in date-input-tsx
- [ ] StandAloneInput sharable in range input (name="start-day" || name="end-day" )
- [ ] StandAloneInput's fn sharable in range input
- [ ] Rebase from master and Import Calendar
- [ ] Styling the calendar 
- [ ] Handle the interaction for start and end `StandAloneInput` in the calendar
- [ ] Styling the start and end Calendar
- [ ] Handle the interaction for start and end `StandAloneInput` from input itself

Note to this PR
1. date-input is the one handle final onChange to user.
2. date-input receive value from `StandAloneInput`  and `Calendar` Component

Wilker0725:
Updated:
1. Uppercase the `INVALID_VALUE= "Invalid Date"` from `Invalid date`.
2. Add `performOnBlurHandler` and `performOnChangeHandler` inside `date-input-tsx`. return value {start: "DD-MM-YYYY"} or `{ start: "Invalid Date"}`.
3. Updated those `onChange`/`onChangeRaw`/`onBlur`/`onBlurRaw` => `ChangeValueTypes`

Wilker0725:
Row alignment been applied in `375px` mobile devices which suggested from Selig. He also aware our Design System not exist `374px` but that is the best user experience in those `375px` devices.

PR #91, opened 2023-02-16
Fix input style

keithtxw:
**Changes**
Style fix to `Input` and relevant components

- [delete] branch

**Changelog entry**

- Style fixes to `Input`, `InputGroup` and `PhoneNumberInput` components

### DIFF
--- a/src/input-group/input-group-list-addon.tsx
+++ b/src/input-group/input-group-list-addon.tsx
@@ -191,7 +191,7 @@ export const InputGroupListAddon = <T, V>({
                 <Selector
                     ref={selectorRef}
                     type="button"
-                    aria-readonly={readOnly}
+                    disabled={otherProps.disabled}
                     data-testid={selectorTestId || "addon-selector"}
                     onClick={handleAddonSelectorClick}
                 >
--- a/src/input-group/input-group-list-addon.tsx
+++ b/src/input-group/input-group-list-addon.tsx
@@ -8,6 +8,7 @@ import {
     LabelContainer,
     PlaceholderLabel,
     Selector,
+    SelectorReadOnly,
     StyledChevronIcon,
     ValueLabel,
     Wrapper,
@@ -19,6 +20,7 @@ export const InputGroupListAddon = <T, V>({
     addon,
     error,
     onChange,
+    readOnly,
     ...otherProps
 }: InputGroupProps<T, V>) => {
     const {
@@ -37,6 +39,8 @@ export const InputGroupListAddon = <T, V>({
         onShowOptions,
         "data-selector-testid": selectorTestId,
     } = addon.attributes as ListAddon<T, V>;
+
+    const { position } = addon;
 
     // =============================================================================
     // CONST, STATE, REF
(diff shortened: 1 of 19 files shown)


src/input-group/input-group-list-addon.tsx
```diff
@@ -173,19 +177,40 @@ export const InputGroupListAddon = <T, V>({
         </>
     );
 
+    const renderSelector = () => {
+        if (readOnly) {
+            return selected ? (
+                <SelectorReadOnly>
```
keithtxw, line 183: For the read only display, it's text only and I think we should not render a button element at all (i.e. style it to look not clickable and make it not clickable)

src/input/input.style.tsx
```diff
@@ -96,14 +97,15 @@ export const Container = styled.div<InputStyleProps>`
...
+
     // overwrite default styles
     background: transparent;
     border: none;
     height: 3rem;
     width: 100%;
 
-    ${TextStyleHelper.getTextStyle("Body", "regular")}
-    padding: 0.2rem 1rem 0.3rem; // Somehow the input text appears lower
+    padding: 0 0 0.1rem; // Somehow the input text appears lower
```
keithtxw, line 108: use the parent to set the padding/margins rather than the input to give more flexibility

src/input-group/input-group.style.tsx
```diff
@@ -143,4 +124,18 @@ export const AddOnContainer = styled.div<AddonStyleProps>`
             `;
         }
     }}
+
```
qroll, line 127: need to pass the position prop to AddOnContainer from `input-group.tsx`

src/input-group/input-group.style.tsx
```diff
@@ -80,40 +82,19 @@ export const MainInput = styled(Input)<MainInputStyleProps>`
     // overwrite default styles
     background: transparent;
     border: none;
+    padding: 0;
```
qroll, line 85: noticed that input color takes the browser default `#000`

because of that, for readonly, there's a slight mismatch with the addon text which is N1

src/input/input.tsx
```diff
@@ -24,17 +24,29 @@ const Component = (
     ref: InputRef
 ): JSX.Element => {
     // =============================================================================
-    // CONST, STATE, REFS
+    // CONST, STATE, REF
     // =============================================================================
+    const elementRef = useRef<HTMLInputElement>();
+    useImperativeHandle(ref, () => elementRef.current, []);
```
qroll, line 30: this is a cool trick, it's easier than passing in a ref callback

qroll (changes requested):

src/input-group/input-group.style.tsx
keithtxw: Good catch. Let me go rectify

src/input-group/input-group.style.tsx
keithtxw: Lol thanks for catching. I wonder how did i miss it. :/

src/unit-number/unit-number-input.tsx
```diff
@@ -272,69 +273,85 @@ export const UnitNumberInput = ({
...
+                        : getPlaceholder(placeholder)[1]
+                }
+            />
+        </>
+    );
+
+    const renderReadOnly = () => {
+        const displayValueArr = value.split("-");
+
+        return (
```
keithtxw, line 328: This is to enable selection as a whole or readonly text

src/unit-number/unit-number-input.tsx
```diff
@@ -272,69 +273,85 @@ export const UnitNumberInput = ({
...
         >
-            <AddOnContainer
+            <HashContainer
                 data-testid="addon"
                 disabled={disabled}
                 $readOnly={readOnly}
             >
                 #
-            </AddOnContainer>
-            <InputContainer $readOnly={readOnly}>
```
keithtxw, line 291: Did away with this div. Not required

qroll (changes requested):
do date input and timepicker also need to be refactored to use the shared wrapper?

src/input-select/input-select.tsx
```diff
@@ -222,6 +222,7 @@ export const InputSelect = <T, V>({
             <Selector
                 ref={selectorRef}
                 type="button"
+                aria-readonly={otherProps.readOnly}
```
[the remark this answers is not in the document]
keithtxw: i read that aria-readonly should be applied to combo box as well. Are you saying because this selector is just a button under the hood, we dont need?

Got it on the button missing disabled. Will rectify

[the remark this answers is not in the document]
keithtxw: Ok i took a closer read and it says to indicate an editable field is not editable. I guess in this case a button cannot be editable. I've removed the prop already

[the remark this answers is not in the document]
qroll: yes, aria-readonly would be applied on the element with `role=combobox`. in most implementations, that's the div/input element that displays the selected value, while button is used here.

I'm not sure what the implications are if we set `role=combobox` on `Selector`, because that would override its implicit button role. but this is out of scope, there's additional attributes required other than the role to make it compliant anyway

qroll (approved):

PR #92, opened 2023-02-20
Add fixed country code feature

keithtxw:
**Changes**
[IMPORTANT] Merge PR[#91](https://github.com/LifeSG/react-design-system/pull/91) before reviewing this

Addition of fixed country code feature for `PhoneNumberInput`

- [delete] branch

**Changelog entry**

- Add fixed country code feature for `PhoneNumberInput`

**Additional information**

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/MOL-12260)

### DIFF
--- a/src/phone-number-input/phone-number-input-helper.tsx
+++ b/src/phone-number-input/phone-number-input-helper.tsx
@@ -45,7 +45,7 @@ export namespace PhoneNumberInputHelper {
     );
 
     export const formatNumber = (
-        text: string,
+        text: string | undefined,
         country: CountryValue | undefined
     ): string => {
         if (!country) return text;
@@ -81,7 +81,7 @@ export namespace PhoneNumberInputHelper {
             },
             {
                 formattedText: "",
-                remainingText: text.split(""),
+                remainingText: text ? text.split("") : "",
             }
         );
 
--- a/src/phone-number-input/phone-number-input.tsx
+++ b/src/phone-number-input/phone-number-input.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect, useRef, useState } from "react";
+import { AddonProps, LabelAddon, ListAddon } from "../input-group/types";
 import { InputGroup } from "../input-group/input-group";
 import { PhoneNumberInputHelper } from "./phone-number-input-helper";
 import { Wrapper } from "./phone-number-input.styles";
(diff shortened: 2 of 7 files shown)


src/phone-number-input/phone-number-input.tsx
```diff
@@ -118,6 +120,40 @@ export const PhoneNumberInput = ({
         };
     };
 
+    const getAddonProps = (): AddonProps<CountryValue, string> => {
+        if (fixedCountry) {
+            return {
+                type: "label",
+                attributes: {
+                    value: selectedCountryCode,
+                } as LabelAddon,
```
qroll, line 129: no need to cast here?

src/phone-number-input/phone-number-input.tsx
```diff
@@ -118,6 +120,40 @@ export const PhoneNumberInput = ({
...
+                    searchPlaceholder: optionSearchPlaceholder,
+                    valueExtractor: (option) => `+${option.countryCode}`,
+                    listExtractor: (option) => ({
+                        title: option.name,
+                        secondaryLabel: `+${option.countryCode}`,
+                    }),
+                    onSelectOption: handleSelectOption,
+                    onHideOptions: onHideOptions,
+                    onShowOptions: onShowOptions,
+                } as ListAddon<CountryValue, string>,
```
qroll, line 149: no need to cast here?

src/phone-number-input/phone-number-input-helper.tsx
```diff
@@ -81,7 +81,7 @@ export namespace PhoneNumberInputHelper {
             },
             {
                 formattedText: "",
-                remainingText: text.split(""),
+                remainingText: text ? text.split("") : "",
```
qroll, line 84: the logic still works with string, but for correctness, it should be initialised to an empty array
```suggestion
                remainingText: text ? text.split("") : [],
```

qroll, line 84: can add a unit test for calling `formatNumber` with an empty number?

qroll (changes requested):

src/phone-number-input/phone-number-input.tsx
keithtxw: We need to else I will be able to pass list addon props

src/phone-number-input/phone-number-input.tsx
keithtxw: same as above. vice versa

PR #96, opened 2023-02-22
Add fixed country code feature

keithtxw:
**Changes**

- Add fixed country code feature
- Also refactored existing components

- [delete] branch

**Changelog entry**

- Add fixed country code for `PhoneNumberInput`

**Additional information**

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/MOL-12260)


src/input-group/input-group-list-addon.tsx
```diff
@@ -45,7 +44,7 @@ export const InputGroupListAddon = <T, V>({
     // =============================================================================
     // CONST, STATE, REF
     // =============================================================================
-    const [selected, setSelected] = useState<T>(value);
```
keithtxw, line 48: Remove erroneous use of `value` prop

qroll (changes requested):

src/phone-number-input/phone-number-input.tsx
```diff
@@ -25,26 +23,29 @@ export const PhoneNumberInput = ({
...
+        const selectedOption = options.filter(
+            (country) =>
+                country.countryCode === normaliseCountryCode(value?.countryCode)
+        )[0];
+
+        setSelectedCountry(selectedOption);
+        setInputValue(
+            PhoneNumberInputHelper.formatNumber(value?.number, selectedOption)
+        );
+    }, [options, value]);
```
qroll, line 48: continuing from above, no need to use options as a hook dependency

also compare the actual values instead of the object ref to prevent useEffect from running unnecessarily

```suggestion
    }, [value.?countryCode, value?.number]);
```

qroll (changes requested):

src/input-group/input-group.tsx
```diff
@@ -5,7 +5,7 @@ import { InputGroupListAddon } from "./input-group-list-addon";
 import { InputWrapper } from "../shared/input-wrapper/input-wrapper";
 
 const Component = <T, V>(
-    { addon, error, ...otherProps }: InputGroupProps<T, V>,
+    { addon, error, className, ...otherProps }: InputGroupProps<T, V>,
```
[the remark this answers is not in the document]
keithtxw: Ah I missed this out. Thanks for spotting

src/phone-number-input/phone-number-input.tsx
```diff
@@ -25,26 +23,29 @@ export const PhoneNumberInput = ({
     // =============================================================================
     // CONST, STATE, REF
     // =============================================================================
-    const [inputNumber, setInputNumber] = useState(value?.number || "");
-    const [selectedCountryCode, setSelectedCountryCode] = useState(
-        value?.countryCode || ""
-    );
+    const options = useMemo(() => PhoneNumberInputHelper.getCountries, []);
```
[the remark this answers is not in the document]
keithtxw: Well it is actually a function since it does a map. I can convert it to be a function to be more aligned and skip the memoisation. But I think maybe adding to the state might be better else the function will keep running upon re-render

src/phone-number-input/phone-number-input.tsx
```diff
@@ -45,7 +46,7 @@ export const PhoneNumberInput = ({
         setInputValue(
             PhoneNumberInputHelper.formatNumber(value?.number, selectedOption)
         );
-    }, [options, value]);
+    }, [value]);
```
keithtxw, line 49: i think this is sufficient since either change will trigger if you specify them separately

qroll: with `<PhoneNumberInput value={{ number, countryCode }} />` the object gets recreated on every rerender. but not a huge deal, the parent can also do its own memoization

src/phone-number-input/phone-number-input.tsx
```diff
@@ -133,7 +134,7 @@ export const PhoneNumberInput = ({
                     onSelectOption: handleSelectOption,
                     onHideOptions: onHideOptions,
                     onShowOptions: onShowOptions,
-                } as ListAddon<CountryValue, string>,
+                } as unknown as ListAddon<CountryValue, string>,
```
qroll, line 137: didn't realise `readonly` was problematic. can revert the change to avoid casting to `unknown`

keithtxw: Actually i see value in setting readonly. But ya this is the workaround. I have no comments/preference. I'll just keep it simple

qroll (approved):

PR #97, opened 2023-02-24
Style refactor

keithtxw:
**Changes**
Refactoring of components

Reasons:
- Found unnecessary layering of components that add to complexity
- Also found inconsistent use of flex and grid. On some cases, where grid was needed, it wasnt used. On some when it wasn't it was used instead. 
- Nesting styles from parent makes it hard to track the specification. Best to apply to the child directly unless it is more efficient in nesting

- [keep] branch

### DIFF
--- a/src/calendar/calendar-day.style.tsx
+++ b/src/calendar/calendar-day.style.tsx
@@ -59,16 +59,6 @@ export const GrowDayCell = styled.div<DayCellStyleProps>`
 
         if ($selected) {
             return css`
-                /* ${LeftCircle} {
-                    background-color: ${Color.Accent.Light[5]};
-                    border: 1px solid ${Color.Primary};
-                }
-
-                ${RightCircle} {
-                    background-color: ${Color.Accent.Light[5]};
-                    border: 1px solid ${Color.Primary};
-                } */
-
                 ${DayLabel} {
                     ${TextStyleHelper.getTextStyle("H5", "semibold")}
                     color: ${Color.Primary}
--- a/src/calendar/calendar-day.style.tsx
+++ b/src/calendar/calendar-day.style.tsx
@@ -1,10 +1,9 @@
 import styled, { css } from "styled-components";
 import { Color } from "../color";
-import { Text, TextStyleHelper } from "../text";
+import { Text } from "../text";
 import { DayVariant } from "./calendar-day";
 
 export interface DayCellStyleProps {
-    $selected?: boolean;
     $point?: "start" | "middle" | "end";
     $hovered?: boolean;
     $disabled?: boolean;
@@ -13,11 +12,14 @@ export interface DayCellStyleProps {
 interface DayLabelStyleProps {
     $variant: DayVariant;
     $disabled: boolean;
+    $selected: boolean;
 }
 
 interface OverflowDisplayProps {
     $position: "left" | "right";
 }
+
+interface InteractiveCircleProps extends Omit<DayLabelStyleProps, "$variant"> {}
 
 // interface DayNumberProps {
 //     $variant: DayVariant;
@@ -25,8 +27,6 @@ interface OverflowDisplayProps {
 // }
 
 // interface GrowDayCellProps extends DayCellStyleProps {}
-
-interface InteractiveCircleProps extends Omit<DayLabelStyleProps, "$variant"> {}
 
 // =============================================================================
 // STYLING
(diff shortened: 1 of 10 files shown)


src/calendar/calendar-day.style.tsx
```diff
@@ -10,46 +10,44 @@ export interface DayCellStyleProps {
...
 // =============================================================================
 // STYLING
 // =============================================================================
-
-export const WeekDaysContainer = styled.div`
-    display: flex;
-    flex-direction: row;
-    justify-content: space-around;
-    margin-bottom: 0.5rem;
+export const Wrapper = styled.div`
```
keithtxw, line 34: This component should apply the `grid` style instead as it is more appropriate

src/calendar/calendar-day.style.tsx
```diff
@@ -10,46 +10,44 @@ export interface DayCellStyleProps {
...
+    display: grid;
+    grid-template-columns: repeat(7, 1fr);
+    row-gap: 0.5rem;
 `;
 
-export const WeekCell = styled.div`
-    ${TextStyleHelper.getTextStyle("H6", "semibold")}
+export const HeaderCell = styled.div`
```
keithtxw, line 40: I will use `Text.H6` explicitly instead of setting it this way

src/calendar/calendar-day.style.tsx
```diff
@@ -80,47 +68,60 @@ export const GrowDayCell = styled.div<GrowDayCellProps>`
     }}
 `;
 
-const DayCellBasic = styled.div`
-    position: relative;
+export const OverflowDisplay = styled.div<OverflowDisplayProps>`
```
keithtxw, line 71: Right now from what i see, no matter what, there'll always be a circle on the start or end. Hence i think there is no need for a semi circle concept. But again I could be wrong. We can discuss

src/calendar/calendar-day.tsx
```diff
@@ -79,16 +64,62 @@ export const CalendarDay = ({
...
+                );
+            });
+        });
+    };
+
     return (
         <>
-            <WeekDaysContainer>
+            <Wrapper>
+                {renderHeader()}
```
keithtxw, line 112: We can actually render everything on the same layer instead of having an additional `WeekDaysContainer` and `CalendarDaySection`

src/calendar/calendar.tsx
```diff
@@ -22,7 +24,7 @@ import { CalendarYear } from "./calendar-year";
 import { CalendarHelper } from "../util/calendar-helper";
 import { Text } from "../text";
 
-export type View = "day" | "month" | "year";
+export type View = "default" | "month-options" | "year-options";
```
keithtxw, line 27: I think this is clearer for understanding

src/calendar/calendar.tsx
```diff
@@ -99,56 +101,108 @@ export const Calendar = ({
...
-        setCurrentView("month");
+        if (currentView !== "month-options") {
+            setCurrentView("month-options");
+        } else {
+            setCurrentView("default");
+        }
     };
 
     const toggleYearView = () => {
-        setCurrentView((prev) => {
```
keithtxw, line 106: Somehow the state doesnt change accurately. Hence i switched to be direct.

src/calendar/calendar.tsx
```diff
@@ -99,56 +101,108 @@ export const Calendar = ({
...
+                        id="year-dropdown"
+                        onClick={toggleYearView}
+                    >
+                        <Text.H4 weight="regular">
+                            {getYearHeaderText()}
+                        </Text.H4>
+                        <IconChevronDown />
+                    </DropdownButton>
+                </HeaderDropdown>
+                {renderCalendarContent()}
```
keithtxw, line 181: Here i am flattening the layers as well

src/calendar/calendar.tsx
```diff
@@ -99,56 +101,108 @@ export const Calendar = ({
...
+                        onSelect={handleDateSelect}
+                    />
+                );
+        }
     };
 
     return (
         <Container $type={type} {...otherProps}>
-            <HeaderDropdown $view={currentView}>
+            <SideArrowButton $direction="left" onClick={handleLeftArrowClick}>
```
keithtxw, line 153: Instead of adding false margin, we should position them correctly to make full use of `justify-content` style in `Container`. I also added a larger hit zone for the side arrow buttons for easier clicking

src/calendar/calendar.style.tsx
```diff
@@ -148,37 +169,62 @@ const ViewSectionBase = styled.div`
...
+    padding: 0.625rem;
+    background: transparent;
+
+    :focus,
+    :active {
+        background: transparent;
+    }
+`;
+
+export const SideArrowButton = styled(IconButton)<SideArrowButtonStyleProps>`
```
keithtxw, line 186: This is styled to give a larger hit zone for clicking

src/calendar/calendar-day.tsx
Wilker0725, line 116: the `Fragments` seem didn't any purpose , i will removed.

PR #99, opened 2023-03-01
[BOOKINGSG-3979][AS] Otp Input Component

afiq0209:
**Changes**
Otp Input Component (Standalone)

- [keep] branch

**Additional information**

- You may refer to this [BOOKINGSG-3979](https://jira.ship.gov.sg/secure/RapidBoard.jspa?rapidView=1845&projectKey=BOOKINGSG&view=planning&selectedIssue=BOOKINGSG-3979&quickFilter=5612&issueLimit=100)
- OTP Input Behavior: 

1. Except only number
2. Paste should only be allow if have the correct length.
3. Input focus stay on user clicked field while pasting.
4. Accepted Keyboard Interaction (Tab, Backspace & Shift + Tab)


src/otp-input/types.ts
```diff
@@ -0,0 +1,20 @@
+import { ButtonProps } from "src/button";
+
+export interface OtpInputProps extends React.AriaAttributes {
+    value?: string | undefined;
+    error?: boolean | undefined;
```
keithtxw, line 5: i think this can be derived from `errorMessage`? Don't need the extra prop

src/otp-input/types.ts
```diff
@@ -0,0 +1,20 @@
...
+    errorMessage?: string | undefined;
+    "data-testid"?: string | undefined;
+    cooldownDuration: number | undefined;
+    ctaProps: ButtonProps | undefined;
```
keithtxw, line 9: rename to actionButtonProps for clarity

src/otp-input/types.ts
```diff
@@ -0,0 +1,20 @@
+import { ButtonProps } from "src/button";
+
+export interface OtpInputProps extends React.AriaAttributes {
+    value?: string | undefined;
+    error?: boolean | undefined;
+    errorMessage?: string | undefined;
+    "data-testid"?: string | undefined;
+    cooldownDuration: number | undefined;
```
keithtxw, line 8: Mandatory prop should not accept `undefined`. Please review all the other props too.

src/otp-input/types.ts
```diff
@@ -0,0 +1,20 @@
...
+    ctaProps: ButtonProps | undefined;
+    numOfInput: number | undefined;
+    onChange?: ((value: string) => void) | undefined;
+
+    // Input props
```
keithtxw, line 13: None of the props here seem to be relevant. I think they can be omitted

src/otp-input/types.ts
```diff
@@ -0,0 +1,20 @@
+import { ButtonProps } from "src/button";
+
+export interface OtpInputProps extends React.AriaAttributes {
+    value?: string | undefined;
```
keithtxw, line 4: add `id`, `className` props as well. `className` is important else the user can't modify the style

stories/otp-input/doc-elements.tsx
```diff
@@ -0,0 +1,7 @@
+import styled from "styled-components";
+
+export const Container = styled.div`
+    display: flex;
+    flex-direction: column;
+    max-width: 500px;
```
keithtxw, line 6: use `rem`

src/otp-input/otp-input.tsx
```diff
@@ -0,0 +1,176 @@
...
+        styleType = "secondary",
+        ...otherCtaProps
+    } = ctaProps ?? {};
+    const inputRefs = useRef<(HTMLInputElement | null)[]>([]);
+
+    const [otpValues, setOTPValues] = useState<string[]>(
+        new Array(numOfInput).fill("")
+    );
+    const [countDown, setCountDown] = useState<number>(cooldownDuration);
+    const [enableCountDownTimer, setEnableCountDownTimer] =
```
keithtxw, line 34: I am thinking if this flag might not be necessary. On the click of a button, we can just set the `countDown` to the `cooldownDuration`.

src/otp-input/otp-input.tsx
```diff
@@ -0,0 +1,176 @@
...
+                setEnableCountDownTimer(false);
+            }
+        }, 1000);
+        return () => clearInterval(interval);
+    }, [enableCountDownTimer]);
+
+    // =============================================================================
+    // EVENT HANDLERS
+    // =============================================================================
+    const handleOnKeyDown = (
```
keithtxw, line 71: our convention is `handle<Action>` hence the word `on` is not necessary. `on` is used for the prop type to indicate a callback event

keithtxw (changes requested):

src/otp-input/otp-input.tsx
```diff
@@ -0,0 +1,176 @@
...
+        event:
+            | React.KeyboardEvent<HTMLInputElement>
+            | React.ChangeEvent<HTMLInputElement>,
+        index: number
+    ) => {
+        const { key } = event as React.KeyboardEvent<HTMLInputElement>;
+        const invalidKeys = ["E", "e", "+", "-", "."];
```
keithtxw, line 78: Based on past experiences, we had issues trying to track the key properly since some browsers return it as `code` instead. Can I trouble you to test this across browsers to make sure it works. 

Else an alternative is whether you could use the actual value instead? If the value isnt permitted, unset it and dont set it to the `newOtpValues` array. That way, it wouldn't be filled i suppose?

Can refer to [DateInput](https://github.com/LifeSG/react-design-system/blob/8ad923b41805e83832f7a3a52e357324813687b8/src/date-input/date-input.tsx#L195) which has something similar to restrict to numeric values only.

keithtxw (changes requested):

src/otp-input/types.ts
afiq0209: I think might need some of later in form. but will remove for now.

stories/otp-input/doc-elements.tsx
afiq0209: temp stories for standalone, might no longer need this if we move the stories to form

src/otp-input/otp-input.tsx
afiq0209: to be safe I added both key & code. But yes tested it in Chrome, Safari, Edge & Mozilla.

stories/otp-input/doc-elements.tsx
keithtxw: I dont foresee it will be moved to `Form`. In that case perhaps using `<StoryContainer>` will help?

src/otp-input/otp-input.tsx
```diff
@@ -0,0 +1,182 @@
...
+                            key={index}
+                            id={"otpValues-input"}
+                            data-testid={"otpValues-input"}
+                            ref={(el) => (inputRefs.current[index] = el)}
+                            onKeyDown={(e) => handleKeyDown(e, index)}
+                            type={"number"}
+                            value={data}
+                            error={!!errorMessage}
+                            inputMode={"decimal"}
+                            {...otherProps}
```
keithtxw, line 162: Props like `className`, `id`, `data-testid` should not be passed here. Instead it has to be passed onto `Wrapper`

src/otp-input/otp-input.tsx
```diff
@@ -0,0 +1,182 @@
...
+                        />
+                    );
+                })}
+            </InputContainer>
+            {errorMessage && (
+                <ErrorContainer weight="semibold">
```
keithtxw, line 168: Why not use `FormErrorMessage`?

src/otp-input/otp-input.tsx
```diff
@@ -0,0 +1,182 @@
...
+        if (!value) {
+            return false;
+        }
+        return RegExp(`^[0-9]{${length}}$`).test(value);
+    };
+
+    const renderCTALabel = () => {
+        return otherCtaProps.children
+            ? `${otherCtaProps.children}`
+            : `Resend OTP ${
```
keithtxw, line 140: there is an extra space if there is no countdown. Consider adding the space before "in"

src/otp-input/otp-input.tsx
```diff
@@ -0,0 +1,182 @@
...
+    };
+
+    // =============================================================================
+    // RENDER FUNCTIONS
+    // =============================================================================
+    const validateUserInput = (value: string, length = 1) => {
+        if (!value) {
+            return false;
+        }
+        return RegExp(`^[0-9]{${length}}$`).test(value);
```
keithtxw, line 134: Merge into ternary operator

src/otp-input/otp-input.tsx
```diff
@@ -0,0 +1,182 @@
...
+    };
+
+    return (
+        <Wrapper>
+            <InputContainer>
+                {otpValues.map((data, index) => {
+                    return (
+                        <InputField
+                            key={index}
+                            id={"otpValues-input"}
```
keithtxw, line 154: can omit the curly brackets for purely string values

src/otp-input/otp-input.tsx
```diff
@@ -0,0 +1,182 @@
...
+                setEnableCountDownTimer(false);
+            }
+        }, 1000);
+        return () => clearInterval(interval);
+    }, [enableCountDownTimer]);
+
+    // =============================================================================
+    // EVENT HANDLERS
+    // =============================================================================
+    const handleKeyDown = (
```
keithtxw, line 66: Considering using function curry

```suggestion
    const handleKeyDown = (index: number) => 
            (event: React.KeyboardEvent<HTMLInputElement>) => {
                     // implementations here
            };
```

src/otp-input/otp-input.tsx
```diff
@@ -0,0 +1,182 @@
...
+        <Wrapper>
+            <InputContainer>
+                {otpValues.map((data, index) => {
+                    return (
+                        <InputField
+                            key={index}
+                            id={"otpValues-input"}
+                            data-testid={"otpValues-input"}
+                            ref={(el) => (inputRefs.current[index] = el)}
+                            onKeyDown={(e) => handleKeyDown(e, index)}
```
keithtxw, line 157: Suppose you curried the function then this would be suficient 
```suggestion
                            onKeyDown={handleKeyDown(index)}
```

keithtxw (changes requested):

stories/otp-input/otp-input.stories.mdx
```diff
@@ -0,0 +1,47 @@
...
+
+In the case that you require the otp input field as a standalone, you can do this.
+
+```tsx
+import { OtpInput } from "@lifesg/react-design-system/otp-input";
+```
+
+<Canvas>
+    <Story name="Standalone usage">
+        {() => {
```
keithtxw, line 27: Actually you need not do this. Just render the elements directly

keithtxw (changes requested):

src/otp-input/otp-input.tsx
```diff
@@ -0,0 +1,182 @@
...
+        onClick,
+        styleType = "secondary",
+        ...otherCtaProps
+    } = actionButtonProps ?? {};
+    const inputRefs = useRef<(HTMLInputElement | null)[]>([]);
+
+    const [otpValues, setOtpValues] = useState<string[]>(
+        new Array(numOfInput).fill("")
+    );
+    const [countDown, setCountDown] = useState<number>(cooldownDuration);
```
qroll, line 33: I would recommend using a timestamp to track the start of the countdown. Chrome may delay the timer in inactive tabs, resulting in desynced countdowns

src/otp-input/otp-input.tsx
keithtxw: I am able to type `.` using Android Chrome. I still feel the checking of value as I suggested is a more robust approach. Some keys in mobile devices do not translate to an event key or code

qroll (changes requested):

src/otp-input/otp-input.tsx
keithtxw: Amended to use the timestamp. The countdown is just for cosmetic display

src/otp-input/otp-input.tsx
```diff
@@ -0,0 +1,206 @@
...
+        return otherCtaProps.children
+            ? `${otherCtaProps.children}`
+            : `Resend OTP${
+                  countDown
+                      ? ` in ${countDown} second${countDown > 1 ? "s" : ""}`
+                      : ""
+              }`;
+    };
+
+    const inputId = (index: number) =>
```
keithtxw, line 164: This is a function but the name doesn't look like that. I think the generated id can be the same for the test id also. 
Suggestion is to create a function and keep it in `HELPER FUNCTIONS`

```suggestion
    const generateId = (index: number, prefix?: string) => {
        return prefix
            ? `${prefix}-otp-input-${index + 1}`
            : `otp-input-${index + 1}`;
    };

```

keithtxw (changes requested):

keithtxw (approved):

src/otp-input/otp-input.tsx
```diff
@@ -118,62 +114,96 @@ export const OtpInput = ({
...
-    const validateUserInput = (value: string, length = 1) => {
-        if (!value) {
-            return false;
-        }
-        return RegExp(`^[0-9]{${length}}$`).test(value);
-    };
-
     const renderCTALabel = () => {
         return otherCtaProps.children
             ? `${otherCtaProps.children}`
```
qroll, line 166: return children directly as it could be a jsx element

```suggestion
            ? otherCtaProps.children
```

src/otp-input/otp-input.tsx
```diff
@@ -44,65 +51,54 @@ export const OtpInput = ({
     }, []);
 
     useEffect(() => {
-        if (!enableCountDownTimer || disabled) {
-            return;
+        if (countDown === cooldownDuration) {
+            if (onCooldownStart) onCooldownStart();
+            beginCountdown();
```
qroll, line 56: return a clean up function to clear the interval

src/otp-input/otp-input.tsx
```diff
@@ -118,62 +114,96 @@ export const OtpInput = ({
...
+            currentTime < lastCtaTimestamp.valueOf() + coolDownInMilliseconds
+        );
+    };
+
+    const beginCountdown = () => {
+        let timer = cooldownDuration;
+        const interval = setInterval(() => {
+            timer--;
+            setCountDown(timer);
+            if (timer === 0) {
```
qroll, line 148: 1. calculate remaining seconds left with `Math.ceil((lastCtaTimestamp.valueOf() + cooldownDuration * 1000 - Date.now()) / 1000)`
2. use that for the countdown instead
3. when remaining seconds <= 0, clear interval and trigger `onCooldownEnd`

mainly to keep `onCooldownEnd` synced as much as possible

src/otp-input/otp-input.tsx
```diff
@@ -0,0 +1,212 @@
...
+            if (onChange) {
+                onChange(newOtpValues.join(""));
+            }
+        };
+
+    const handlePaste = (event: ClipboardEvent): void => {
+        const pastedValue = event.clipboardData.getData("text");
+        if (pastedValue && validateUserInput(pastedValue, numOfInput)) {
+            setOtpValues(pastedValue.split(""));
+            if (onChange) {
```
qroll, line 107: risk of stale props: `numOfInput` (unlikely) or `onChange` (possibly)

if this is an issue, these variables will need to be ref'd to keep up to date

src/otp-input/otp-input.tsx
keithtxw: do you mean something like this?
```
const beginCountdown() => {
        const interval = setInterval(....);
        
        return () => {
             clearInterval(interval);
        };
};
```

Then in `useEffect`

```
const cleanup = beginCountdown();

() => {
     cleanup();
};

```

qroll (approved):

PR #101, opened 2023-03-01
Setup testing environment and create unit tests for four components 

jeraldlyh:
**Changes**
- Setup environment for `@testing-library/react`
- Create unit tests for the following components:
  - Accordion
  - Alert
  - Box Container
  - Breadcrumb

**Additional information**
- [keep] branch
- Unable to test `fadeColor` for `Breadcrumb` component


tests/modules/accordion.spec.tsx
```diff
@@ -0,0 +1,122 @@
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
```
keithtxw, line 1: Could you put these files in the same folder format as the `src` directory. I.e. `tests/accordion/accordion.spec.tsx`

Same for the rest

tests/modules/accordion.spec.tsx
```diff
@@ -0,0 +1,122 @@
...
+import { Accordion, AccordionProps, Text } from "../../src";
+import { DEFAULT_TEXT, DEFAULT_TITLE } from "../common";
```
keithtxw, line 3: While it is repetitive, i think it is better if we have the constants reside in their file. It keeps the context within the file and makes it easier to maintain in future, esp when there will be more files present

tests/modules/accordion.spec.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+const renderComponent = (
```
keithtxw, line 5: There are a couple of helper functions here. I tend to prefer the unit tests spec to be as high up the file as possible (unless mocking involved). May I suggest to move these helpers down (below the tests) and add a section header instead?

keithtxw (changes requested):

tests/modules/accordion.spec.tsx
qroll, line 5: these functions can be moved to the end of the tests with a H1 divider indicating `TEST HELPERS`

tests/modules/accordion.spec.tsx
```diff
@@ -0,0 +1,122 @@
...
+        it("should not render expand/collapse button if disabled", () => {
+            renderComponent({ enableExpandAll: false });
+
+            expect(getExpandAllItemsButton(true)).not.toBeInTheDocument();
+        });
+
+        it("should be able to toggle the collapse all button for entire accordion", async () => {
+            renderComponent();
+
+            await waitFor(() => fireEvent.click(getExpandAllItemsButton()));
```
qroll, line 108: what is the use of the `waitFor`? I was expecting an `act` here for state updates would be sufficient

tests/modules/alert.spec.tsx
```diff
@@ -0,0 +1,71 @@
...
+                });
+            }
+        );
+    });
+
+    describe("actionLink", () => {
+        it("should render action link if the prop is provided", () => {
+            renderComponent({ actionLink: { href: "www.google.com" } });
+
+            expect(getActionLink()).toBeInTheDocument();
```
qroll, line 55: would be good to check the link name to verify that the `actionLink.children` is rendered too

tests/modules/accordion.spec.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+            /**
+             * NOTE: Accordion text is wrapped within two nested divs,
+             * hence, requires two traversals up to the div container that contains
+             * the styles
+             */
+            screen.getAllByText(DEFAULT_TEXT).forEach((element) =>
+                expect(element.parentElement.parentElement).toHaveStyle({
```
qroll, line 116: an alternative if you don't want to hardcode the traversals is to use ``.closest(`[data-testid="expandable-container"]`)``

qroll (changes requested):

tests/modules/accordion.spec.tsx
```diff
@@ -0,0 +1,122 @@
...
+        renderComponent();
+
+        expect(screen.getByText(DEFAULT_TITLE)).toBeInTheDocument();
+    });
+
+    it("should not render title if not specified", () => {
+        renderComponent();
+
+        expect(
+            screen.queryByRole("heading", { level: 2 })
```
keithtxw, line 68: i think looking for a particular test id might be more deterministic

keithtxw (changes requested):

keithtxw (approved):

PR #102, opened 2023-03-01
Add conventions documentation

keithtxw:
**Changes**
Add agreed coding conventions

- [delete] branch

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -9,7 +9,12 @@ To enable ease of understanding and consistency, we recommend following the stru
 ```tsx
 // component-file.tsx
 
-// Import statements
+// Import statements here
+/**
+ * When importing other components, use the relative import path
+ * to prevent circular dependency issues.
+ * E.g. import { Text } from "../text/text";
+ */
 
 /**
  * For local props. If the props are to be exported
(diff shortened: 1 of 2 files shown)


CONVENTIONS.md
```diff
@@ -0,0 +1,78 @@
...
+## Code structure
+
+To enable ease of understanding and consistency, we recommend following the structure as such:
+
+> Note: Add headers to the respective sections
+
+```tsx
+// component-file.tsx
+
+// Import statements
```
qroll, line 12: is there a preference for imports to be `"../color/color"` vs `"../color"` vs `"src/color"`? (other than cases where importing from the index may result in some circular dep)

CONVENTIONS.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+<br />
+
+## Prop specification
+
+Here are some guidelines on prop specification:
+
+-   For callbacks, denote with `on` and the action. E.g. `onClick`, `onDismiss`
+-   For optional props, use a Union type and add `undefined` to it
```
qroll, line 68: would be good to include these ones (I tend to miss out on them also -_-)
- include `data-testid` as well as `className`, `id` if not extending from a standard HTML element
- prefer string literals in kebab-case

CONVENTIONS.md
keithtxw: it can't be `src/color` due to circular dep concerns. I'll add that in.

qroll (approved):

PR #110, opened 2023-03-14
[MISC][AS] Handle value props and change return value as string[]

afiq0209:
**Changes**
- Handle value props
- Change the return value as string[]

- [delete] branch

### DIFF
--- a/src/otp-input/otp-input.tsx
+++ b/src/otp-input/otp-input.tsx
@@ -11,6 +11,7 @@ import { FormErrorMessage } from "src/form/form-label";
 
 export const OtpInput = ({
     id,
+    value,
     "data-testid": dataTestId,
     className,
     cooldownDuration,
@@ -64,6 +65,12 @@ export const OtpInput = ({
         onChangeRef.current = onChange;
     }, [onChange]);
 
+    useEffect(() => {
+        if (value) {
+            setOtpValues(value);
+        }
+    }, [value]);
+
     // =============================================================================
     // EVENT HANDLERS
     // =============================================================================
@@ -80,7 +87,7 @@ export const OtpInput = ({
                 setOtpValues(newOtpValues);
 
                 if (onChange) {
-                    onChange(newOtpValues.join(""));
+                    onChange(newOtpValues);
                 }
             }
         };
@@ -94,14 +101,16 @@ export const OtpInput = ({
                 if (newOtpValues[index] !== "") {
                     newOtpValues[index] = "";
                 } else {
-                    newOtpValues[index - 1] = "";
-                    inputRefs.current[index - 1]?.focus();
+                    if (index !== 0) {
+                        newOtpValues[index - 1] = "";
+                        inputRefs.current[index - 1]?.focus();
+                    }
                 }
 
                 setOtpValues(newOtpValues);
 
                 if (onChange) {
-                    onChange(newOtpValues.join(""));
+                    onChange(newOtpValues);
                 }
             }
         };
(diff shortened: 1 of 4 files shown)


src/otp-input/otp-input.tsx
```diff
@@ -111,7 +120,7 @@ export const OtpInput = ({
         if (pastedValue && validateUserInput(pastedValue, numOfInput)) {
             setOtpValues(pastedValue.split(""));
```
keithtxw, line 121: just create a const for `pastedValue.split("")`

src/otp-input/otp-input.tsx
```diff
@@ -11,6 +11,7 @@ import { FormErrorMessage } from "src/form/form-label";
 
 export const OtpInput = ({
     id,
+    value,
```
keithtxw, line 14: why not default to empty array and then u need not check for it's existence in the effect?

src/otp-input/otp-input.tsx
```diff
@@ -94,14 +101,16 @@ export const OtpInput = ({
                 if (newOtpValues[index] !== "") {
                     newOtpValues[index] = "";
                 } else {
```
keithtxw, line 103: Use a `else-if`

src/otp-input/types.ts
```diff
@@ -4,17 +4,17 @@ export interface OtpInputProps extends React.AriaAttributes {
     id?: string | undefined;
     "data-testid"?: string | undefined;
     className?: string | undefined;
-    /** The combined value of the Otp inputs */
-    value?: string | undefined;
+    /** The value of the Otp inputs */
```
keithtxw, line 7: `/** The array of input values */`

src/otp-input/types.ts
```diff
@@ -4,17 +4,17 @@ export interface OtpInputProps extends React.AriaAttributes {
...
     errorMessage?: string | undefined;
     /** The duration (in seconds) to disable the submit button after a click is made */
     cooldownDuration: number;
     /** The props for the submit button */
     actionButtonProps?: ButtonProps | undefined;
     /** The number of characters for the Otp */
     numOfInput: number;
-    /** Called when one of the input is changed. Returns the combined value of all Otp inputs */
-    onChange?: ((value: string) => void) | undefined;
+    /** Called when one of the input is changed. Returns the value of all Otp inputs */
```
keithtxw, line 16: `Called when one of the input is changed. Returns an array of all the input values`

stories/otp-input/otp-input.stories.mdx
```diff
@@ -62,13 +62,16 @@ you with it.
     <Story name="With custom action button">
         {() => {
             const [isInCooldown, setIsInCooldown] = useState(true);
+            const [otpValues, setOtpValues] = useState(new Array(4).fill(""));
```
keithtxw, line 65: for a good example, set a const of `numOfInput = 4` then use it to form up the state and pass it to the component

src/otp-input/otp-input.tsx
```diff
@@ -64,6 +65,12 @@ export const OtpInput = ({
         onChangeRef.current = onChange;
     }, [onChange]);
 
+    useEffect(() => {
+        if (value) {
```
keithtxw, line 69: Here we might want to check if the length of array matches the `numOfInput` before assigning.

keithtxw (changes requested):

keithtxw (changes requested):

keithtxw (approved):

PR #117, opened 2023-03-21
Refactor main logic, add reducer

Wilker0725:
**Changes**
Description of changes...

- [delete] branch

**Changelog entry**

- [BREAKING] Description of change in `input-reducer`

**Additional information**

- You may refer to this [ticket](https://www.figma.com/file/us90jOpWBahsqK2VAluPvD/BookingSG-Design-System?node-id=3360-23100&t=1MSRxEGlM83Jsv4X-0)


src/date-input/dateInputReducer.tsx
keithtxw, line 1: File name does not conform to convention

src/date-input/useEventListener.tsx
keithtxw, line 1: Filename does not conform to the naming convention. Also think that the file should reside in `util` folder instead

src/util/date-input-helper.ts
```diff
@@ -1,34 +1,128 @@
...
+import isBetween from "dayjs/plugin/isBetween";
+import { INVALID_VALUE } from "../date-input/stand-alone-input";
+import { ChangeValueTypes, RawInputValues } from "../date-input";
+
+dayjs.extend(isBetween);
 
 export namespace DateInputHelper {
-    export const validate = (beginDate: string, afterDate: string): boolean => {
+    export const validate = (
+        beginDate: string,
```
keithtxw, line 10: can you rename to `startDate` and `endDate` for clarity

src/util/date-input-helper.ts
```diff
@@ -1,34 +1,128 @@
...
+        afterDate: string,
+        disabledDates?: string[] | undefined,
+        between?: string[] | undefined
+    ): boolean => {
         let pass = false;
 
-        if (!beginDate.length || !afterDate.length) return false;
+        if (!beginDate?.length || !afterDate?.length) return false;
```
keithtxw, line 17: For clarity

```suggestion
        if (startDate.length === 0 || endDate.length === 0) return false;
```

src/util/date-input-helper.ts
```diff
@@ -1,34 +1,128 @@
...
 
-        if (dayjs(beginDate).isAfter(afterDate)) {
-            pass = false;
+        if (dayjs(beginDate).isBefore(afterDate)) {
+            pass = true;
+        }
 
-            return pass;
+        if (dayjs(afterDate).isAfter(beginDate)) {
```
keithtxw, line 23: unnecessary check. the first condition would have already covered the check. Furthermore both resolves to `pass = true`

src/util/date-input-helper.ts
```diff
@@ -1,34 +1,128 @@
...
+    };
+
+    export const singleValidation = (
+        value: string,
+        disabledDates?: string[] | undefined,
+        between?: string[] | undefined
+    ) => {
+        let pass = true;
+
+        if (!value || value === "") return false;
```
keithtxw, line 66:
```suggestion
        if (value.length === 0) return false;
```

keithtxw (changes requested):

src/util/date-input-helper.ts
```diff
@@ -1,34 +1,128 @@
...
+                    "[]"
+                )
+            );
+
+            pass = result.every(Boolean);
+
+            if (!pass) return false;
+        }
+
+        if (dayjs(beginDate).isAfter(afterDate)) {
```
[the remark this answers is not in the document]
Wilker0725: this was check for the `false` return. previous cases was check for the `true` return.

if startDate isAfter endDate return false.
if startDate isBefore endDate return true.

src/util/date-input-helper.ts
```diff
@@ -1,34 +1,128 @@
...
             pass = false;
 
             return pass;
         }
 
-        if (dayjs(beginDate).isBefore(afterDate)) {
-            pass = true;
+        return pass;
```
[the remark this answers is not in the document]
Wilker0725: yes. it will straight return the value if the value is false. 

Only return true at the end once the both value pass in the first condition and skip other below condition.

```
if (dayjs(startDate).isBefore(endDate)) {
      pass = true;
}
``

File: src/date-input/useEventListener.tsx
Comment (Wilker0725) [reply src/date-input/useEventListener.tsx:1]: useEventListener will be naming to use-event-listener?

File: src/calendar/calendar.tsx [hunk 7]
```diff
@@ -398,10 +424,32 @@ export const Component = (
...
+        const disabled = !isDayView ? false : isDisabled;
+
         return (
             <ActionButtonSection>
                 <CancelButton onClick={handleCancelButton}>Cancel</CancelButton>
-                <DoneButton onClick={handleDoneButton}>Done</DoneButton>
+                <DoneButton
+                    ref={doneButtonRef}
+                    onClick={() => handleDoneButton(disabled)}
+                    $disabled={disabled}
```
qroll, line 449: for buttons, use the `disabled` prop instead

src/util/date-input-helper.ts
```diff
@@ -1,34 +1,124 @@
...
+    };
+
+    export const singleValidation = (
+        value: string,
+        disabledDates?: string[] | undefined,
+        between?: string[] | undefined
+    ) => {
+        let pass = true;
```
qroll, line 60: rename this fn to `validateSingle`

no need to maintain this pass variable. can do something like

```
() => {
  if (cond1) {
    if (cond2) {
      return false
    }
  }
  return true;
}
```

src/date-input/date-input.tsx
```diff
@@ -254,56 +274,128 @@ export const DateInput = ({
...
+            onChange(returnValue);
         }
 
         if (onChangeRaw) {
-            const rawInputValues = getFormattedRawValue(values);
+            const returnRawValue =
+                DateInputHelper.getFormattedRawValue(returnValue);
 
-            onChangeRaw(rawInputValues);
+            onChangeRaw(returnRawValue);
```
qroll, line 288: intention of `onChangeRaw` is even if user enters an invalid date like `30-02-2023`, the callback would return `["30","02","2023"]`. but since now it's derived from standalone-input's `onChange` you'd now get ["Invalid Date", "Invalid Date", "Invalid Date"] instead

perhaps you can expose an `onChangeRaw` from standalone-input as well

src/date-input/date-input.tsx
```diff
@@ -466,28 +574,33 @@ export const DateInput = ({
...
                 action={startDate.currentType}
-                isActive={calendarOpen}
+                focusType={currentElement.type}
+                isOpen={calendarOpen}
+                isError={isError}
+                withButton={withButton}
             />
             {RenderRangeInput()}
-            {RenderIndicatedBar()}
+            {RenderIndicateBar()}
```
qroll, line 596: these functions should be in camelCase as they are not components

qroll (changes requested):

src/date-input/date-input.tsx
```diff
@@ -69,21 +73,36 @@ export const DateInput = ({
...
+        // avoid to run in first mounted
+        if (!isMouted.current) {
+            isMouted.current = true;
+            return;
+        }
+
+        if (!calendarOpen) {
+            handleBlur(startDate, endDate);
+            setIsError(false);
+        }
```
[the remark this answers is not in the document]
Wilker0725: Good question!

Actually the value haven't been updated into reducer yet.  

Flow/Scenario:
a. blank input value -> select value -> clickOutSide/Cancel/Done -> useEffect able to get the latest value
b. blank input value -> select value -> clickOutSide/Cancel/Done -> handleBlurContainer get the previous value since the reducer not been updated yet.

src/calendar/calendar.tsx
```diff
@@ -398,10 +424,32 @@ export const Component = (
...
 
+        let isDisabled = true;
+        const isDayView = ["default"].includes(currentView);
+
+        switch (variant) {
+            case "single":
+                isDisabled = selectedStartDate?.length ? false : true;
+                break;
+            case "range":
+                isDisabled = ![selectedStartDate, selectedEndDate].every(
```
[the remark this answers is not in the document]
Wilker0725: Oops, it may not work correctly.

button disabled when both value didn't been selected yet.
`isDisabled` value  `true`

button active when both value been selected
`isDisabled`  value `false`

so when 
!selectedStartDate &&  !selectedEndDate  
          true                                  false              // return false, button active
          true                                  true               // correct one -> both value selected, button active

src/calendar/calendar.tsx
```diff
@@ -398,10 +424,32 @@ export const Component = (
     const renderCancelDoneButton = () => {
         if (type === "standalone" || !withButton) return;
 
+        let isDisabled = true;
+        const isDayView = ["default"].includes(currentView);
+
+        switch (variant) {
+            case "single":
+                isDisabled = selectedStartDate?.length ? false : true;
```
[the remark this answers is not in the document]
Wilker0725: Thanks! 

I will update to

```
selectedStartDate ? false : true
```

**selectedStartDate**|**button status**
:-----:|:-----:
true|disabled
False|active

src/util/date-input-helper.ts
Wilker0725: Note:

my bad, i couldn't understand clearly. i updated to this. let me know if it is not what you expected.
```
() => {
  if (value.length === 0) {
    return false;
  }

  if (disabledDates && disabledDates.some((disabledDate) => value === disabledDate)) {
    return false;
  }

  if (between && !dayjs(value).isBetween(between[0], between[1], "day", "[]")) {
    return false;
  }

  return true;
}

```

src/util/date-input-helper.ts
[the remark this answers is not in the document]
Wilker0725: remove the variable in validation helper, updated to:

```
() => {
 if(cond1) {
    return false
  }

  if(cond2) {
    return false
  }

 return true
}

```

src/util/date-input-helper.ts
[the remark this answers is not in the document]
Wilker0725: removed the `pass` variable.  updated to:

```
() => {
 if(cond1) {
    return false
  }

  if(cond2) {
    return false
  }

 return true
}

```

src/date-input/date-input.tsx
Wilker0725: Note: current i was return an object because we need to indicate startDate and endDate. Let me know if require to return an array type, please.

```
{
  start: {year: "", month: "", day: ""},
  end: {year: "", month: "", day: ""},
}
```

src/calendar/calendar.tsx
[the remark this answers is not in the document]
qroll: ah, it should have been `!selectedStartDate || !selectedEndDate`

src/calendar/calendar.tsx
```diff
@@ -91,14 +92,24 @@ export const Component = (
     }, [isOpen]);
 
     useEffect(() => {
-        if (type !== "standalone") return;
+        if (doneButtonRef.current) {
```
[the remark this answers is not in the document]
Wilker0725: hmm.. please check with this.

i use event.target instead and aria-label html attribute for some reason.

qroll (approved):

src/calendar/calendar.tsx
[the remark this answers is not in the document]
qroll: follow up comment: to stick to the custom tab implementation for now. `onBlur` was being triggered unexpectedly but was not able to find the cause.

PR #120, opened 2023-03-22
Creation of Toggle component

keithtxw:
**Changes**
Creation of `Toggle` as part of enhancement to existing `ToggleButton`.

Note: This does not include sub label variant due to pending confirmation on UX side

- [delete] branch

**Changelog entry**

- Add `Toggle` as a replacement for `ToggleButton`
- [WARNING] `ToggleButton` will be deprecated in future releases

**Additional information**

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/MOL-12477)

### DIFF
--- a/src/toggle/toggle.styles.tsx
+++ b/src/toggle/toggle.styles.tsx
@@ -1,0 +1,201 @@
+import styled, { css } from "styled-components";
+import { Color } from "../color";
+import { Text, TextStyle, TextStyleHelper, TextWeight } from "../text";
+import { ToggleStyleType } from "./types";
+
+// =============================================================================
+// STYLE INTERFACES, transient props are denoted with $
+// See more https://styled-components.com/docs/api#transient-props
+// =============================================================================
+interface StyleProps {
+    $selected?: boolean;
+    disabled?: boolean;
+    $indicator?: boolean;
+}
+
+interface ContainerStyleProps extends StyleProps {
+    $styleType?: ToggleStyleType;
+    $error?: boolean;
+    $hasSubLabel?: boolean;
+}
+
+interface LabelStyleProps extends StyleProps {}
+
+// =============================================================================
+// STYLING
+// =============================================================================
+
+const getPadding = (styleType: ToggleStyleType) => {
+    switch (styleType) {
+        case "no-border":
+            return css`
+                padding: 0.75rem 1.0625rem; // compensate the 1px border
+            `;
+        default:
+            return css`
+                padding: 0.6875rem 1rem;
+            `;
+    }
+};
+
+export const Container = styled.div<ContainerStyleProps>`
+    position: relative;
+    display: inline-flex;
+    min-width: 10.375rem;
+    border-radius: 4px;
+    border-width: 1px;
+    border-style: solid;
+    cursor: ${(props) => (props.disabled ? "not-allowed" : "pointer")};
+    ${(props) => getPadding(props.$styleType)}
+
+    // Content positioning style
+    ${(props) => {
+        if (!props.$indicator) {
+            return css`
+                justify-content: center;
+            `;
+        }
+    }}
+
+    // Background, Hover and Border style
+    ${(props) => {
+        switch (props.$styleType) {
+            case "no-border": {
+                if (props.$error) {
+                    return css`
+                        border-color: ${Color.Validation.Red.Icon};
+
+                        :hover {
+                            box-shadow: 0 0 4px 1px ${Color.Shadow.Red};
+                        }
+                    `;
+                } else if (!props.disabled) {
+                    return css`
+                        border: none;
+
+                        :hover {
+                            background: ${Color.Accent.Light[6]};
+                        }
+                    `;
+                } else {
+                    return css`
+                        border: none;
+                    `;
+                }
+            }
+
+            default: {
+                if (props.disabled && !props.$selected) {
+                    return css`
+                        background: ${Color.Neutral[6]};
+                        border-color: ${Color.Neutral[5]};
+                    `;
+                } else if (props.disabled && props.$selected) {
+                    return css`
+                        background: ${Color.Neutral[6]};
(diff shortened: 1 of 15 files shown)


qroll (changes requested):

src/toggle/toggle.styles.tsx
```diff
@@ -0,0 +1,201 @@
...
+                    return css`
+                        border-color: ${Color.Validation.Red.Icon};
+
+                        :hover {
+                            box-shadow: 0 0 4px 1px ${Color.Shadow.Red};
+                        }
+                    `;
+                } else if (!props.disabled) {
+                    return css`
+                        border: none;
```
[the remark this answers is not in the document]
keithtxw: Hmm... good idea. I can do that.

src/toggle/toggle.tsx
```diff
@@ -0,0 +1,157 @@
...
+            <Input
+                ref={inputRef}
+                type={type === "checkbox" ? "checkbox" : "radio"}
+                data-testid="toggle-input"
+                disabled={disabled}
+                onChange={handleOnChange}
+                checked={selected}
+            />
+            <TextContainer>
+                <Label
```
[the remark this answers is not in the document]
keithtxw: This is a good point. Initially i saw this as a label for a button but I get where you are coming from since it's a checkbox or radio input under the hood. And it'll be unlikely that someone will still put an explicit label for this.

I went for the generated short uuid or prop passed id approach as it feels cleaner structurally.

src/toggle/toggle.tsx
```diff
@@ -0,0 +1,161 @@
...
+    className,
+    "data-testid": testId,
+    onChange,
+}: ToggleProps) => {
+    // =============================================================================
+    // CONST, STATE, REF
+    // =============================================================================
+    const [selected, setSelected] = useState<boolean>(checked);
+    const inputRef = useRef<HTMLInputElement>();
+    const generatedInputId = id ? `${id}-input` : crypto.randomUUID();
```
qroll, line 32: can make use of lazy initial state to generate a fixed id for the lifecycle of the component

`const [state, setState] = useState(() => crypto.randomUUID())`

keithtxw: in that case won't this work too? 
```
const [inputId] = useState<string>(id ? `${id}-input` : crypto.randomUUID());
```

qroll: hmm yes. although need to be careful to keep `id` up to date.

the main difference is in the first one, the function only gets called once, whereas the second one will generate uuid on every rerender (just that the new value is not saved)

how about
```
const [uuid] = useState(() => crypto.randomUUID())
const generatedInputId = id ? `${id}-input` : uuid;
```

src/toggle/toggle.styles.tsx
```diff
@@ -0,0 +1,187 @@
...
+        }
+    }}
+
+    // Background, Hover and Border style
+    ${(props) => {
+        switch (props.$styleType) {
+            case "no-border": {
+                if (props.$error) {
+                    return css`
+                        border-color: ${Color.Validation.Red.Icon};
```
qroll, line 52: just want to double check that the toggle bg remains transparent here? and also has no bg on hover?

src/toggle/toggle.styles.tsx
```diff
@@ -0,0 +1,187 @@
...
+    ${(props) => {
+        if (props.$disabled) {
+            return css`
+                color: ${Color.Neutral[3](props)};
+            `;
+        } else if (props.$selected) {
+            return css`
+                color: ${Color.Primary(props)};
+            `;
+        }
```
qroll, line 159: if indicator + disabled + selected, color is N1

src/toggle/types.ts
```diff
@@ -0,0 +1,36 @@
...
+    type?: ToggleType | undefined;
+    /** Specifies if the indicator icon should be displayed */
+    indicator?: boolean | undefined;
+    /** Indicates if the element is checked */
+    checked?: boolean | undefined;
+    /** The style type of the Toggle. Values: "default" | "no-border" */
+    styleType?: ToggleStyleType | undefined;
+    /** The Toggle label */
+    children: string;
+    /** The description label displayed below the main label */
```
qroll, line 24: may need to comment out this line with `//` as vscode is associating it with the disabled prop

src/toggle/toggle.tsx
```diff
@@ -0,0 +1,163 @@
...
+    //             {component}
+    //         </SubLabel>
+    //     );
+    // };
+
+    return (
+        <Container
+            $selected={selected}
+            $disabled={disabled}
+            role={type === "checkbox" ? "checkbox" : "radio"}
```
qroll, line 129: can remove this `role` attribute as well, it's causing an issue with Voiceover in Safari

src/toggle/toggle.tsx
```diff
@@ -0,0 +1,163 @@
...
+                                setSelected(true);
+                            }
+                        }
+                        break;
+                }
+
+                if (onChange) onChange(event);
+            }
+        },
+        []
```
qroll, line 73: should include `onChange` dependency here

src/toggle/toggle.styles.tsx
keithtxw: Clarified and we'll stick to N3 for now for all disabled states regardless if indicator present

src/toggle/toggle.tsx
keithtxw: actually i dont think this `useCallback` is required. Will remove.

src/toggle/toggle.styles.tsx
keithtxw: It should have. Will add back. Missed out.

src/toggle/toggle.tsx
keithtxw: Ok this is better. will do that

qroll (approved):

PR #123, opened 2023-03-28
LoadingDotsSpinner: Add ability to change color props and data-testid

afiq0209:
**LoadingDotsSpinner Changes**
- Add ability to replace/change colour
- Add data-testid 

- [delete] branch

**Additional information**

- You may refer to this [BOOKINGSG-4053](https://jira.ship.gov.sg/browse/BOOKINGSG-4053)

### DIFF
--- a/src/animations/loading-dots-spinner/types.ts
+++ b/src/animations/loading-dots-spinner/types.ts
@@ -1,4 +1,0 @@
-export interface LoadingDotsSpinnerProps {
-    color?: string | undefined;
-    "data-testid"?: string | undefined;
-}
--- a/src/animations/loading-dots-spinner/types.ts
+++ b/src/animations/loading-dots-spinner/types.ts
@@ -1,0 +1,4 @@
+export interface LoadingDotsSpinnerProps {
+    color?: string | undefined;
+    "data-testid"?: string | undefined;
+}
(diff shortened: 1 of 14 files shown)


src/animations/loading-dots-spinner/types.ts
```diff
@@ -0,0 +1,4 @@
+export interface LoadingDotsSpinnerProps {
```
keithtxw, line 1: I feel that this is going to be a generic prop for all lottie animations in future. Perhaps this could be made generic and added on a higher level instead.

keithtxw:
Btw, do take note branch naming convention is kebab-case.

src/animations/loading-dots-spinner/types.ts
keithtxw: I have applied the changes fyi.

stories/animation/doc-elements.tsx
```diff
@@ -21,7 +22,7 @@ export const AnimationPane = styled.div`
     display: flex;
     justify-content: center;
     align-items: center;
-    background: ${Color.Accent.Light[6]};
+    background: #fbfcfe;
```
keithtxw, line 25: Change this to keep constant. Else could change if we change the theme in storybook

src/animations/types.ts
```diff
@@ -0,0 +1,13 @@
+export interface BaseAnimationProps {
```
keithtxw, line 1: Make this a generic prop which all the animations should have

keithtxw (approved):

PR #125, opened 2023-04-06
Add BookingSG ErrorDisplay assets

nobeltws:
**Changes**
- Added error display data for BSG
- Added function to get error display based on theme (e.g. base/bookingsg)
- Updated mock for error display tests

- [delete] branch

**Changelog entry**
- Added error display data for BSG

**Additional information**

- You may refer to this [BOOKINGSG-3895](https://jira.ship.gov.sg/browse/BOOKINGSG-3895)
- This was a requirement from Michelle so that we have copy and illustrations for BSG directly from DS (Illustrations are not done yet, will raise another PR to update illustrations once its done)

PR #128, opened 2023-04-10
Add resourceScheme key to the theme

keithtxw:
**Changes**
Add `resourceScheme` key to the theme object for use in `ErrorDisplay` in [PR](https://github.com/LifeSG/react-design-system/pull/125)

- [delete] branch

**Changelog entry**

- Add `resourceScheme` key to the theme object for toggling of image assets and even text content

### DIFF
--- a/src/theme/types.ts
+++ b/src/theme/types.ts
@@ -29,6 +29,11 @@ export type DesignTokenCollectionsMap = {
 };
 
 // =============================================================================
+// RESOURCE THEMES
+// =============================================================================
+export type ResourceScheme = "base" | "bookingsg";
+
+// =============================================================================
 // GENERAL
 // =============================================================================
 export enum ThemeContextKeys {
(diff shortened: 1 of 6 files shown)


nobeltws (approved):
Looks good! will also update this into the theme on BSG repo

src/theme/types.ts
```diff
@@ -54,6 +60,8 @@ export interface ThemeSpec {
     [ThemeContextKeys.designTokenScheme]: DesignTokenScheme;
     /** Sets the layout scheme of the theme */
     [ThemeContextKeys.layout]?: ThemeLayout | undefined;
+    /** Sets the resource scheme (e.g. images, text) of the theme */
+    [ThemeContextKeys.resourceScheme]: ResourceScheme | undefined;
```
qroll, line 64:
```suggestion
    [ThemeContextKeys.resourceScheme]: ResourceScheme;
```

qroll (changes requested):
can try adding a theme declaration so that the keys can be accessed without Typescript errors

https://styled-components.com/docs/api#create-a-declarations-file

src/theme/types.ts
keithtxw: Ah my bad. i made it optional previously. thanks for spotting

qroll (approved):

PR #133, opened 2023-04-19
Bookingsg 3758  time range picker

baktavatsalam86:
**Changes**
Description of changes...

- [delete] branch

- [WARNING ] Description of change in `Time range Picker`

**Additional information**

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/BOOKINGSG-3758)
- Any other information you would like to include [https://www.figma.com/file/us90jOpWBahsqK2VAluPvD/BookingSG-Design-System?node-id=3375-28403&t=MZHAvOeNRHWajDiu-0](url)
-


qroll:
the storybook build is failing due to a case-sensitivity issue with the directory/file names. you can try the `git mv` command to fix it

also the repository convention is to use `kebab-case`, so I would suggest renaming `timerangepicker` to `time-range-picker`

src/form/form-timerangepicker.tsx
```diff
@@ -0,0 +1,29 @@
+import { TimeRangePicker } from "../timeRangePicker";
```
keithtxw, line 1: i am seeing the file name all in lower caps but you are doing a camelCase here? Either way this is against the convention. 

Suggestion to call it `time-range-picker` instead

src/timerangepicker/timerangepicker.tsx
keithtxw, line 1: Instead of copying the code in Timepicker and duplicating here, I am hoping you could extract out the common things in Timepicker into a shared component for it to be used by Timepicker and TimeRangePicker.

src/timerangepicker/helper.ts
keithtxw, line 1: This is an exact copy of the helper file in timepicker. I do not think this is an ideal implementation method.

keithtxw (changes requested):

src/time-range-picker/time-range-picker.tsx
```diff
@@ -0,0 +1,220 @@
...
+                    onFocus={handleStartTimeFocus}
+                    $focused={showStartTimeSelector}
+                    readOnly={readOnly}
+                    placeholder={"From"}
+                    value={TimeRangePickerHelper.formatValue(
+                        startTimeVal,
+                        format
+                    )}
+                    defaultValue={defaultValue}
+                    $disabled={disabled}
```
qroll, line 170: inputs come with a `disabled` prop, you should use that instead to get the correct native browser behaviour.

src/time-range-picker/time-range-picker.tsx
```diff
@@ -0,0 +1,220 @@
...
+                    $error={error}
+                    data-testid={
```
qroll, line 172: use `data-testid` prop instead

src/time-range-picker/time-range-picker.tsx
```diff
@@ -0,0 +1,220 @@
...
+                <TimeRangePickerDropdown
+                    id={id}
+                    show={showStartTimeSelector}
+                    value={startTimeVal || defaultValue}
+                    format={format}
+                    onCancel={handleSelectionDropdownCancel}
+                    onChange={handleStartTime}
+                />
+
+                <ArrowRangeIcon>
```
qroll, line 187: `ArrowRangeIcon` does not need to be a button as it has no functionality

src/time-range-picker/types.ts
```diff
@@ -0,0 +1,29 @@
+export type TimeRangePickerFormat = "12hr" | "24hr";
+
+export interface TimeRangePickerProps {
```
qroll, line 3: is this interface supposed to share the same name as the one below?

src/time-range-picker/types.ts
```diff
@@ -0,0 +1,29 @@
...
+    startTime: string;
+    endTime: string;
+}
+
+export interface TimeRangePickerProps extends React.AriaAttributes {
```
qroll, line 8: some thoughts on the typings
- no need to extend `AriaAttributes`
- `name`, `tabIndex`, and `onSelectionCancel` can be removed, I don't think they're relevant
- I also think `placeholder` and `defaultValue` can be removed for now, but if there's a use case for them, they should at least match the  typing of `value` since they should correspond to two inputs

qroll (changes requested):

src/time-range-picker/time-range-picker.tsx
```diff
@@ -0,0 +1,220 @@
...
+        onSelectionCancel && onSelectionCancel();
+    };
+
+    const handleStartTime = (value: string) => {
+        setShowStartTimeSelector(false);
+        setShowEndTimeSelector(true);
+        setStartTimeVal(value);
+
+        const timeValue: TimeRangePickerProps = {
+            startTime: startTimeVal,
```
qroll, line 107: the `startTimeVal` variable has not been updated in this callback yet. same issue for the end time callback
```suggestion
            startTime: value,
```

src/time-range-picker/time-range-picker.tsx
```diff
@@ -0,0 +1,220 @@
...
+            setShowStartTimeSelector(true);
+        }
+    }, [showStartTimeSelector]);
+
+    const handleEndTimeFocus = useCallback(() => {
+        if (!disabled && !readOnly && !showEndTimeSelector) {
+            setShowStartTimeSelector(false);
+            setShowEndTimeSelector(true);
+        }
+    }, [showEndTimeSelector]);
```
qroll, line 77: are the `useCallback`s needed? it seems like these event handlers need to access the updated state values anyway

src/time-range-picker/time-range-picker.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+    useEffect(() => {
+        if (value) {
+            setStartTimeVal(value.startTime);
+            setEndTimeVal(value.endTime);
+        }
+    }, []);
+    useEffect(() => {
+        document.addEventListener("mousedown", handleMouseDownEvent);
+        document.addEventListener("keyup", handleKeyUpEvent);
```
qroll, line 54: noticed some inconsistent behaviour with Shift+Tab

https://github.com/LifeSG/react-design-system/assets/16853062/3daca06c-ba88-4a5c-9a9c-7e7a31a72614

it might be easier to use `onBlur` to manage focus events

src/time-range-picker/timerangepicker-dropdown.styles.tsx
keithtxw, line 1: Could you rename file to `time-range-picker-dropdown.styles.tsx`

src/time-range-picker/timerangepicker.styles.tsx
```diff
@@ -0,0 +1,227 @@
...
+export const BottomHighlightStartTime = styled.div`
+    position: absolute;
+    bottom: 0;
+    height: 2px;
+    left: 1rem;
+    right: 23rem;
+    background: ${BookingSGColorSet.Neutral[8]};
+    background-color: ${Color.Accent.Light[1]};
+
+    ${MediaQuery.MaxWidth.tablet} {
```
keithtxw, line 109: moving downwards, they are all 40% so there is no need to specify for the smaller viewports anymore

src/time-range-picker/timerangepicker.styles.tsx
```diff
@@ -0,0 +1,227 @@
...
+`;
+
+export const ArrowRight = styled(ArrowRightIcon)`
+    color: ${Color.Neutral[3]};
+    cursor: pointer;
+    width: 1.125rem;
+    height: 1.125rem;
+`;
+
+export const BottomHighlightStartTime = styled.div`
```
keithtxw, line 100: Actually i think you can control the border with the :focus? isn't that possible?

src/time-range-picker/timerangepicker.styles.tsx
```diff
@@ -0,0 +1,227 @@
...
+            `;
+        }
+    }}
+`;
+
+export const InputSelectorEndTimeElement = styled.input<StyleProps>`
+    ${TextStyleHelper.getTextStyle("Body", "regular")}
+
+    display: block;
+    // padding: 0.2rem 1rem 0.3rem 1rem; // Somehow the input text appears lower
```
keithtxw, line 197: remove if not needed

keithtxw (changes requested):

keithtxw (changes requested):
I think for my changes, we can deal with them in the proper refactor. But we need to fix the true bugs

src/timerangepicker/types.ts
```diff
@@ -0,0 +1,29 @@
...
+    style?: React.CSSProperties | undefined;
+    tabIndex?: number | undefined;
+    readOnly?: boolean | undefined;
+    "data-testid"?: string | undefined;
+
+    // Input-specific attributes
+
+    value?: Value | undefined;
+    placeholder?: string | undefined;
+    defaultValue?: string | undefined;
```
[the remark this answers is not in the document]
qroll: no longer relevant as prop has been removed

qroll (approved):

keithtxw (approved):

PR #137, opened 2023-05-03
Date input long lived

Wilker0725:
**Changes**
Description of changes...

- [delete] branch

**Changelog entry**

- [BREAKING | WARNING | remove if none] Description of change in `Calendar & DateInput`

**Additional information**


qroll (changes requested):

qroll:
month/year view takes the day of the previous selection

https://user-images.githubusercontent.com/16853062/236089129-e593098b-3f2c-4f81-9170-b8275a0aa171.mov

src/calendar/types.ts
```diff
@@ -1,21 +1,61 @@
...
     onSelect?: ((value: string) => void) | undefined;
-    /** Selected date in `YYYY-MM-DD` format */
+    /** Called when day cell is hover, returns value in `YYYY-MM-DD` */
+    onHover?: ((value: string) => void) | undefined;
+    /** Selected start date in `YYYY-MM-DD` format */
     value?: string | undefined;
     /** The display type of the component. Values `standalone` | `input` */
     type?: CalendarType | undefined;
 
+    // DateInput with calendar component props
```
qroll, line 20: the following props all seem to be for internal use with the date input only

we can introduce a standalone Calendar that exposes only these props
- between
- disabledDates
- onSelect
- onHover
- value

and internally uses this shared calendar component as `type=standalone`

and in the future we could have a RangedCalendar that also takes in
- endValue
- onSelect (with start + end)

and it would manage the `currentFocus` on the shared calendar component

Wilker0725:
> month/year view takes the day of the previous selection
> 
>  Screen.Recording.2023-05-04.at.9.29.28.AM.mov

thanks for your time and spotted this bug. i didn't check it carefully.

src/calendar/calendar.tsx
```diff
@@ -156,52 +389,124 @@ export const Calendar = ({
...
+    };
+
+    // handle the animation transition
+    if (type === "input" && !isOpen)
+        return <Container $type={type} tabIndex={-1} />;
+
     return (
-        <Container $type={type} {...otherProps}>
-            <SideArrowButton $direction="left" onClick={handleLeftArrowClick}>
+        <Container $type={type} $isOpen={isOpen} {...otherProps}>
```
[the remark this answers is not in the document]
Wilker0725: thanks for your time to figure it out for this matter

it happened a bug when click another input element after select a value from calendar.

this function already been removed from current updates. let me debug it.

keithtxw (changes requested):
I am still reviewing but off hand here are some changes i am proposing

stories/form/form-date-input/form-date-input.stories.mdx
```diff
@@ -24,24 +26,107 @@ The field that allows a user to enter a date value.
...
+        </StoryContainer>
+    </Story>
+</Canvas>
+
+<Heading3>Using a range selection</Heading3>
+
+You can use `variant='range'` to specify to range selection.
+
+<Canvas>
+    <Story name="Ranged DateInput">
```
keithtxw, line 52:
```suggestion
    <Story name="Range input">
```

keithtxw: Also for this story, can you add a state to it. Right now because of the lack of state, I cannot see the range display properly

keithtxw (changes requested):

src/date-input/stand-alone-input.tsx
```diff
@@ -166,13 +164,20 @@ export const StandAloneInput = ({
         // remove the cursor if it
         // - field is 'none'
         // - error happened from range value check
-        if (focusType === "none" || isError) {
+        if (focusType === "none") {
```
qroll, line 167: i think this useEffect is no longer needed as the blur event is now handled by the parent. `isError` is also no longer required?

src/shared/internal-calendar/internal-calendar-day.tsx
```diff
@@ -192,8 +191,18 @@ export const CalendarDay = ({
 
         const styleLeftProps: StyleProps = {},
             styleRightProps: StyleProps = {},
-            styleCircleProps: StyleCircleProps = {},
-            styleLabelProps: StyleLabelProps = {};
+            styleCircleProps: StyleProps = {},
+            styleLabelProps: StyleProps = {};
+
+        let isSelectedSame = false;
```
qroll, line 197: can be a const declaration

src/util/date-input-helper.ts
```diff
@@ -42,6 +44,14 @@ export namespace DateInputHelper {
             return false;
         }
 
+        if (
```
qroll, line 47: combine this with the previous block since there's overlapping conditions
```
if (between && between.length) {
  if invalid format for between
        return false
  if start or end are invalid
       return false
}
```

stories/form/form-date-input/form-date-input.stories.mdx
```diff
@@ -25,59 +26,121 @@ The field that allows a user to enter a date value.
...
+                dayjs().day(15).format("YYYY-MM-DD"),
+                dayjs().day(25).format("YYYY-MM-DD"),
+            ]);
+            return (
+                <StoryContainer>
+                    <Container>
+                        <Form.DateInput
+                            label="This date input with disabled dates"
+                            disabledDates={disabledDates}
+                        />
```
qroll, line 111: can include the ranged variant in the same story as well

qroll (approved):

qroll:
noting down some behaviours that were discussed
- onChange to trigger when a valid selection is made
  - for `withButton=true`, this means when user confirms the selection with the Done button
  - otherwise, when a valid input or calendar selection is made
  - for ranged selection, both start + end should be selected as well
- onChangeRaw triggers when onChange does
- onBlur when the field is defocused

in most cases onChange should be sufficient

PR #142, opened 2023-05-06
Fix multi select bug

keithtxw:
**Changes**
- Fix issue of checkbox checkmark not rendering when checked
- Issue occurs if you use a non-object array as the `options`

- **NOTE** this is merging into release branch `v1.0.2`. A separate PR will be created to merge the release branch back to `master`
- [ delete] branch

**Changelog entry**

- Fix checkmark indicator not appearing for selected items in `Form.MultiSelect` when passing in non-object arrays as options

### DIFF
--- a/src/shared/dropdown-list/dropdown-list.tsx
+++ b/src/shared/dropdown-list/dropdown-list.tsx
@@ -1,4 +1,4 @@
-import findIndex from "lodash/findIndex";
+import indexOf from "lodash/indexOf";
 import React, { useEffect, useRef, useState } from "react";
 import { useSpring } from "react-spring";
 import { Spinner } from "../../button/button.style";
@@ -172,7 +172,7 @@ export const DropdownList = <T, V>({
     };
 
     const checkListItemSelected = (item: T): boolean => {
-        return findIndex(selectedItems, item as any) > -1;
+        return indexOf<T>(selectedItems, item) > -1;
     };
 
     const filterAndUpdateList = (searchValue: string) => {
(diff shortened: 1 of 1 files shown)


src/shared/dropdown-list/dropdown-list.tsx
```diff
@@ -172,7 +172,7 @@ export const DropdownList = <T, V>({
     };
 
     const checkListItemSelected = (item: T): boolean => {
-        return findIndex(selectedItems, item as any) > -1;
+        return indexOf<T>(selectedItems, item) > -1;
```
qroll, line 175: it will be more reliable to use `valueExtractor` to compare items instead of relying on object reference. although I understand there's currently no requirement that value has to be unique

the below doesn't work, for example

```
<Form.MultiSelect
    label="This is the select field"
    options={[
        { value: "A", label: "Option A" },
        { value: "B", label: "Option B" },
        { value: "C", label: "Option C" },
        { value: "D", label: "Option D" },
    ]}
    selectedOptions={[{ value: "A", label: "Option A" }]}
    valueExtractor={(item) => item.value}
    listExtractor={(item) => item.label}
    displayValueExtractor={(item) => item.label}
/>
```

alternatively you could use `isMatches` as the comparator only when item is typeof object

qroll (changes requested):

src/shared/dropdown-list/dropdown-list.tsx
keithtxw: Ah my bad. i failed to check properly. Tested on the first few examples which did not have the `selectedOptions` passed. Will take your considerations into account

PR #149, opened 2023-05-16
Fix dropdown height still taking up space even after dismi…

keithtxw:
**Changes**
Noticed that there was excess space generated in Storybook and suspect that the current approach for rendering the dropdown might be causing it. Anyway I switched to make it the appearance and animation consistent to other dropdowns using `react-spring`

- [delete] branch

### DIFF
--- a/src/calendar/calendar.tsx
+++ b/src/calendar/calendar.tsx
@@ -17,5 +17,5 @@ export const Calendar = ({ className, ...otherProps }: CalendarProps) => {
 const Wrapper = styled.div`
     width: 41rem;
     border: 1px solid ${Color.Neutral[5]};
-    border-radius: 8px;
+    border-radius: 12px;
 `;
--- a/src/calendar/calendar.tsx
+++ b/src/calendar/calendar.tsx
@@ -1,6 +1,21 @@
+import styled from "styled-components";
 import { InternalCalendar } from "../shared/internal-calendar";
+import { Color } from "../color";
 import { CalendarProps } from "./types";
 
-export const Calendar = (props: CalendarProps) => {
-    return <InternalCalendar type="standalone" {...props} />;
+export const Calendar = ({ className, ...otherProps }: CalendarProps) => {
+    return (
+        <Wrapper className={className}>
+            <InternalCalendar type="standalone" {...otherProps} />
+        </Wrapper>
+    );
 };
+
+// =============================================================================
+// STYLING
+// =============================================================================
+const Wrapper = styled.div`
+    width: 41rem;
+    border: 1px solid ${Color.Neutral[5]};
+    border-radius: 8px;
+`;
--- a/src/shared/internal-calendar/internal-calendar.style.tsx
+++ b/src/shared/internal-calendar/internal-calendar.style.tsx
@@ -9,11 +9,6 @@ import { IconButton } from "../../icon-button";
 import { TextStyleHelper } from "../../text";
 import { CalendarType } from "./types";
 
-interface ContainerStyleProps {
-    $type: CalendarType;
-    $isOpen?: boolean | undefined;
-}
-
 interface SideArrowButtonStyleProps {
     $direction: "left" | "right";
 }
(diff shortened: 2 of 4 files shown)


src/shared/internal-calendar/internal-calendar.tsx
```diff
@@ -450,56 +455,63 @@ export const Component = (
...
+        // React spring animation configuration
+        const styles = useSpring({
+            height: isOpen
+                ? resizeDetector.height + 64 // include vertical padding
+                : 0,
+        });
+
+        return <AnimatedDiv style={styles}>{renderContent()}</AnimatedDiv>;
+    } else {
+        return <>{renderContent()}</>;
```
qroll, line 513: a wrapper is still needed for the standalone calendar to set className prop and/or constrain the width. it can be added to `calendar.tsx`

src/shared/internal-calendar/internal-calendar.tsx
```diff
@@ -450,56 +455,63 @@ export const Component = (
...
+                        <ArrowRight />
+                    </SideArrowButton>
+                )}
+            </Container>
+        );
+    };
+
+    if (type === "input") {
+        // React spring animation configuration
+        const styles = useSpring({
```
qroll, line 505: this is a hook so it's not exactly idiomatic to use it conditionally

src/shared/internal-calendar/internal-calendar.style.tsx
```diff
@@ -35,73 +35,55 @@ interface OverlayStyleProps {
 // =============================================================================
 // STYLING
 // =============================================================================
+// -----------------------------------------------------------------------------
+// ICONS
+// -----------------------------------------------------------------------------
+const directionalIconStyle = css`
+    color: ${Color.Neutral[3]};
```
qroll, line 42: can add a default `1rem` width and height here for all icons

src/shared/internal-calendar/internal-calendar.style.tsx
```diff
@@ -190,6 +172,11 @@ export const DropdownButton = styled.button<DropdownButtonStyleProps>`
             `;
         }
     }}
+
+    ${IconChevronDown} {
+        width: 1.125rem;
```
qroll, line 177: this only applies to standalone

src/shared/internal-calendar/internal-calendar.style.tsx
```diff
@@ -35,73 +35,55 @@ interface OverlayStyleProps {
...
             return css`
-                top: calc(100% + 0.5rem);
-                opacity: 1;
-            `;
-        } else if (props.$type === "input" && !props.$isOpen) {
-            return css`
-                height: 24.8rem; // fake height for animation
-                opacity: 0;
-                pointer-events: none;
+                flex-direction: row;
```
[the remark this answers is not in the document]
keithtxw: Yup you are correct. Missed it.

src/calendar/calendar.tsx
```diff
@@ -1,6 +1,21 @@
...
+    );
 };
+
+// =============================================================================
+// STYLING
+// =============================================================================
+const Wrapper = styled.div`
+    width: 41rem;
+    border: 1px solid ${Color.Neutral[5]};
+    border-radius: 8px;
```
qroll, line 20:
```suggestion
    width: 100%;
    max-width: 41rem;
    border: 1px solid ${Color.Neutral[5]};
    border-radius: 0.75rem;
```

qroll, line 20:
```suggestion
    border-radius: 0.75rem;
```

keithtxw: Usually for border radius, we use absolute values to prevent any discrepancies

qroll: ah ok, I thought that convention usually applies to small values like 1-3px only. but yes make sense to preserve border radius

qroll (approved):

PR #151, opened 2023-05-17
Refactor time picker

qroll:
**Changes**

- [delete] branch

Refactoring/updating time picker inputs
- extract shared timepicker dropdown component
- moved timepicker helper to `utils` folder and renamed to `TimeHelper`
- align styles with Figma and DateInput
- added readonly state to the original Timepicker

Other changes
- the indicator bar is now red for error state
- a plain text input is used in various components, have extracted it and placed it together with `InputWrapper`
- updated DateInput and TimeRangePicker to use `InputWrapper`

**Additional information**

- [Figma ref](https://www.figma.com/file/us90jOpWBahsqK2VAluPvD/Flagship-Design-System?type=design&node-id=3375-28945&t=rirLIrfhKJ6L40r5-0)


src/shared/timepicker-dropdown/timepicker-dropdown.tsx
```diff
@@ -399,6 +394,7 @@ export const TimepickerDropdown = ({
             <Container
                 ref={resizeDetector.ref}
                 data-testid={getTestId("timepicker-dropdown")}
+                inert={show ? undefined : ""}
```
qroll, line 397: when set, the `inert` attribute disables interactivity with the element and its children. this includes exclusion from the tab sequence

it's supported in the majority of browsers now but unfortunately [still not included in React](https://github.com/facebook/react/pull/24730), so a workaround is to augment the types and whitelist it from eslint

src/time-range-picker/time-range-picker.tsx
```diff
@@ -150,64 +147,57 @@ export const TimeRangePicker = ({
...
                             : "timepicker-selector"
                     }
                 />
-                <TimeRangePickerDropdown
+                <TimepickerDropdown
                     id={id}
                     show={showEndTimeSelector}
                     value={endTimeVal}
                     format={format}
                     onCancel={handleSelectionDropdownCancel}
```
keithtxw, line 188: We're missing a `onSelectionCancel` callback on the component level

keithtxw (changes requested):

keithtxw:
Can we also change the `TimeRangeInputValue` to just `start` and `end`. And to add jsdocs for `format`, `value` and `onChange`

src/time-range-picker/time-range-picker.tsx
qroll: what's the use of this callback? to detect the closing of the floating dropdown?

it doesn't seem useful for the ranged field since `onChange` only triggers on confirmation.

keithtxw: It was to trigger the error message iirc. else can trigger an `onBlur` instead. If so, to standardise with the single input one.

qroll: was planning to put `onSelectionCancel` back, but found out the cancel event also causes a blur event. instead of managing focus, i just standardised both to call `onBlur`

src/time-range-picker/time-range-picker.tsx
```diff
@@ -31,30 +29,23 @@ export const TimeRangePicker = ({
...
     }, []);
-    useEffect(() => {
-        document.addEventListener("mousedown", handleMouseDownEvent);
-        document.addEventListener("keyup", handleKeyUpEvent);
 
-        return () => {
-            document.removeEventListener("mousedown", handleMouseDownEvent);
-            document.removeEventListener("keyup", handleKeyUpEvent);
-        };
-    }, [showEndTimeSelector]);
```
qroll, line 57: missed out during review. these callbacks have stale state. have also amended similar issues in Timepicker

keithtxw (approved):

PR #160, opened 2023-05-20
Add RangeInputInnerContainer

keithtxw:
**Changes**
Add `RangeInputInnerContainer` component to standardise and manage range inputs with indicators

- Component accepts exactly 2 child components and renders the arrow as well as indicator automatically
- For the split or wrap layout, we're using media query (with option of arbitrary width) for now on with assumption that there would not be any use case whereby
     - You would give a small width for a bigger screen 
- Have amended `TimeRangePicker` and `InputRangeSelect` to use the component. 
- `DateInput` not updated due to the refactor about to happen

- [delete] branch

### DIFF
--- a/src/shared/range-input-inner-container/range-input-inner-container.styles.tsx
+++ b/src/shared/range-input-inner-container/range-input-inner-container.styles.tsx
@@ -70,7 +70,9 @@ export const ArrowRight = styled(ArrowRightIcon)`
 export const Indicator = styled.div<IndicatorStyleProps>`
     position: absolute;
     background: ${(props) =>
-        props.$error ? Color.Validation.Red.Border : Color.Primary};
+        props.$error
+            ? Color.Validation.Red.Border(props)
+            : Color.Primary(props)};
     height: 0.125rem;
     // half - half padding - half icon width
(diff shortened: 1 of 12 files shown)


src/shared/range-input-inner-container/range-input-inner-container.styles.tsx
```diff
@@ -0,0 +1,110 @@
...
+        if (props.$minWidthBeforeWrap) {
+            return css`
+                ${getDynamicMediaWidthQuery(props.$minWidthBeforeWrap)} {
+                    /* Parent container need to provide space */
+                    flex-wrap: wrap;
+
+                    [data-id="range-container-elem1-container"],
+                    [data-id="range-container-elem2-container"] {
+                        // 100% - Icon size - 2padding
+                        max-width: calc(100% - 1.125rem - 1rem);
```
qroll, line 41: this seems to only work when `Wrapper` has set an explicit width like `width: 100%`

keithtxw: Discussed to leave style as it is and amend accordingly when used with `DateRangeInput`

qroll (approved):

PR #153, opened 2023-05-17
[RBS-312][IT] Enhancement calendar layout and adding showBorder, minDate and maxDate prop

ivantps:
**Changes**
- modified the calendar layout with the left/right arrows to be on top right and Month/Year dropdown to be on top left.
- added showBorder, minDate and maxDate prop
- NOT ready to merge

**Additional information**

- You may refer to this (https://jira.ship.gov.sg/browse/RBS-312)

PR #173, opened 2023-05-25
Calendar style and prop update (original behaviour)

keithtxw:
**Changes**
Some adaptations from this [PR](https://github.com/LifeSG/react-design-system/pull/153)

- Update style to have navigation buttons at the top
- Introduce prop to control if the bordered style is rendered
- Update to `minDate` and `maxDate` props from `between`
- **Note: Original behaviour is retained. Update to be dealt with in another PR**

**TODO**
- Update the disabled behaviour as discussed

- [delete] branch

**Changelog entry**

- [BREAKING] Replace `between` to `minDate` and `maxDate`
- Add `styleType` prop to control if a border should appear or not
- Style updates


tests/utils/calendar-helper.spec.ts
```diff
@@ -99,6 +99,87 @@ describe("CalendarHelper", () => {
...
+                date,
+                minDate,
+                maxDate
+            );
+
+            expect(isWithinRange).toBeTruthy();
+        });
+
+        it("should return true if the date is within range (inclusive check)", () => {
+            const date = dayjs("2023-02-01");
```
qroll, line 119: can check for 2023-01-01 also (to assert min date date is correct)

src/shared/internal-calendar/internal-calendar.style.tsx
```diff
@@ -100,20 +95,22 @@ export const OptionsOverlay = styled.div<OverlayStyleProps>`
     }}
 `;
 
-export const Header = styled.div`
+export const Header = styled.div<GeneralStyleProps>`
     display: flex;
     justify-content: space-between;
-    margin-bottom: 0.5rem;
-`;
+    margin: 0 0.5rem 1rem 0.5rem;
```
qroll, line 101: i think Container can deal with the horizontal spacing which is 1.25rem for input and 2rem for standalone, header just needs to deal with the bottom margin

qroll: actually some spacing and styles have deviated from the current Figma. better to address in another ticket.

qroll (changes requested):

qroll:
side note: to update date input and date range input which still use `between` to decide if manual date input is valid

src/shared/internal-calendar/internal-calendar.style.tsx
```diff
@@ -191,6 +188,18 @@ export const HeaderArrowButton = styled(IconButton)`
     :active {
         background: transparent;
     }
+
+    ${(props) => {
+        if (props.$type === "standalone") {
+            return css`
+                ${ArrowLeft},
+                ${ArrowRight} {
+                    height: 1.125rem;
```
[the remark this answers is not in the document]
keithtxw: Oh really? They unified the specs?

src/shared/internal-calendar/internal-calendar.style.tsx
keithtxw: will deal with it in the refactor to have CalendarWrapper

stories/storybook-common/api-table/markup-helpers.tsx
```diff
@@ -0,0 +1,3 @@
+import React from "react";
+
+export const code = (content: string) => <code>{content}</code>;
```
keithtxw, line 3: I think we can put the `quote` function here too

qroll (approved):

PR #178, opened 2023-05-29
[MISC][WK] fix the 2px off from children calendar component width

Wilker0725:
**Changes**
Description of changes...

* Fix calendar component to extend the exact the parent width.

Can reference this issue in [stackoverflow](https://stackoverflow.com/questions/14706866/css-borders-interfering-with-absolute-positioning)

- [delete] branch

Description of change in `InputWrapper`

**Changelog entry**
1. Replace to outline from border css.

### DIFF
--- a/src/shared/internal-calendar/animated-internal-calendar.style.tsx
+++ b/src/shared/internal-calendar/animated-internal-calendar.style.tsx
@@ -6,7 +6,7 @@ export const AnimatedDiv = styled(animated.div)`
     position: absolute;
     top: calc(100% + 0.5rem);
     left: 0;
-    width: 100%;
+    width: calc(100% + 2px);
     max-width: 41rem;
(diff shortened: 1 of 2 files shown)


src/shared/internal-calendar/animated-internal-calendar.style.tsx
```diff
@@ -6,7 +6,7 @@ export const AnimatedDiv = styled(animated.div)`
     position: absolute;
     top: calc(100% + 0.5rem);
     left: 0;
-    width: 100%;
+    width: calc(100% + 2px);
```
qroll, line 9: still need to offset to the left by 1px. because `left: 0` means it still starts within the parent boundary

qroll (changes requested):

qroll (approved):

PR #185, opened 2023-06-05
Allow selection of disabled dates in date inputs

qroll:
**Changes**
Enhancement to the date picker components to allow selection of dates that are normally disabled by minDate, maxDate and disabledDates

- [delete] branch

**Changelog entry**

- Add option to allow selection of disabled dates in `DateInput` and `DateRangeInput`

**Additional information**

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-246)

### DIFF
--- a/src/shared/internal-calendar/internal-calendar-day.style.tsx
+++ b/src/shared/internal-calendar/internal-calendar-day.style.tsx
@@ -8,8 +8,8 @@ import { DayVariant } from "./internal-calendar-day";
 // See more https://styled-components.com/docs/api#transient-props
 // =============================================================================
 export interface StyleProps {
-    $disabled?: boolean;
-    $disabledDisplay?: boolean; // visually disabled but still interactable
+    $disabledDisplay?: boolean;
+    $interactive?: boolean;
     $overlap?: boolean;
     $hovered?: boolean;
     $selected?: boolean;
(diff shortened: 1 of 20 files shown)


src/shared/internal-calendar/internal-calendar-day.style.tsx
```diff
@@ -5,6 +5,7 @@ import { DayVariant } from "./internal-calendar-day";
 
 export interface StyleProps {
     $disabled?: boolean;
+    $disallowed?: boolean;
```
keithtxw, line 8: this could be confusing for someone coming in for the first time. Perhaps `$disabledDisplayOnly` or `$disabledDisplay` might be clearer

same goes to the object key in the status

keithtxw (changes requested):
Amend the name of the key

src/shared/internal-calendar/internal-calendar-day.style.tsx
qroll: since `TimeSlotWeekView` also has some interaction management, I reworked these two style props to be `$interactive` which manages hover effects and `$disabledDisplay` which handles disabled styling

keithtxw:
Approved but to resolve conflicts first

PR #199, opened 2023-06-14
File upload part 2

keithtxw:
**Changes**
- This includes the edit feature. Only image files (refer to `helper.ts` to see what is considered as image files)
- Broke out the list rendering logic into a component of it's own in prep for part 3 (drag reorder)
- Behaviour expectation:
    - Multiple edits at once is fine
    - Adding files while an item is in edit mode is fine
- To ignore 
    - Label sizes on the Textarea as the update to form labels will be handled in another ticket
    - Error banner (to be done in another ticket)

- [delete] branch

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-252)
- [Figma](https://www.figma.com/file/us90jOpWBahsqK2VAluPvD/Flagship-Design-System?type=design&node-id=8137-51564&t=OEgeClHySghxv7oj-0)

Example array that can be added to initialise the fileItems state of the editable file items story 
```
[
                {
                    id: "1",
                    name: "img-1.jpg",
                    type: "image/jpeg",
                    size: 5000,
                    description: "Some text here",
                },
                {
                    id: "2",
                    name: "img-2.jpg",
                    type: "image/jpeg",
                    size: 10000,
                },
                {
                    id: "3",
                    name: "img-3.jpg",
                    type: "image/jpeg",
                    size: 13000,
                },
                {
                    id: "4",
                    name: "img-4.jpg",
                    type: "image/jpeg",
                    size: 5000,
                    description: "Some text here",
                },
                {
                    id: "5",
                    name: "pdf-doc-5.pdf",
                    type: "application/pdf",
                    size: 3000,
                },
                {
                    id: "6",
                    name: "img-6.jpg",
                    type: "image/jpeg",
                    size: 12000,
                },
            ]
```

### DIFF
--- a/src/file-upload/file-item.tsx
+++ b/src/file-upload/file-item.tsx
@@ -1,7 +1,6 @@
 import { BinIcon } from "@lifesg/react-icons/bin";
 import { CrossIcon } from "@lifesg/react-icons/cross";
 import { useEffect, useRef, useState } from "react";
-import { useResizeDetector } from "react-resize-detector";
 import { ProgressBar } from "../shared/progress-bar";
 import { StringHelper } from "../util";
 import {
@@ -46,7 +45,6 @@ export const FileItem = ({
     const isLoading = progress < 1;
 
     const nameSectionRef = useRef<HTMLDivElement>();
-    const { width: wrapperWidth, ref: wrapperRef } = useResizeDetector();
 
     // =========================================================================
     // EFFECTS
@@ -143,7 +141,6 @@ export const FileItem = ({
     return (
         <Item
             id={id}
-            ref={wrapperRef}
             $error={!!errorMessage}
             $loading={isLoading}
             $editable={FileUploadHelper.isSupportedImageType(type)}
--- a/src/file-upload/file-item.tsx
+++ b/src/file-upload/file-item.tsx
@@ -1,5 +1,6 @@
 import { BinIcon } from "@lifesg/react-icons/bin";
 import { CrossIcon } from "@lifesg/react-icons/cross";
+import { PencilIcon } from "@lifesg/react-icons/pencil";
 import { useEffect, useRef, useState } from "react";
 import { ProgressBar } from "../shared/progress-bar";
 import { StringHelper } from "../util";
@@ -19,7 +20,6 @@ import {
 } from "./file-item.styles";
 import { FileUploadHelper } from "./helper";
 import { FileItemProps } from "./types";
-import { PencilIcon } from "@lifesg/react-icons/pencil";
 
 interface Props {
     fileItem: FileItemProps;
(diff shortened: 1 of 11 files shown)


src/file-upload/file-upload.tsx
```diff
@@ -65,23 +99,142 @@ export const FileUpload = ({
...
+                renderMode:
+                    FileUploadHelper.isSupportedImageType(item.type) &&
+                    item.description === undefined
+                        ? "edit"
+                        : "default",
+            };
+        });
+    };
 
+    const getArrangedItemsToRender = (
```
qroll, line 164: what was the consideration for nesting editable items in an array and rendering it within another wrapper container?

src/file-upload/file-item.tsx
```diff
@@ -124,7 +132,7 @@ export const FileItem = ({
                         >
                             <PencilIcon />
                         </IconButton>
-                    )} */}
+                    )}
```
qroll, line 135: should set the usual `aria-label` and `aria-hidden` attributes for the edit button

src/file-upload/file-upload.tsx
keithtxw: There was a ui requirement that I could not achieve with css child flags, and find it cumbersome to use props. Grouping allowed me to do that

src/file-upload/file-list.tsx
```diff
@@ -0,0 +1,237 @@
...
+        setArrangedItems(getArrangedItemsToRender(itemsRenderMode));
+    }, [itemsRenderMode]);
+
+    // =========================================================================
+    // EVENT HANDLERS
+    // =========================================================================
+    const handleDescriptionUpdate =
+        (item: FileItemProps) => (description: string) => {
+            const updatedItem = { ...item, description };
+            onItemUpdate(updatedItem);
```
qroll, line 70: can the item render mode be immediately updated to `list` here if description is not undefined?

src/file-upload/file-list.tsx
```diff
@@ -0,0 +1,237 @@
...
+        };
+
+    const handleCancel = (item: FileItemProps) => () => {
+        setItemsRenderMode((prevState) => {
+            return prevState.map((renderItem) => {
+                if (isEqual(item, renderItem.fileItem)) {
```
qroll, line 76: comparing the file item id should be sufficient

src/file-upload/file-list.tsx
```diff
@@ -0,0 +1,237 @@
...
+    };
+
+    const getItemsRenderMode = (
+        fileItems: FileItemProps[]
+    ): FileItemRenderProps[] => {
+        if (!fileItems || fileItems.length === 0) return [];
+
+        return fileItems.map((item) => {
+            const existingIndex = findIndex(itemsRenderMode, (renderedItem) => {
+                return isEqual(renderedItem.fileItem, item);
```
qroll, line 130: should compare the file item id here, otherwise a description change would also count as a new item. for existing rendered items I think it could just retain the current `renderMode`

src/file-upload/file-list.tsx
keithtxw: as i am relying on the user of the component to assign an id, i cant really be certain if they are doing a proper job in doing so. Hence comparing the whole object might be safer. Unless we cant trust that the user of the component doesn't screw up. haha

src/file-upload/file-list.tsx
keithtxw: Actually i am looking to rewrite this. Let me refactor

qroll (approved):

PR #208, opened 2023-06-22
Add min width for animated internal calendar

afiq0209:
**Changes**
- Add min-width in animated calendar so that the date input & date range input calendar does not get cut off
- Add default data-testid

- [delete] branch

**Additional information**

- You may refer to this [BOOKINGSG-4349](https://jira.ship.gov.sg/browse/BOOKINGSG-4349)

### DIFF
--- a/src/shared/internal-calendar/calendar-manager.tsx
+++ b/src/shared/internal-calendar/calendar-manager.tsx
@@ -355,7 +355,7 @@ const Component = (
 
     const renderHeader = () => {
         return (
-            <Header data-id="calendar-header">
+            <Header data-testid="calendar-header">
                 <HeaderInputDropdown>
                     {renderDropdownButtons()}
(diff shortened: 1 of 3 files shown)


src/shared/internal-calendar/calendar-manager.tsx
```diff
@@ -355,7 +355,7 @@ const Component = (
 
     const renderHeader = () => {
         return (
-            <Header data-id="calendar-header">
+            <Header data-testid="calendar-header">
```
qroll, line 358: `data-id` is used to target certain child elements when styling with styled-components. example is `time-slot-week-view` which has custom padding for the header

if `data-testid` is still needed, can set it as an additional prop instead

qroll (changes requested):

qroll (approved):

PR #210, opened 2023-06-22
File Upload Part 3 - Drag sorting

keithtxw:
**Changes**
- In this part, we implement the drag sorting behaviour of the file items
- Notable info
    - Drag sorting is enabled only if there are no items in edit mode
    - While dragging, all items including the upload button is disabled

- [delete] branch

**Additional information**

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-254)
- [Figma](https://www.figma.com/file/us90jOpWBahsqK2VAluPvD/Flagship-Design-System?type=design&node-id=8475-54069&mode=design&t=hP3MU2zKNTpJhbND-0)


src/file-upload/file-upload.tsx
```diff
@@ -26,16 +27,19 @@ export const FileUpload = ({
     capture,
     multiple,
     disabled,
+    sortable = false,
     descriptionMaxLength,
-    editableFileItems = true,
+    editableFileItems = false,
```
keithtxw, line 32: Selig mentioned that it is false by default

src/file-upload/file-upload-context.tsx
keithtxw, line 1: Use context to avoid having to pass props around particularly for the drag info

src/file-upload/file-list.tsx
```diff
@@ -32,28 +49,41 @@ export const FileList = ({
...
 
+    /**
+     * As the default drag sensors interfere with click events
+     * on the file items, we'll need to configure the sensor to
+     * only activate the drag if the mouse moves a certain distance
+     */
+    const sensors = useSensors(
+        useSensor(PointerSensor, {
+            activationConstraint: {
+                distance: 8, // mouse drag of 8px then activate the drag event
```
keithtxw, line 75: Arbitrary value

src/file-upload/file-list.tsx
```diff
@@ -83,6 +120,36 @@ export const FileList = ({
         onItemDelete(item);
     };
 
+    const handleDragEnd = (event: DragEndEvent) => {
+        if (onReorder) {
+            const { active, over } = event;
+
+            if (active.id !== over.id) {
```
qroll, line 127: to check `over` is defined. found if the item is dragged outside of the parent boundary, it's set to null

src/file-upload/file-item.styles.tsx
```diff
@@ -125,8 +175,7 @@ export const LoadingActionContainer = styled(ItemActionContainer)`
     }
 `;
 
-// TODO: Replace with new IconButton when ready
-export const IconButton = styled(Button.Small)`
+export const IconButton = styled(DSIconButton)`
```
qroll, line 178: no need to set height/width any more, can specify the `sizeType` prop as small

qroll (changes requested):

src/file-upload/file-item.tsx
```diff
@@ -152,37 +179,55 @@ export const FileItem = ({
...
-            $error={!!errorMessage}
-            $loading={isLoading}
-            $editable={FileUploadHelper.isSupportedImageType(type)}
+            ref={setNodeRef}
+            $sortable={sortable}
+            $disabled={disabled}
+            $focus={isFocused}
+            $focusOther={isFocusedOthers}
+            style={style}
+            {...attributes}
```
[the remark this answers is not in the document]
keithtxw: This is a good point. I missed this out

src/file-upload/file-list.tsx
keithtxw: But nothing actually happens yes?

src/file-upload/file-item.styles.tsx
keithtxw: ah ok. i think because i started this before IconButton was updated. Thanks for pointing

src/file-upload/file-list.tsx
qroll: nothing happens because a TypeError is thrown 🙃

src/file-upload/file-list.tsx
```diff
@@ -169,28 +172,30 @@ export const FileList = ({
...
-                 * description value.
-                 *
-                 * If editing was cancelled previously, we will leave it as
-                 * it is and not render the edit display
-                 */
-                updatedRenderModes[item.id] = shouldRenderEditMode(item)
+                newRenderModes[item.id] = renderModes[item.id];
+            } else if (item.errorMessage) {
+                newRenderModes[item.id] = "error";
+            } else if (item.progress && item.progress < 1) {
```
qroll, line 189: progress = 0 is also falsy, can check for undefined/null or `item.progress === "number"` instead

qroll:
on testing, noticed a couple of interaction conflicts
- keyboard space on action button activates drag mode instead of triggering the action. workaround suggested [here](https://github.com/clauderic/dnd-kit/issues/710#issuecomment-1125318163) is to stop the keydown event propagation
- when initiating drag on action button, sometimes the component remains stuck in drag mode
  - to replicate, try targeting the button border rather than the icon svg
  - if I set `const shouldDisableActions = false` it works fine, so I suspect the drag end event isn't triggering for disabled buttons

src/file-upload/file-list.tsx
keithtxw: For this, actually i am just intending to check if progress is specified and it is less than 1. I would think 
```
item.progess < 1 
```
would suffice since `progress === undefined` will resolve to false

PR #220, opened 2023-06-30
File upload part 5 - edit interactions

keithtxw:
**Changes**
- Amend "Save" and "Cancel" behaviours in edit mode
- **NOTE**: Dependent [PR](https://github.com/LifeSG/react-design-system/pull/216) has to be merged before this can be

- [delete] branch

**Additional information**

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-280)
- [Figma](https://www.figma.com/file/us90jOpWBahsqK2VAluPvD/Flagship-Design-System?type=design&node-id=8475-54069&mode=design&t=hP3MU2zKNTpJhbND-0)

### DIFF
--- a/src/file-upload/file-item-edit.tsx
+++ b/src/file-upload/file-item-edit.tsx
@@ -19,16 +19,18 @@ interface Props {
     fileItem: FileItemProps;
     wrapperWidth: number;
     descriptionMaxLength?: number | undefined;
-    onEdit: (description: string) => void;
-    onCancel: (description: string) => void;
+    onSave: (description: string) => void;
+    onCancel: () => void;
+    onBlur: (value: string) => void;
 }
 
 const Component = ({
     fileItem,
     descriptionMaxLength,
     wrapperWidth,
-    onEdit,
+    onSave,
     onCancel,
+    onBlur,
 }: Props) => {
     // =========================================================================
     // CONST, STATE, REFS
@@ -43,9 +45,7 @@ const Component = ({
     } = fileItem;
 
     const [formattedName, setFormattedName] = useState<string>();
-    const [currentDescription, setCurrentDescription] = useState<string>(
-        description || ""
-    );
+    const [currentDescription, setCurrentDescription] = useState<string>("");
 
     const textareaRef = useRef<HTMLTextAreaElement>();
     const nameSectionRef = useRef<HTMLDivElement>();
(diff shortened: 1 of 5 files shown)


src/file-upload/file-item-edit.tsx
```diff
@@ -138,11 +148,13 @@ export const FileItemEdit = ({
...
+                    onClick={onCancel}
                 >
                     Cancel
                 </ActionButton>
             </ActionButtonsSection>
         </Item>
     );
 };
+
+export const FileItemEdit = memo(Component);
```
qroll, line 160: what's the intention behind memoizing the list items? specifically for `FileItemEdit`, the memoization does not help as a new `fileItem` object reference is created on each render

src/file-upload/file-list.tsx
```diff
@@ -97,26 +107,30 @@ export const FileList = ({
...
+        descriptionsValueRef.current[item.id] = value;
     };
 
-    const handleInitiateEdit = (item: FileItemProps) => () => {
-        const updatedRenderModes = { ...renderModes };
-        updatedRenderModes[item.id] = "edit";
+    const handleCancel = (item: FileItemProps) => () => {
+        if (!item.description || item.description.length === 0) {
+            // New addition
+            onItemDelete(item);
```
qroll, line 125: `removeDescription` should also be run here, as the item id might have been set previously

src/file-upload/file-list.tsx
```diff
@@ -175,29 +189,29 @@ export const FileList = ({
...
-                renderModes[item.id] === "cancelled-edit"
-            ) {
-                newRenderModes[item.id] = renderModes[item.id];
-            } else if (item.errorMessage) {
+            if (item.errorMessage) {
                 newRenderModes[item.id] = "error";
-            } else if (item.progress < 1) {
-                newRenderModes[item.id] = "loading";
-            } else {
+            } else if (!renderModes[item.id]) {
```
qroll, line 194: does `shouldRenderEditMode` need to check for empty description now? otherwise the `onItemDelete` behaviour for new items only kicks in when the user explicitly clicks edit+cancel, which feels confusing

qroll (changes requested):

src/file-upload/file-item-edit.tsx
keithtxw: I think this was done previously when I was testing. Will revert.

src/file-upload/file-list.tsx
keithtxw: It will be unlikely so but i think I should do so given it could be altered by the parent.

src/file-upload/file-list.tsx
keithtxw: As discussed, added check for undefined and empty string. Also prevent empty string from being able to be saved

src/file-upload/file-item-edit.tsx
```diff
@@ -54,19 +56,23 @@ export const FileItemEdit = ({
...
-        if (textareaRef.current) {
-            onEdit(textareaRef.current.value || "");
-        } else {
-            onEdit(undefined);
-        }
+        onSave(textareaRef.current.value);
+    };
+
+    const handleChange = (event: React.ChangeEvent<HTMLTextAreaElement>) => {
+        setCurrentDescription(event.target.value.trim());
```
qroll, line 71: trimming in onChange can result in some funny behaviour, I think the trim could be done on save + in the disabled check instead

https://github.com/LifeSG/react-design-system/assets/16853062/2647bc73-1119-4999-9229-9a6638e3b9b6

keithtxw: Yea i think typing space in between text could also be an issue. Should have figured. Careless. Yes i think trimming on save will be a better way. For the disabled check, i think we can check for empty values explicitly

qroll (approved):

PR #224, opened 2023-07-03
Allow Accordion items to be collapsed on first load

chiakaiming:
**Changes**
- Allow Accordion items to be collapsed on first load

- [delete] branch

**Changelog entry**

- Add ability to collapse or expand all items on render

**Additional information**
- You may refer to this [MOL-13135](https://jira.ship.gov.sg/browse/MOL-13135)

### DIFF
--- a/src/accordion/accordion.tsx
+++ b/src/accordion/accordion.tsx
@@ -13,10 +13,11 @@ const AccordionBase = ({
     children,
     title,
     enableExpandAll = true,
+    initialExpandAll = true,
     showTitleInMobile = false,
     className,
 }: AccordionProps): JSX.Element => {
-    const [expandAll, setExpandAll] = useState<boolean>(true);
+    const [expandAll, setExpandAll] = useState<boolean>(initialExpandAll);
 
     const handleExpandCollapseClick = (event: React.MouseEvent) => {
         event.preventDefault();
--- a/src/accordion/accordion.tsx
+++ b/src/accordion/accordion.tsx
@@ -13,11 +13,13 @@ const AccordionBase = ({
     children,
     title,
     enableExpandAll = true,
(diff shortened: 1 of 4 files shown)


src/accordion/types.ts
```diff
@@ -2,6 +2,7 @@ export interface AccordionProps {
     children: JSX.Element | JSX.Element[];
     title?: string | undefined;
     enableExpandAll?: boolean | undefined;
+    initialExpandAll?: boolean | undefined;
```
keithtxw, line 5:
```suggestion
    expandAll?: boolean | undefined;
```

src/accordion/accordion.tsx
```diff
@@ -13,10 +13,11 @@ const AccordionBase = ({
     children,
     title,
     enableExpandAll = true,
+    initialExpandAll = true,
     showTitleInMobile = false,
     className,
 }: AccordionProps): JSX.Element => {
-    const [expandAll, setExpandAll] = useState<boolean>(true);
+    const [expandAll, setExpandAll] = useState<boolean>(initialExpandAll);
```
keithtxw, line 20: We've decided to allow the expandAll props to be passed down. Hence for this to rename state name to `expandAllState` and also add a `useEffect` for any `expandAll` prop changes. 

Do remember to document this new prop also.

keithtxw (changes requested):

src/accordion/accordion.tsx
keithtxw: Sorry after thinking it carefully, we'll forget about the `useEffect`. Let's just use the new suggested prop name and leave everything as it is. Still remember to document the prop

keithtxw (approved):

PR #226, opened 2023-07-05
File upload part 6 misc

keithtxw:
**Changes**
- Error display (component level)
- Readonly display
- Bugs (e.g. progress bar in mobile)
- Documentation update

- [delete] branch

**Additional information**

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-284)

### DIFF
--- a/src/file-upload/file-list-item/file-list-item.tsx
+++ b/src/file-upload/file-list-item/file-list-item.tsx
@@ -134,6 +134,8 @@ const Component = ({
 
     const shouldDisable = () => disabled || !!activeId;
 
+    const shouldEnableSort = () => sortable && !readOnly;
+
     // =========================================================================
     // RENDER FUNCTIONS
     // =========================================================================
@@ -286,7 +288,9 @@ const Component = ({
             $focusType={focusType}
             {...(sortable ? sortableProps : {})}
         >
-            {sortable && <DragHandleIcon $disabled={shouldDisable()} />}
+            {shouldEnableSort() && (
+                <DragHandleIcon $disabled={shouldDisable()} />
+            )}
             <Box
                 $focused={focusType === "self"}
                 $error={!!errorMessage}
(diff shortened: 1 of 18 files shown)


src/file-upload/file-list.tsx
```diff
@@ -178,7 +180,13 @@ export const FileList = ({
     };
 
     const shouldRenderEditMode = (item: FileItemProps) => {
-        return checkEditable(item) && !item.description;
+        return (
+            !item.errorMessage &&
+            !readOnly &&
+            !(item.progress && item.progress < 1) &&
```
qroll, line 186: handle other falsy values
```suggestion
            !(item.progress < 1) &&
```

src/file-upload/file-list-item/file-list-item.tsx
qroll, line 1: the `sortable`-related props passed to `<Item />` should also check the readonly state. `file-list.tsx` already has a `shouldEnableSort()` so I think the check can be included there

qroll (changes requested):

stories/file-upload/props-table.tsx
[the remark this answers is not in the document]
keithtxw: Only concern is that i am not inheriting all the props. Hence the explicit specification.

src/file-upload/file-list.tsx
keithtxw: This will allow value like `null` to be assessed. Since we are only expecting numbers, will change to something more robust like a type check

stories/file-upload/props-table.tsx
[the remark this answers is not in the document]
qroll: I think the explicit specification can stay but on the documentation side we should point the dev to the format of `accept` and `capture`. `multiple` is ok since it's self-explanatory.

qroll (approved):

PR #233, opened 2023-07-07
Preserve text caret position in formatted input fields

qroll:
**Changes**

- Add logic in `onChange` handlers to calculate the next text caret position
  - The position is offset by the number of characters that are removed and added, which can be derived from the difference between the length of the original input and the final formatted input
  - Not perfect e.g. with `(12) ┃345`, typing space results in `(12┃) 345` but keeping it simple to avoid handling all the edge cases
- Misc fix: pass refs down to the main input field in `InputGroup` variants
- [delete] branch

**Changelog entry**

Bug fix
- Prevent text caret jumping to end when typing in the middle of input in `PhoneNumberInput` and `UnitNumberInput`

### DIFF
--- a/src/phone-number-input/phone-number-input.tsx
+++ b/src/phone-number-input/phone-number-input.tsx
@@ -70,10 +70,10 @@ export const PhoneNumberInput = ({
         }
     };
 
-    const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-        const { nextValue, updateCursorPosition } = getNextInputState();
+    const handleInputChange = () => {
+        const { nextValue, updateCaretPosition } = getNextInputState();
 
-        updateCursorPosition();
+        updateCaretPosition();
         performLocalChangeHandler(nextValue, selectedCountry);
         if (onChange) {
             performOnChangeHandler(nextValue, selectedCountry);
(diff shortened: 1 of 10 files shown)


src/phone-number-input/phone-number-input.tsx
weili-govtech, line 1: notice a minor quirk:
1. I have a phone number with parenthesis at the beginning `(12) 345`
2. my text caret position is at 1 `(1┃2) 345`
3. I press backspace
4. text caret position goes to 0, before the parenthesis `┃(23) 45`

weili-govtech:
I think it's worthwhile to have unit tests covering these behaviours, it's likely to be missed by anyone else updating the components

src/phone-number-input/phone-number-input.tsx
qroll: after some consideration, this would need to track the previous caret position and the input type (i.e. insertion or deletion). I found some useful properties from events like `selectionchange` and `beforeinput` but there's additional handling required. for now, perhaps we can stick to the current behaviour to reduce complexity

keithtxw (approved):

weili-govtech (approved):

PR #234, opened 2023-07-07
onDismiss for Toast

monK87:
**Changes**
Adding `onDismiss` callback to the `Toast` component.

- [delete] branch

**Changelog entry**

- Adding `onDismiss` callback to the `Toast` component.

### DIFF
--- a/src/toast/toast.tsx
+++ b/src/toast/toast.tsx
@@ -47,7 +47,6 @@ export const Toast = ({
 
         setTimeout(() => {
             setVisible(false);
-            onDismiss?.();
         }, timer);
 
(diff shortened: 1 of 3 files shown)


src/toast/toast.tsx
qroll, line 1: ideally `onDismiss` should wait for the animation to end before firing. `useSpring` accepts a `onRest` callback that fires when the animation stops, can try that

qroll (changes requested):

src/toast/toast.tsx
monK87: moved the callback to onRest

qroll (approved):

PR #245, opened 2023-07-17
Remove the default value of expanded and only allow expanded prop to set initial state for Accordion.Item

chiakaiming:
**Changes**
Remove the default value of expanded in `Accordion.Item` to allow expandAll === false from parent
`Accordion.Item` height will not use animation on first load (to allow easier scrollHeight calculation)

Add handles to allow expand/collapse of individual items. 

- [delete] branch

**Changelog entry**

- [WARNING] Remove the default value of expanded in `Accordion.Item`

**Additional information**

- You may refer to this[MOL-13135](https://jira.ship.gov.sg/browse/MOL-13135)

### DIFF
--- a/src/accordion/accordion-item.tsx
+++ b/src/accordion/accordion-item.tsx
@@ -1,9 +1,9 @@
-import { isFunction } from "lodash";
 import React, {
     forwardRef,
     useContext,
     useEffect,
     useImperativeHandle,
+    useRef,
     useState,
 } from "react";
 import { useResizeDetector } from "react-resize-detector";
@@ -29,27 +29,12 @@ function Component(
         type = "default",
         ...otherProps
     }: AccordionItemProps,
-    ref: React.Ref<HTMLDivElement>
+    ref: React.Ref<AccordionItemHandle>
 ) {
-    useImperativeHandle(
-        ref,
-        () => {
-            if (!isFunction(ref)) {
-                return Object.assign(ref.current, {
-                    expand(): void {
-                        setExpand(true);
-                    },
-                    collapse(): void {
-                        setExpand(false);
-                    },
-                });
-            }
-        },
-        []
-    );
     // =============================================================================
     // CONST, STATE, REF
     // =============================================================================
+    const elementRef = useRef<HTMLDivElement>();
     const expandAll = useContext(AccordionContext);
     const [expand, setExpand] = useState<boolean>(expanded ?? expandAll);
     const [hasFirstLoad, setHasFirstLoad] = useState<boolean>(false);
@@ -58,6 +43,20 @@ function Component(
 
     const resizeDetector = useResizeDetector();
     const childRef = resizeDetector.ref;
+
+    useImperativeHandle(
+        ref,
+        () =>
+            Object.assign(elementRef.current, {
+                expand(): void {
+                    setExpand(true);
+                },
+                collapse(): void {
+                    setExpand(false);
+                },
+            }),
+        []
+    );
 
     // =============================================================================
     // EFFECTS
@@ -137,7 +136,7 @@ function Component(
             data-testid={testId}
             className={otherProps.className}
             $isCollapsed={expand}
-            ref={ref}
+            ref={elementRef}
         >
             <TitleContainer>
                 {renderTitle()}
--- a/src/accordion/accordion-item.tsx
+++ b/src/accordion/accordion-item.tsx
@@ -17,7 +17,7 @@ import { AccordionItemProps } from "./types";
 export const AccordionItem = ({
     title,
     children,
-    expanded = true,
+    expanded,
     type = "default",
     ...otherProps
 }: AccordionItemProps) => {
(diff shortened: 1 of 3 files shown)


src/accordion/accordion-item.tsx
```diff
@@ -37,13 +38,14 @@ export const AccordionItem = ({
     // =============================================================================
 
     useEffect(() => {
-        setExpand(expanded);
-    }, [expanded]);
```
keithtxw, line 41: If you remove this, we cannot expand it mid way already. This should still be there.

src/accordion/types.ts
```diff
@@ -12,6 +12,9 @@ export type AccordionItemType = "default" | "small";
 export interface AccordionItemProps {
     title: string;
     children: JSX.Element | JSX.Element[];
+    /**
+     * Only for initial state, do not use to programmatically expand/collapse Accordion.Item after first load
```
keithtxw, line 16: The `expanded` prop is used for controlling the states for each item though. It is not restricted to the initial load only

src/accordion/accordion-item.tsx
keithtxw: Instead since the purpose of this is to expand it on initial load if the `expandAll` is otherwise, then we should do this
```
 useEffect(() => {
        if (hasFirstLoad) setExpand(expanded);
    }, [expanded]);
```

src/accordion/accordion-item.tsx
```diff
@@ -17,15 +17,16 @@ import { AccordionItemProps } from "./types";
...
     type = "default",
     ...otherProps
 }: AccordionItemProps) => {
     // =============================================================================
     // CONST, STATE, REF
     // =============================================================================
     const expandAll = useContext(AccordionContext);
-    const [expand, setExpand] = useState<boolean>(expandAll ?? expanded);
+    const [expand, setExpand] = useState<boolean>(expanded ?? expandAll);
+    const [hadFirstLoad, setHadFirstLoad] = useState<boolean>(false);
```
keithtxw, line 29: prefer to use present tense. So `hasFirstLoad` is more ideal

keithtxw (changes requested):
Think the code snippet I left in our conver would solve your requirements? Unless there is another use case I am missing.

src/accordion/accordion-item.tsx
```diff
@@ -58,14 +66,13 @@ export const AccordionItem = ({
...
-        height: expand ? resizeDetector.height : 0,
-    });
+    const resizeHeight = { height: expand ? resizeDetector.height : 0 };
+    const expandableStyles = useSpring(resizeHeight);
 
     const renderContent = () => {
         return (
             <Expandable
-                style={expandableStyles}
+                style={hasFirstLoad ? expandableStyles : resizeHeight}
```
keithtxw, line 75: I think this is not necessary

keithtxw (changes requested):

src/accordion/accordion-item.tsx
chiakaiming: the animation affects how much we are able to scroll due to page height, and we need to scroll on page load

keithtxw (changes requested):

qroll (changes requested):

stories/accordion/props-table.tsx
```diff
@@ -84,9 +84,9 @@ const ACCORDION_ITEM_DATA: ApiTableSectionProps[] = [
             },
             {
                 name: "expanded",
-                description: "Specifies if the item is expanded",
+                description:
+                    "Specifies if the item is expanded. Omit or set this to undefined to allow 'Hide All'/'Show All' in parent to take precedence for cases where we want the default expand behaviour",
```
qroll, line 88: for clarity

```
                    <>
                        Specifies if the item is expanded. When set, this takes
                        precedence over {code("initialDisplay")} of the
                        {code("Accordion")}
                    </>
```

src/accordion/accordion-item.tsx
```diff
@@ -12,20 +19,40 @@ import {
...
+        ...otherProps
+    }: AccordionItemProps,
+    ref: React.Ref<HTMLDivElement>
+) {
+    useImperativeHandle(
+        ref,
+        () => {
+            if (!isFunction(ref)) {
+                return Object.assign(ref.current, {
+                    expand(): void {
```
keithtxw, line 39: Can just return the functions. Don't need the check and assignment of object

keithtxw (changes requested):

src/accordion/accordion-item.tsx
```diff
@@ -25,6 +31,22 @@ function _AccordionItem(
     }: AccordionItemProps,
     ref: React.Ref<HTMLDivElement>
 ) {
+    useImperativeHandle(
+        ref,
+        () => {
+            if (!isFunction(ref)) {
```
qroll, line 37: this check here would mean ref callbacks won't work though

if the underlying div element isn't needed, then this will be sufficient

```
    useImperativeHandle(
        ref,
        () => ({
            expand(): void {
                setExpand(true);
            },
            collapse(): void {
                setExpand(false);
            },
        }),
        []
    );
```

if you do need the reference to the div, tracking an internal element ref could work
```
    const elementRef = useRef<HTMLDivElement>(); // this is passed to the container
    useImperativeHandle(ref, () => {
        return {
            ...elementRef.current,
            expand(): void {
                setExpand(true);
            },
            collapse(): void {
                setExpand(false);
            },
        };
    });
```

chiakaiming: need the ref to div, had to use Object.assign instead of the spread operator

qroll (approved):

keithtxw (approved):

PR #249, opened 2023-07-18
Update RangeSelect component

ivantps:
**Changes**
Update RangeSelect component

- delete branch

**Additional information**

- You may refer to this [RBS-512](https://jira.ship.gov.sg/browse/RBS-512)
- [Figma](https://www.figma.com/file/us90jOpWBahsqK2VAluPvD/Flagship-Design-System?node-id=13151%3A119501&mode=dev)

### DIFF
--- a/src/input-range-select/input-range-select.tsx
+++ b/src/input-range-select/input-range-select.tsx
@@ -5,16 +5,14 @@ import {
     Divider,
     LabelContainer,
     PlaceholderLabel,
-    Selector,
     ValueLabel,
 } from "../shared/dropdown-wrapper/dropdown-wrapper.styles";
 import { RangeInputInnerContainer } from "../shared/range-input-inner-container";
 import { StringHelper } from "../util/string-helper";
 import { InputRangeSelectProps } from "./types";
-import { ClearContainer, ClearIcon } from "../input/input.style";
+import { ClearIcon } from "../input/input.style";
 import {
     ClearIconContainer,
-    EmptyIconWrapper,
     InputSelectorWrapper,
     SelectorDropdown,
 } from "./input-range-select.style";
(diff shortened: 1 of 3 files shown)


src/input-range-select/input-range-select.tsx
```diff
@@ -291,6 +299,15 @@ export const InputRangeSelect = <T, V>({
                     {renderSelectorContent("from")}
                     {renderSelectorContent("to")}
                 </RangeInputInnerContainer>
+                <ClearContainer onClick={handleClear} type="button">
```
qroll, line 302: 1. should check for readOnly and disabled states as well
2. the conditional rendering logic should include the ClearContainer, otherwise there is still an invisible button that can clicked/tabbed to
3. does this component always need to show the clear button? if not, you can introduce a prop called `allowClear`

src/input-range-select/input-range-select.tsx
```diff
@@ -134,6 +136,12 @@ export const InputRangeSelect = <T, V>({
         }
     };
 
+    const handleClear = (event: { stopPropagation: () => void }) => {
```
qroll, line 139: use the inbuilt React event type

```suggestion
    const handleClear = (event: React.MouseEvent) => {
```

src/input-range-select/input-range-select.tsx
```diff
@@ -134,6 +136,12 @@ export const InputRangeSelect = <T, V>({
...
+        setSelectedFromValue(undefined);
+        setSelectedToValue(undefined);
```
qroll, line 141: you should also fire the `onSelectOption` callback prop here to indicate that the selection has been updated

qroll (changes requested):

src/input-range-select/input-range-select.tsx
ivantps: 3. yes this component always need to show the clear button.

qroll:
hmm the diff is reflecting additional commits which isn't ideal. could you rebase your changes or cherry-pick them on top of the latest master instead?

src/input-range-select/input-range-select.tsx
```diff
@@ -134,6 +136,13 @@ export const InputRangeSelect = <T, V>({
         }
     };
 
+    const handleClear = (event: React.MouseEvent) => {
+        event.stopPropagation();
+        setSelectedFromValue(undefined);
+        setSelectedToValue(undefined);
+        onSelectOption(undefined, undefined);
```
qroll, line 143: based on the types the first param should be an object containing the from/to properties

`onSelectOption` is also optional so it's best to check if it is defined before calling it

```suggestion
        onSelectOption({ from: undefined, to: undefined }, undefined);
```

src/input-range-select/input-range-select.tsx
```diff
@@ -291,6 +300,17 @@ export const InputRangeSelect = <T, V>({
                     {renderSelectorContent("from")}
                     {renderSelectorContent("to")}
                 </RangeInputInnerContainer>
+                {focusedInput === "none" &&
```
qroll, line 303: missed this earlier, but there's a `validateDOMNesting(...): <button> cannot appear as a descendant of <button>.` warning in the console

since the Selector is already a button, the clear button has to be moved out and rendered outside of it

qroll (changes requested):

src/input-range-select/input-range-select.tsx
qroll: noticed the addition of a `InputSelectorWrapper`, but the clear icon hasn't been moved so the warning is still present

qroll:
cleaned up the commit history and made some adjustments to the layout to increase touch targets

PR #253, opened 2023-07-19
Fix popover rendering out of screen

ijasm:
**Changes**
1. Fixed popover rendering out of the top of the screen.
2. Increase z-index so that it renders on top of other elements.

- Delete branch

**Additional information**

- You may refer to the screenshot attached to this [ticket](https://github.com/LifeSG/react-design-system/pull/251) for both issues.

### DIFF
--- a/src/popover/popover.styles.tsx
+++ b/src/popover/popover.styles.tsx
@@ -29,7 +29,7 @@ const getVisibilityStyle = (visible: boolean) => {
             visibility: visible;
             opacity: 1;
             transition: ${Transition.Base};
-            z-index: 1000;
+            z-index: 2;
         `;
     } else {
         return css`
--- a/src/popover/popover.styles.tsx
+++ b/src/popover/popover.styles.tsx
@@ -29,7 +29,7 @@ const getVisibilityStyle = (visible: boolean) => {
             visibility: visible;
             opacity: 1;
             transition: ${Transition.Base};
-            z-index: 2;
+            z-index: 50;
         `;
     } else {
         return css`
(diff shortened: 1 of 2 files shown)


src/popover/popover.styles.tsx
```diff
@@ -29,7 +29,7 @@ const getVisibilityStyle = (visible: boolean) => {
             visibility: visible;
             opacity: 1;
             transition: ${Transition.Base};
-            z-index: 2;
+            z-index: 1000;
```
qroll, line 32: would prefer a smaller value like 50 to avoid inflating z-index numbers

qroll (changes requested):

src/popover/popover.styles.tsx
ijasm: sorry i think this change is not required. i have reverted it. will create another PR and use a smaller value if the z-index needs to be updated

qroll: hmm I did notice that the popover is indeed covered by the navbar in certain cases (when the content is short enough to be rendered above but also tall enough to overlap with the navbar) so this is still good to have

ijasm: ya but i was thinking of lowering the z-index of the navbar instead, currently its at 999 on mylegacy, not sure why its so inflated. im not certain what layer the navbar needs to be at but if it needs to be higher than 1, that means the z-index of the popover still has to be higher than 2... i suppose i will bump the popover z-index to 50 to be safe, thanks!

qroll (approved):

PR #260, opened 2023-07-25
[MISC][DH] add `autoDismissTime` prop to Toast component to customize…

monK87:
**Changes**
Added optional `autoDismissTime` prop to `Toast` component to customize the auto dismiss timing. Defaults to 4 seconds as previously.

- [delete] branch

**Changelog entry**

- Added optional `autoDismissTime` prop to `Toast` component to customize the auto dismiss timing

### DIFF
--- a/src/toast/toast.tsx
+++ b/src/toast/toast.tsx
@@ -19,11 +19,14 @@ import {
 } from "./toast.styles";
 import { ToastProps } from "./types";
 
+const defaultAutoDismissTime = 4000;
+
 export const Toast = ({
     type = "success",
     title,
     label,
     autoDismiss,
+    autoDismissTime = defaultAutoDismissTime,
     onDismiss,
     ...otherProps
 }: ToastProps) => {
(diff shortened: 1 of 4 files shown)


stories/toast/props-table.tsx
```diff
@@ -58,12 +58,25 @@ const DATA: ApiTableSectionProps[] = [
                 description: (
                     <>
                         Specifies if the <code>Toast</code> will be
-                        automatically dismiss after <strong>4 seconds</strong>
+                        automatically dismissed after{" "}
+                        <strong>autoDismissTime</strong>.
```
qroll, line 62:
```suggestion
                        <code>autoDismissTime</code>.
```

src/toast/toast.tsx
```diff
@@ -19,11 +19,14 @@ import {
 } from "./toast.styles";
 import { ToastProps } from "./types";
 
+const defaultAutoDismissTime = 4000;
```
qroll, line 22: bit nitpicky but can declare this const in uppercase
```suggestion
const DEFAULT_AUTO_DISMISS_TIME = 4000;
```

qroll (changes requested):
thanks for fixing on the timer id in the useEffect! ⚡

qroll (approved):

PR #263, opened 2023-07-27
Cater for searching secondary label in lowercase

AnuMutyala:
**Changes**

Convert secondary label to lowercase before matching against search input

- [delete] branch

**Changelog entry**

Bug fix
- Update search logic in `InputSelect`, `InputMultiSelect` and `InputGroup` to be case-insensitive for secondary label

### DIFF
--- a/src/shared/dropdown-list/dropdown-list.tsx
+++ b/src/shared/dropdown-list/dropdown-list.tsx
@@ -205,9 +205,16 @@ export const DropdownList = <T, V>({
                     typeof label === "object"
                         ? label.title.toLowerCase()
                         : label.toLowerCase();
-                const secondaryLabel = typeof label === "string" ? undefined : label.secondaryLabel.toLowerCase();;
-
-                return (title.includes(searchValue.trim().toLowerCase()) || (secondaryLabel && secondaryLabel.includes(searchValue.trim().toLowerCase())));
+                const secondaryLabel =
+                    typeof label === "string"
+                        ? undefined
+                        : label.secondaryLabel?.toLowerCase();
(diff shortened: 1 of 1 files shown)


src/shared/dropdown-list/dropdown-list.tsx
```diff
@@ -205,18 +205,9 @@ export const DropdownList = <T, V>({
                     typeof label === "object"
                         ? label.title.toLowerCase()
                         : label.toLowerCase();
+                const secondaryLabel = typeof label === "string" ? undefined : label.secondaryLabel.toLowerCase();;
```
qroll, line 208: based on the types, secondaryLabel is optional so we should also check it exists first
```suggestion
                const secondaryLabel = typeof label === "string" ? undefined : label.secondaryLabel?.toLowerCase();
```

src/shared/dropdown-list/dropdown-list.tsx
```diff
@@ -205,18 +205,9 @@ export const DropdownList = <T, V>({
...
-                    return (
-                        title.includes(searchValue.trim().toLowerCase()) ||
-                        label.secondaryLabel.includes(
-                            searchValue.trim().toLowerCase()
-                        )
-                    );
-                }
-
-                return title.includes(searchValue.trim().toLowerCase());
+                return (title.includes(searchValue.trim().toLowerCase()) || (secondaryLabel && secondaryLabel.includes(searchValue.trim().toLowerCase())));
```
qroll, line 210: can extract `searchValue.trim().toLowerCase()` to a constant to be reused

qroll (changes requested):
can you format the file with prettier as well?

qroll (approved):

PR #266, opened 2023-08-01
[MOL-13286][Jasm] allow popover component to scroll on mobile view when content overflows

ijasm:
**Changes**
enable popover component to scroll on mobile view when content overflows

- delete branch

### DIFF
--- a/src/modal/modal-box.tsx
+++ b/src/modal/modal-box.tsx
@@ -1,5 +1,10 @@
 import React from "react";
-import { Box, CloseButton, CloseIcon } from "./modal-box.styles";
+import {
+    Box,
+    CloseButton,
+    CloseIcon,
+    ContentWrapper,
+} from "./modal-box.styles";
 import { ModalBoxProps } from "./types";
 
 export const ModalBox = ({
@@ -34,7 +39,7 @@ export const ModalBox = ({
     return (
         <Box data-testid={id} {...otherProps} onClick={handleOnClick}>
             {showCloseButton && renderCloseButton()}
-            {children}
+            <ContentWrapper>{children}</ContentWrapper>
         </Box>
     );
 };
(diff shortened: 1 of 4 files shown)


src/modal/modal-box.tsx
```diff
@@ -34,7 +39,7 @@ export const ModalBox = ({
     return (
         <Box data-testid={id} {...otherProps} onClick={handleOnClick}>
             {showCloseButton && renderCloseButton()}
-            {children}
+            <ContentWrapper>{children}</ContentWrapper>
```
qroll, line 42: 1. I'm concerned about adding to modal box as this may affect scroll behaviour for existing Modal usage. can this be applied on Popover? alternatively this can be handled on the dev side by wrapping the `children` in a div container with the desired scroll behaviour
2. for desktop, is there no handling needed?

ijasm: 1. oh yes you're right, it should be applied on Popover. i was thinking of making the change on the DS component itself rather than on the app side because that seems to be the standard behaviour, do you reckon that there may be exceptions to that? if so i can do it over on the app side.
2. for desktop, so far most of the popover content are not long enough to exceed the footer, if they do, the view can be scrolled, although the footer will also be scrolled up and might look a bit strange. not sure if that would be an issue in future when there are long content that exceeds the footer (can see here: https://tst.mylegacy.lifesg.io/find-a-service/eWills/step3/)

qroll: 1. okay, that would be fine for mobile since it's always a modal with fixed positioning. will trouble you to apply the changes
2. it's tricky with desktop being pretty context-specific since the element can be anywhere on the screen. I think it's possible to calculate the remaining space until the end of the page or set some max height, but I'm inclined to let the client handle it for now

qroll (approved):

PR #274, opened 2023-08-06
[BOOKINGSG-4350][WK] fix flat option error if it is  fisrt item on th…

Wilker0725:
**Changes**
Description of changes...
1. Fix error when flat option in first item list on the root
2. Should expanded the first subItems list 

```
// should not cause error if the flat option in root level
// should expanded first subItems list item. That is ["999", "822"]
export const options = [
    {
        label: "Option 1",
        value: { id: 200, name: "Option 1" },
        key: "200",
    },
    {
        label: "Category 1",
        value: { id: 999, name: "category 1" },
        key: "999",
        subItems: [
            {
                label: "Sub Category A",
                value: {
                    id: 820,
                    name: "Sub category a",
                },
                key: "820",
            },
            {
                label: "Sub Category B",
                value: { id: 821, name: "Sub category b" },
                key: "821",
            },
            {
                label: "Sub Category C",
                value: {
                    id: 822,
                    name: "Sub category c",
                },
                key: "822",
                subItems: [
                    {
                        label: "Honey",
                        value: {
                            id: 104,
                            name: "honey",
                        },
                        key: "104",
                    },
                ],
            },
        ],
    },
];

```

- [delete] branch

### DIFF
--- a/src/shared/nested-dropdown-list/nested-dropdown-list-helper.ts
+++ b/src/shared/nested-dropdown-list/nested-dropdown-list-helper.ts
@@ -81,11 +81,18 @@ export namespace NestedDropdownListHelper {
 const getInitialSubItem = <V1, V2, V3>(
     list: Map<string, CombinedFormattedOptionProps<V1, V2, V3>> | undefined
 ): string[] => {
+    let targetItem = null;
     for (const item of list.values()) {
         if (item.subItems && item.subItems.size) {
-            const [firstItemKey] = item.subItems.keys();
-            return item.subItems.get(firstItemKey).keyPath;
+            targetItem = item;
+            break;
         }
-        getInitialSubItem(item.subItems);
     }
+
+    if (!targetItem) {
+        const value = list.values().next().value;
+        return value.keyPath.slice(0, -1);
+    }
+
+    return getInitialSubItem(targetItem.subItems);
 };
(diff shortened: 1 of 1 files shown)


src/shared/nested-dropdown-list/nested-dropdown-list-helper.ts
```diff
@@ -81,11 +81,18 @@ export namespace NestedDropdownListHelper {
 const getInitialSubItem = <V1, V2, V3>(
     list: Map<string, CombinedFormattedOptionProps<V1, V2, V3>> | undefined
 ): string[] => {
+    let targetItem = null;
```
qroll, line 84: type this variable as `CombinedFormattedOptionProps` to get type checking

qroll (changes requested):

src/shared/nested-dropdown-list/nested-dropdown-list-helper.ts
```diff
@@ -81,11 +81,18 @@ export namespace NestedDropdownListHelper {
...
     for (const item of list.values()) {
         if (item.subItems && item.subItems.size) {
-            const [firstItemKey] = item.subItems.keys();
-            return item.subItems.get(firstItemKey).keyPath;
+            targetItem = item;
```
qroll, line 87: actually you could do this here without having to track a separate variable

```
getInitialSubItem(targetItem.subItems)
break;
```

qroll (approved):

PR #282, opened 2023-08-16
Multi nested select

Wilker0725:
**Changes**
Description of changes...
* Create new `NestedMultiSelect` component
* Update`NestedDropdownList` component

- [delete] branch

- [WARNING] Description of change in `NestedMultiSelect`

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/BOOKINGSG-4362)
- [Figma](https://www.figma.com/file/us90jOpWBahsqK2VAluPvD/Flagship-Design-System?node-id=11728%3A111511&mode=dev)


Wilker0725:
A note: Below fn will replace `updateCategoryChecked` when the `[-]` checkbox component is ready.

```
  export const updateCategoryChecked = <V1, V2, V3>(
        targetList: FormattedOptionMap<V1, V2, V3>,
        selectedKeyPaths: string[][]
    ) => {
        const update = (
            item: CombinedFormattedOptionProps<V1, V2, V3>
        ): CombinedFormattedOptionProps<V1, V2, V3> => {
            const matched = selectedKeyPaths.some(
                (key) => JSON.stringify(key) === JSON.stringify(item.keyPath)
            );

            if (!item.subItems) {
                if (matched) {
                    return {
                        ...item,
                        checked: true,
                    };
                } else {
                    return {
                        ...item,
                        checked: false,
                    };
                }
            }

            const subItems: Map<
                string,
                CombinedFormattedOptionProps<V1, V2, V3>
            > = new Map();

            item.subItems.forEach((subItem) => {
                const result = update(subItem);
                if (result) {
                    const key = result.keyPath[result.keyPath.length - 1];
                    subItems.set(key, result);
                }
            });

            const checkedStatus = Array.from(subItems).map(
                (items) => items[1].checked
            );
            const isChecked = checkedStatus.every(Boolean);
            const checkedMinus = checkedStatus.some(Boolean);

            return {
                ...item,
                checked: isChecked ? true : checkedMinus ? "minus" : false,
                subItems,
            } as CombinedFormattedOptionProps<V1, V2, V3>;
        };

        const list = new Map();

        for (const [key, item] of targetList) {
            const result = update(item);

            if (result && result.subItems && result.subItems.size) {
                list.set(key, result);
            }
        }

        return list;
    };
```

src/input-nested-multi-select/input-nested-multi-select.tsx
qroll, line 1: it'll be easier to keep selection management in one place instead of spreading it across `input-nested-multi-select` and `nested-dropdown-list`

if
1. handled by `nested-dropdown-list`
    - calculate the next selection state, then extract all selected subitems from this state
    - pro: might be easier to retrieve the subitems since the list is already formatted into the map
    - con: this component now has to manage selection state
2. handled by `input-nested-multi-select`
    - pro: keep `nested-dropdown-list` simple; it only fires the click events and the parent decides what to do with item/category selection
    - con: managing the `selectedItems` state is not performant

src/shared/nested-dropdown-list/list-item.tsx
```diff
@@ -67,7 +81,13 @@ export const ListItem = <V1, V2, V3>({
     // HELPER FUNCTIONS
     // =============================================================================
     const checkListItemSelected = (keyPath: string[]): boolean => {
-        return JSON.stringify(selectedKeyPath) === JSON.stringify(keyPath);
+        if (!multiSelect) {
+            return JSON.stringify(selectedKeyPath) === JSON.stringify(keyPath);
+        } else {
+            return selectedKeyPaths.some(
+                (key) => JSON.stringify(key) === JSON.stringify(keyPath)
```
qroll, line 88: actually can the `item.checked` state be used here instead

qroll (commented):
noting down some comments from the earlier code review

src/input-nested-multi-select/input-nested-multi-select.tsx
Wilker0725: cool! Appreciate sharing that pro and con!  🎉

src/shared/nested-dropdown-list/list-item.tsx
Wilker0725: due to the changes for `selectedKeyPaths: string[][]` now. below fn can use for single and nested. let me know if still need the `item.checked`

```
 const checkListItemSelected = (keyPath: string[]): boolean =>
        selectedKeyPaths.some(
            (key) => JSON.stringify(key) === JSON.stringify(keyPath)
        );
```

src/input-nested-multi-select/input-nested-multi-select.tsx
```diff
@@ -0,0 +1,487 @@
...
+            }
+
+            return findSelectedItem(subItems, nextKeyPath);
+        };
+
+        const selectedItems = [];
+
+        for (let i = 0; i < selectedKeyPaths.length; i++) {
+            const item = findSelectedItem(options, keyPaths[i]);
+            if (!item) break;
```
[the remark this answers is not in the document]
Wilker0725: it will be removed. thanks spotted this!

src/input-nested-multi-select/input-nested-multi-select.tsx
```diff
@@ -0,0 +1,487 @@
...
+
+        if (disabled || readOnly) {
+            return;
+        }
+
+        setShowOptions(!showOptions);
+        triggerOptionDisplayCallback(!showOptions);
+    };
+
+    const handleSelectItem = (
```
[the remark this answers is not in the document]
Wilker0725: it is good in derive the selected items from `updateSelectedItemFromKey`.

src/shared/nested-dropdown-list/list-item.tsx
```diff
@@ -128,7 +141,11 @@ export const ListItem = <V1, V2, V3>({
         const nextSubItems = item.subItems.values();
 
         return (
-            <List $expanded={item.expanded}>
+            <List
+                $expanded={item.expanded}
+                $level_3={item.keyPath.length + 1 === 3}
```
qroll, line 146: `level_3` doesn't work in the scenario where the first/second level have no subitems

in `ListItemSelector`, try setting the left margin when `$multiSelect` is true

src/input-nested-multi-select/input-nested-multi-select.tsx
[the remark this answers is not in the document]
Wilker0725: it is good derive value inside the parent. let me think. thanks for your time!

src/shared/nested-dropdown-list/list-item.tsx
Wilker0725: please check for this

Wilker0725: cause horizontal scrollable if we set it margin-left in `ListItemSelector`.

src/shared/nested-dropdown-list/nested-dropdown-list-helper.ts
```diff
@@ -36,6 +88,40 @@ export namespace NestedDropdownListHelper {
...
+            keyPaths = _keyPaths;
+            items = _items;
+
+            list = produce(
+                currentItems,
+                (draft: FormattedOptionMap<V1, V2, V3>) => {
+                    keyPaths.forEach((key) => {
+                        const parentKey = key.slice(0, -1);
+                        const item = getItemAtKeyPath(draft, parentKey);
+                        item.expanded = true;
```
qroll, line 112: L1 item does not get expanded if it does not have a non-category L2 item, is that intended?

qroll (changes requested):

src/shared/nested-dropdown-list/nested-dropdown-list.tsx
```diff
@@ -443,6 +493,28 @@ export const NestedDropdownList = <V1, V2, V3>({
...
 
+    const renderSelectAll = () => {
+        if (
+            multiSelect &&
+            initialItems.size > 0 &&
+            !isSearch &&
+            itemsLoadState === "success"
+        ) {
+            return (
+                <SelectAllContainer key="selectAll">
```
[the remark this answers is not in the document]
Wilker0725: nope. hah. 😅

src/shared/nested-dropdown-list/nested-dropdown-list-helper.ts
Wilker0725: it expanded. i use below `options` please let me know if i misunderstand.

```
[
    {
        label: "Category 1",
        value: { id: 999, name: "category 1" },
        key: "999",
        subItems: [
            {
                label: "Option 1",
                value: {
                    id: 10001,
                    name: "Option-1",
                },
                key: "10001",
            },
            {
                label: "Sub Category A",
                value: {
                    id: 820,
                    name: "Sub category a",
                },
                key: "820",
            },
        ],
    },
    {
        label: "Hive",
        value: { id: 487, name: "hive" },
        key: "32",
        subItems: [
            {
                label: "Level 8",
                value: { id: 100, name: "Level 8" },
                key: "100",
            },
            {
                label: "Level 9",
                value: { id: 101, name: "Level 9" },
                key: "101",
            },
            {
                label: "Level 10",
                value: { id: 102, name: "Level 10" },
                key: "102",
            },
        ],
    },
];
``

Comment (Wilker0725) [reply src/shared/nested-dropdown-list/nested-dropdown-list-helper.ts:112]: the L1 is `expanded: true`, but L2 which non-category `expanded: true` as well. 
I'll fix this.

File: src/input-nested-multi-select/input-nested-multi-select.tsx [hunk 7]
```diff
@@ -0,0 +1,401 @@
...
+                ) {
+                    isRemove = true;
+                    removeIndex = i;
+                    break;
+                }
+            }
+
+            if (isRemove) newKeyPaths.splice(removeIndex, 1);
+            else newKeyPaths.push(keyPath);
+        }
```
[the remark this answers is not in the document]
Wilker0725: thanks sharing! you so smart!

src/shared/nested-dropdown-list/nested-dropdown-list-helper.ts
qroll: > may i have the options? i tested, it expanded.

with the current options,
1. collapse Category 1
2. click Select All

Hive and the subcategories are expanded, but not Category 1

Wilker0725: interesting. i didn't expected that... let me check. thanks pointed this out.

src/input-nested-multi-select/input-nested-multi-select.tsx
```diff
@@ -87,47 +87,41 @@ export const InputNestedMultiSelect = <V1, V2, V3>({
...
-                    )
-                );
-            } else {
-                newKeyPaths = [
-                    ...new Map(
-                        [...newKeyPaths, ...keys].map((k) => [k.join("-"), k])
-                    ).values(),
-                ];
+            const selectedCount = selectedKeyPaths.filter((keyPath) =>
+                keyPath.join("-").startsWith(item.keyPath.join("-"))
```
qroll, line 100: can pull this out into a helper function ex.

```
isSubItem(item: string[], category: string[])
```

also it's safer to do
```
JSON.stringify(keyPath) === JSON.stringify(item.keyPath.slice(0, keyPath.length))
```

to avoid case of `["1-2"]` matching `["1", "2"]`

src/input-nested-multi-select/input-nested-multi-select.tsx
```diff
@@ -87,47 +87,41 @@ export const InputNestedMultiSelect = <V1, V2, V3>({
...
-                    JSON.stringify(newKeyPaths[i]) === JSON.stringify(keyPath)
-                ) {
-                    isRemove = true;
-                    removeIndex = i;
-                    break;
-                }
-            }
+            const selected = selectedKeyPaths.some((keyPath) =>
+                keyPath.join("-").startsWith(item.keyPath.join("-"))
+            );
```
qroll, line 114: this should be an equality check as well?

qroll (changes requested):

src/shared/nested-dropdown-list/list-item.styles.tsx
```diff
@@ -82,6 +82,7 @@ export const ListItemSelector = styled.button<ListItemSelectorProps>`
 export const Item = styled.li<ItemProps>`
     ${(props) => {
         switch (props.$level) {
+            case 1:
```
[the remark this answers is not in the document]
Wilker0725: hmmm. can! thanks you!

src/input-nested-multi-select/input-nested-multi-select.tsx
```diff
@@ -106,11 +106,17 @@ export const InputNestedMultiSelect = <V1, V2, V3>({
...
 
             if (selectedCount < selectableOptionKeyPaths.length) {
-                newKeyPaths = selectableOptionKeyPaths;
+                newKeyPaths = [
+                    ...new Map(
+                        [...selectedKeyPaths, ...selectableOptionKeyPaths].map(
+                            (k) => [k.join("-"), k]
+                        )
+                    ).values(),
+                ];
```
qroll, line 115: can explain this part?

was expecting `selectableOptionKeyPaths` to be directly concatenated to the `newKeyPaths` array

src/input-nested-multi-select/input-nested-multi-select.tsx
```diff
@@ -0,0 +1,407 @@
...
+                item.keyPath
+            );
+
+            const selectedCount = selectedKeyPaths.filter((keyPath) =>
+                isSubItem(keyPath, item.keyPath)
+            ).length;
+
+            newKeyPaths = selectedKeyPaths.filter(
+                (keyPath) =>
+                    !keyPath.join("-").startsWith(item.keyPath.join("-"))
```
qroll, line 105: this can use `isSubItem` as well

src/input-nested-multi-select/input-nested-multi-select.tsx
Wilker0725: `...new Map().values()` try to get the value unique.

Scenario:
1. Select first one category L2 item which key path is `["100", "200"]`;
2. User select the checkbox on L1 category, `selectableOptionKeyPaths` included the `["100", "200"]`.
3. `selectedKeyPaths` also included `["100", "200"]` which step 1 user already selected.
4. `new Map()` try to remove the duplication. similar the `new Set` method.

qroll: ok. this should be fine for most cases. will take this implementation.

in the previous line `newKeyPaths` already removed the subitems, so the duplication shouldn't happen

qroll (approved):

PR #285, opened 2023-08-18
[BOOKINGSG-4362][WK] update to selectedKeyPaths from selectedKeyPath

Wilker0725:
**Changes**
Description of changes...
* selectedKeyPath to plural `string[][]` in single nested select

- [delete] branch
- [WARNING] Description of change in `InputNestedSelect`

Will merge into  [Multi nested select PR](https://github.com/LifeSG/react-design-system/pull/282) later.

### DIFF
--- a/src/input-nested-select/input-nested-select.tsx
+++ b/src/input-nested-select/input-nested-select.tsx
@@ -60,7 +60,7 @@ export const InputNestedSelect = <V1, V2, V3>({
     // CONST, STATE
     // =============================================================================
     const [selectedKeyPaths, setSelectedKeyPaths] = useState<string[][]>(
-        _selectedKeyPath?.flat().length ? [_selectedKeyPath] : []
+        _selectedKeyPath ? [_selectedKeyPath] : []
     );
     const [selectedItem, setSelectedItem] =
         useState<SelectedItemType<V1, V2, V3>>();
@@ -74,9 +74,7 @@ export const InputNestedSelect = <V1, V2, V3>({
     // EFFECTS
     // =============================================================================
     useEffect(() => {
-        const newKeyPath = _selectedKeyPath?.flat().length
-            ? [_selectedKeyPath]
-            : [];
+        const newKeyPath = _selectedKeyPath ? [_selectedKeyPath] : [];
 
         setSelectedKeyPaths(newKeyPath);
         updateSelectedItemFromKey(options, _selectedKeyPath || []);
@@ -137,7 +135,7 @@ export const InputNestedSelect = <V1, V2, V3>({
     const getDisplayValue = (): string => {
         const { label, value } = selectedItem;
 
-        if (valueToStringFunction && value) {
+        if (valueToStringFunction) {
             return valueToStringFunction(value) || value.toString();
         } else {
             return label;
--- a/src/input-nested-select/input-nested-select.tsx
+++ b/src/input-nested-select/input-nested-select.tsx
@@ -234,12 +234,14 @@ export const InputNestedSelect = <V1, V2, V3>({
         if ((options && options.length > 0) || onRetry) {
             return (
                 <NestedDropdownList
-                    data-testid="dropdown-list"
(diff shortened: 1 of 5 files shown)


src/input-nested-select/input-nested-select.tsx
```diff
@@ -74,10 +74,12 @@ export const InputNestedSelect = <V1, V2, V3>({
     // EFFECTS
     // =============================================================================
     useEffect(() => {
-        const newKeyPath = _selectedKeyPath || [];
-        setSelectedKeyPath(newKeyPath);
+        const newKeyPath = _selectedKeyPath?.flat().length
+            ? [_selectedKeyPath]
+            : [];
```
qroll, line 79: can I confirm that the desired behaviour is:

if `selectedKeyPath` is
- undefined, init to `[]`
- empty array, init to `[]`
- non-empty, init to `[selectedKeyPath]`

this can be simplified to:
```suggestion
        const newKeyPath = _selectedKeyPath?.length ? [_selectedKeyPath] : [];
```

I don't think `flat` does anything here because it's already a 1-dimensional array

src/input-nested-select/input-nested-select.tsx
```diff
@@ -59,8 +59,8 @@ export const InputNestedSelect = <V1, V2, V3>({
     // =============================================================================
     // CONST, STATE
     // =============================================================================
-    const [selectedKeyPath, setSelectedKeyPath] = useState<string[]>(
-        _selectedKeyPath || []
+    const [selectedKeyPaths, setSelectedKeyPaths] = useState<string[][]>(
+        _selectedKeyPath?.flat().length ? [_selectedKeyPath] : []
```
qroll, line 63: refer to above comment

src/input-nested-select/input-nested-select.tsx
```diff
@@ -135,7 +137,7 @@ export const InputNestedSelect = <V1, V2, V3>({
     const getDisplayValue = (): string => {
         const { label, value } = selectedItem;
 
-        if (valueToStringFunction) {
+        if (valueToStringFunction && value) {
```
qroll, line 140: can you elaborate on why this check is needed? I thought the `isEmpty` check before this ensures there is a selection

src/shared/nested-dropdown-list/nested-dropdown-list-helper.ts
```diff
@@ -9,23 +9,27 @@ export type FormattedOptionMap<V1, V2, V3> = Map<
...
     ) => {
-        let keyPath = selectedKeyPath;
+        let keyPaths = selectedKeyPaths;
 
-        if (!keyPath || !keyPath.length) {
-            keyPath = getInitialSubItem(currentItems);
-            keyPath = keyPath.slice(0, -1);
+        if (!keyPaths || !keyPaths.length) {
+            keyPaths = getInitialSubItem(currentItems);
+            console.log("keyPaths: ", keyPaths);
```
qroll, line 18: unintended console log?

src/shared/nested-dropdown-list/nested-dropdown-list-helper.ts
```diff
@@ -81,13 +85,13 @@ export namespace NestedDropdownListHelper {
 // =============================================================================
 const getInitialSubItem = <V1, V2, V3>(
     list: Map<string, CombinedFormattedOptionProps<V1, V2, V3>> | undefined
-): string[] => {
+): string[][] => {
```
qroll, line 88: since `getInitialSubItem` is expected to always return only one item, the return type should remain as `string[]`. if the parent needs a 2d array, it can do its own wrapping into an array

qroll (changes requested):

src/input-nested-select/input-nested-select.tsx
Wilker0725: my bad. just now type yelling at me.  now is good.
i will remove the value.

qroll (approved):

PR #289, opened 2023-08-23
Add masked input component

ijasm:
**Changes**
Add MaskedInput component

- delete branch

**Additional information**

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/MOL-13583)

### DIFF
--- a/src/masked-input/masked-input.style.tsx
+++ b/src/masked-input/masked-input.style.tsx
@@ -23,7 +23,7 @@ export const InputGroupWrapper = styled(InputGroup)<InputGroupWrapperProps>`
 
 export const IconContainer = styled.div<IconProps>`
     display: flex;
-    height: 3rem;
+    height: calc(3rem - 2px);
     width: 3.25rem;
     align-items: center;
     justify-content: center;


src/masked-input/masked-input.tsx
```diff
@@ -0,0 +1,154 @@
...
+            onUnmask && onUnmask();
+        }
+    }, [isMasked]);
+
+    // =============================================================================
+    // EVENT HANDLERS
+    // =============================================================================
+
+    const handleFocus = () => !readOnly && setIsMasked(false);
+    const handleBlur = () => !readOnly && setIsMasked(true);
```
qroll, line 50: we can propagate the focus and blur events by calling `onFocus` and `onBlur` props here as well

stories/form/form-masked-input/form-masked-input.stories.mdx
```diff
@@ -0,0 +1,173 @@
...
+                <Form.MaskedInput
+                    label="This is the mask field with input"
+                    value="S1234567D"
+                    maskRange={[2, 5]}
+                />
+                <Form.MaskedInput
+                    label="This is the error state"
+                    errorMessage="Input is required"
+                    maskRange={[2, 5]}
+                />
```
qroll, line 47: there was this AC mentioned:
> If the input fails validation, the input should NOT be masked. The mask icon will remain disabled. Masking only occurs when the input is correct and meets the validation requirements.

just to confirm, it's up to the consumer to conditionally set the mask config depending on the value? (this seems to be the most flexible way, instead of hardcoding it to `error === true`)

qroll (changes requested):

src/masked-input/types.ts
```diff
@@ -0,0 +1,22 @@
+import { InputGroupProps } from "../input-group";
+
+export interface MaskedInputProps<T, V>
+    extends Omit<InputGroupProps<T, V>, "addon"> {
```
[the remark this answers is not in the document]
ijasm: the `onBlur` from `InputGroupProps` differs from the one in `InputProps` in that the former doesn't take in an argument and the latter does. if i extend from `InputProps`, the `MaskedInput`'s onBlur callback expects an argument, but the `onBlur` type in `InputGroup` which `MaskedInput` uses doesn't return an argument. so there was a type error which was why i extended from `InputGroupProps`. but anyway i managed to resolve it by setting the event arg to optional, not sure if that's the correct way of handling..

stories/form/form-masked-input/form-masked-input.stories.mdx
ijasm: ya that makes sense. in that case should we also allow consumers to configure the mask icon disabled state as well or should we toggle it depending on whether the value is masked or not?

ijasm:
will be making some changes following new mask icon touch area specifications: https://www.figma.com/file/us90jOpWBahsqK2VAluPvD/Flagship-Design-System?type=design&node-id=13359-122674&mode=design&t=DLV5Fr6zioU5FPkt-0

stories/form/form-masked-input/form-masked-input.stories.mdx
qroll: Hmm yeah it should be handled consistently. If masking is disabled, the icon should also be disabled. How about a `disableMask` boolean (default false) that disables masking entirely?

src/masked-input/types.ts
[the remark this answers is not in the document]
qroll: I see. It's because onBlur event doesn't always originate from the input element in `InputGroup` 😞

In that case MaskedInput should indeed follow `InputGroupProps`. Sorry for the trouble.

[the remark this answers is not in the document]
qroll: on second thought I find leaving the generic here to be weird

can `MaskedInput` specify its own `onBlur` typing where the event is optional?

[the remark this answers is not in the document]
ijasm: but i still think its more correct to extend from `InputProps` since the MaskedInput specifically uses the **custom** `InputGroup` in which the onBlur event does indeed originates from the input element... 🤔 let me know what you think!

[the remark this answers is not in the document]
qroll: agree. it's safe to assert that the blur event is always visible for the custom variant.

perhaps we can just have a comment on why the `?` is needed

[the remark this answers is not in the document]
ijasm: instead of setting the `event` to an optional argument on `MaskedInput`'s `handleBlur`, i fixed the `onBlur` prop type on `InputGroup` to take in an optional `event` argument and added a comment on why that is needed

src/masked-input/masked-input.tsx
```diff
@@ -33,9 +33,7 @@ const Component = (
 ) => {
     const isEmptyReadOnlyState = readOnly && isEmpty(value);
     const [isMasked, setIsMasked] = useState(!disableMask);
-    const [updatedValue, setUpdatedValue] = useState(
```
ijasm: since the `updatedValue`'s length is used for checking whether masking / mask icon should be disabled, the updatedValue shouldn't be set to "-" when the value is empty

src/masked-input/masked-input.style.tsx
```diff
@@ -1,21 +1,41 @@
...
+        cursor: ${({ readOnly, isDisabled }) =>
+            readOnly && !isDisabled ? "pointer" : "initial"};
+    }
+`;
+
 export const IconContainer = styled.div<IconProps>`
     display: flex;
-    width: 1.125rem;
-    height: 1.125rem;
+    height: 3rem;
```
qroll, line 26: note that the input group height is based on the content + border, so we need to `calc(3rem - 2px)` here to keep the overall component height at a consistent 48px

setting a fixed 3rem height on the `InputGroupWrapper` also works

src/masked-input/masked-input.style.tsx
```diff
@@ -1,21 +1,41 @@
 import styled from "styled-components";
 import { Color } from "../color";
+import { InputGroup } from "../input-group";
+
+interface InputGroupWrapperProps {
+    readOnly: boolean;
+    isDisabled: boolean;
```
qroll, line 7: mark as a transient prop since this is not a html attribute

```suggestion
    $isDisabled: boolean;
```

qroll (changes requested):

src/masked-input/masked-input.style.tsx
ijasm: there's a requirement for the touch area of the mask icon to be of height 3rem and width 3.25rem... hmm if i do calc(3rem - 2px), the icon container height doesn't stretch to the entire input group height. any ideas how to resolve this?

ijasm: all good now!

qroll (approved):

qroll:
I missed out on the top-level export in `src/index.ts`, will add that in when bumping the release version

PR #290, opened 2023-08-24
Minor aesthetic changes

vivchong:
**Changes**
1. Shadow (changed to look more like the ones in Figma): 
- Blur ppty: 3px instead of 6px
- Offset ppty: 0px instead of 1px

2. Colour of disabled Label for all Form input fields: 
- N3 instead of N4 (no need to switch colours if disabled)

3.  Colour of Close Icon for modal: 
- N3 instead of N4

Context for changes:
Designers were doing visual audit of the components and noticed some minor inconsistencies.

### DIFF
--- a/src/form/form-label.style.tsx
+++ b/src/form/form-label.style.tsx
@@ -5,7 +5,6 @@ import { Text, TextStyleHelper } from "../text";
 // =============================================================================
 // STYLING
 // =============================================================================
-export const Label = styled.label<LabelProps>`
 export const Label = styled.label`
     ${TextStyleHelper.getTextStyle("H5", "semibold")}
     color: ${Color.Neutral[3]};
--- a/src/form/form-label.style.tsx
+++ b/src/form/form-label.style.tsx
@@ -3,18 +3,11 @@ import { Color } from "../color";
 import { Text, TextStyleHelper } from "../text";
 
 // =============================================================================
-// STYLE INTERFACE
-// =============================================================================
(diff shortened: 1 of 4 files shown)


src/spec/design-token-spec/base-design-token-set.ts
```diff
@@ -4,10 +4,10 @@ import { DesignTokenSet } from "../../design-token/types";
 
 export const BaseDesignTokenSet: DesignTokenSet = {
     InputBoxShadow: css`
-        inset 0 0 6px 1px ${Color.Shadow.Accent}
+        inset 0 0 3px 0px ${Color.Shadow.Accent}
```
qroll, line 7: in the thread you mentioned the blur property should be 4px, is this the final intended style?

vivchong: yeah, I got the 4px from Figma's CSS styling, but it didn't look the same when rendered. 
so I adjusted the values to look more similar to the components in Figma

src/form/form-label.style.tsx
```diff
@@ -15,8 +15,7 @@ interface LabelProps {
 export const Label = styled.label<LabelProps>`
```
keithtxw, line 15: Since disabled styling is no longer needed, can consider removing the style interface entirely.

keithtxw (changes requested):

qroll (approved):

PR #296, opened 2023-08-30
Handle cleared and invalid date values

### LINKED ISSUE #294: Bug when user tries to clear the date in the DateInput component
**Component**
DateInput

**Describe the bug**
When the user tries to clear the date input using backspace, and clicks on Done, and clicks back into the date component again, the calendar shows Invalid Date.

**To Reproduce**

1. Go to https://designsystem.life.gov.sg/react/index.html?path=/story/form-dateinput--date-input
2. Click on any part of the the default date input field, and fill it in with any date, and select Done
3. Click back into the default date input field (with the date that you'd already input previously), and use the backspace to clear the date, and select Done
4. Click back into the default date input field and see the error below 

Screenshot 2023-08-25 at 10 58 40 AM

**Expected behavior**
If the user clicks back into the field after Step 3, the component should revert back to it's default empty state (ie as if the user has not entered any date at all)

**Screenshots**

https://github.com/LifeSG/react-design-system/assets/71697991/832d21b2-72f5-4cf0-8908-33a3fc12d353

**Additional context**
This also affects the DateInput without buttons, and DateRangeInput as well.

qroll:
Resolves #294

**Changes**
- Fix issue where cleared values resulted in "Invalid Date"s on the calendar day grid
- Fix issue where invalid dates like "xxx-x-x" results in NaN display in the date input
- [delete] branch

**Changelog entry**

Bug fixes
- Handle display of cleared and invalid date values in `DateInput` and `DateRangeInput`

### DIFF
--- a/src/util/date-helper.ts
+++ b/src/util/date-helper.ts
@@ -1,4 +1,4 @@
-import dayjs from "dayjs";
+import dayjs, { Dayjs } from "dayjs";
 
 const MONTHS_WITH_31_DAYS = [1, 3, 5, 7, 8, 10, 12];
 const MONTHS_WITH_30_DAYS = [4, 6, 9, 11];
@@ -79,5 +79,9 @@ export namespace DateHelper {
         const startTime = dayjs(start, format);
         const endTime = dayjs(end, format);
         return endTime.diff(startTime, "minute");
+    };
+
+    export const toDayjs = (date: string): Dayjs => {
+        return date ? dayjs(date) : dayjs();
     };
 }
--- a/src/util/date-input-helper.ts
+++ b/src/util/date-input-helper.ts
@@ -1,7 +1,4 @@
 import dayjs from "dayjs";
-import isBetween from "dayjs/plugin/isBetween";
-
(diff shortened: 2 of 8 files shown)


src/util/date-input-helper.ts
```diff
@@ -31,4 +28,14 @@ export namespace DateInputHelper {
 
         return false;
     };
+
+    export const sanitizeInput = (date: string): string => {
```
qroll, line 32: this helps to convert invalid date strings like "2023-x-x" to an empty string (i.e. those values get ignored)

src/util/date-helper.ts
```diff
@@ -80,4 +80,8 @@ export namespace DateHelper {
         const endTime = dayjs(end, format);
         return endTime.diff(startTime, "minute");
     };
+
+    export const toDayjs = (date: string): Dayjs => {
```
qroll, line 84: to circumvent the behaviour where `dayjs("")` results in invalid day instead of returning the current day like `dayjs(undefined)`

keithtxw (approved):

PR #300, opened 2023-09-04
Data table styling

SasithaRajapakasha:
**Changes**
Style update to data table

- [delete] branch

**Additional information**

- You may refer to this [BOOKINGSG-4463](https://jira.ship.gov.sg/browse/BOOKINGSG-4633)

### DIFF
--- a/src/data-table/data-table.styles.tsx
+++ b/src/data-table/data-table.styles.tsx
@@ -148,3 +148,7 @@ export const ErrorDisplayElement = styled(ErrorDisplay)`
         margin-top: 1rem;
     }
 `;
+
+export const EmptyViewCell = styled.td`
+    padding: 4rem 0;
+`;
--- a/src/data-table/data-table.styles.tsx
+++ b/src/data-table/data-table.styles.tsx
@@ -35,7 +35,6 @@ export const TableWrapper = styled.div`
     width: 100%;
     border: 1px solid ${borderColor};
     border-radius: 0.5rem;
-    overflow: auto;
 `;
 
 export const Table = styled.table`
--- a/src/data-table/data-table.styles.tsx
+++ b/src/data-table/data-table.styles.tsx
@@ -1,8 +1,8 @@
 import styled, { css } from "styled-components";
 import { Color } from "../color";
-import { TextStyleHelper } from "../text";
-import { Checkbox } from "../checkbox";
 import { ErrorDisplay } from "../error-display";
+import { DesignToken } from "../design-token";
+import { Text } from "../text";
 
 // =============================================================================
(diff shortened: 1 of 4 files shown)


src/data-table/data-table.tsx
```diff
@@ -124,10 +131,11 @@ export const DataTable = ({
             <HeaderCell
                 data-testid={getDataTestId("header-selection")}
                 $clickable={false}
+                $isCheckbox={true}
             >
                 <CheckBoxWrapper>
                     {enableSelectAll && (
-                        <CheckboxElement
+                        <Checkbox
```
qroll, line 138: there should be an indeterminate state set for the case when not all items are selected

qroll (commented):
accessibility note: Voiceover does not announce the column header but the table markup seems fine, will try to find out why

qroll:
a scrollbar appears for the loading state

Screenshot 2023-09-05 at 1 48 55 PM

src/data-table/data-table.tsx
SasithaRajapakasha: I have updated indeterminate changes with next changes, Can I do the update of the indeterminate in there?

src/data-table/data-table.styles.tsx
```diff
@@ -35,7 +35,6 @@ export const TableWrapper = styled.div`
     width: 100%;
     border: 1px solid ${borderColor};
     border-radius: 0.5rem;
-    overflow: auto;
```
qroll, line 27: this is needed to constrain the table backgrounds within the border radius, so you'll have to put this back

as for the scrollbar, I found that an alternative fix is to set `overflow: hidden;` on the `LoaderWrapper`. can add a comment in the code if you decide to use this fix

qroll:
> accessibility note: Voiceover does not announce the column header but the table markup seems fine, will try to find out why

Hmm I tried again on a different Macbook and it's working as expected. I assume it's a bug specific to the previous device I was testing on. No action needed.

src/data-table/data-table.tsx
```diff
@@ -180,17 +188,25 @@ export const DataTable = ({
                 data-testid={getDataTestId(`row-${cellId}`)}
                 key={cellId}
                 style={style}
+                $isCheckbox={false}
             >
-                {cellData}
+                {typeof cellData === "string" ? (
+                    <BodyCellContent>{cellData}</BodyCellContent>
+                ) : (
+                    cellData
```
qroll, line 196: something that I missed out, if `cellData` is a function, it should be called with the row item and `RowRenderProps`

qroll (approved):

PR #304, opened 2023-09-06
Data table action bar

SasithaRajapakasha:
**Changes**
Data table action bar and disabled checkbox

- [delete] branch

**Additional information**

- You may refer to this [BOOKINGSG-4633](https://jira.ship.gov.sg/browse/BOOKINGSG-4633)

### DIFF
--- a/src/data-table/data-table.styles.tsx
+++ b/src/data-table/data-table.styles.tsx
@@ -42,7 +42,6 @@ export const TableWrapper = styled.div`
 
 export const TableContainer = styled.div`
     overflow: scroll;
-    flex-grow: 1;
 `;
 
 export const Table = styled.table`
@@ -58,13 +57,9 @@ export const Table = styled.table`
     tr:nth-child(2) {
         border-top: none;
     }
-    max-height: 5rem;
-    overflow-y: scroll;
-    width: 100%;
 `;
 
 export const SelectionBar = styled.div`
-    flex: 1;
     display: flex;
     flex-direction: row;
     width: 100%;
--- a/src/data-table/data-table.styles.tsx
+++ b/src/data-table/data-table.styles.tsx
@@ -32,6 +32,8 @@ const fontColor = Color.Neutral[1];
 // STYLES
 // =============================================================================
 export const TableWrapper = styled.div`
+    display: flex;
+    flex-direction: column;
     width: 100%;
     border: 1px solid ${borderColor};
     border-radius: 0.5rem;
@@ -39,9 +41,8 @@ export const TableWrapper = styled.div`
 `;
 
 export const TableContainer = styled.div`
-    width: 100%;
-    max-height: 90vh;
     overflow: scroll;
+    flex-grow: 1;
 `;
 
 export const Table = styled.table`
@@ -60,6 +61,17 @@ export const Table = styled.table`
     max-height: 5rem;
     overflow-y: scroll;
     width: 100%;
+`;
+
+export const SelectionBar = styled.div`
+    flex: 1;
+    display: flex;
+    flex-direction: row;
+    width: 100%;
+    height: 3.5rem;
+    padding: 1rem;
+    border-top: 1px solid ${borderColor};
+    background-color: ${DesignToken.Table.Cell.Selected};
 `;
 
 export const HeaderRow = styled.tr`
@@ -165,16 +177,6 @@ export const ErrorDisplayElement = styled(ErrorDisplay)`
 
 export const EmptyViewCell = styled.td`
     padding: 4rem 0;
-`;
-
-export const SelectionBar = styled.div`
-    display: flex;
-    flex-direction: row;
-    width: 100%;
-    height: 3.5rem;
-    padding: 1rem;
-    border-top: 1px solid ${borderColor};
-    background-color: ${DesignToken.Table.Cell.Selected};
 `;
 
 export const ClearSelectionAction = styled(Text.H5)`
--- a/src/data-table/data-table.styles.tsx
+++ b/src/data-table/data-table.styles.tsx
@@ -11,6 +11,7 @@ import { Text } from "../text";
 interface TableProps {
     $end: boolean;
     $scrollable: boolean;
+    $stickyHeader: boolean;
 }
 interface TableBodyProps {
     $showLastRowBottomBorder: boolean;
@@ -84,11 +85,17 @@ export const Table = styled.table<TableProps>`
     border-spacing: 0;
     width: 100%;
 
-    thead {
-        position: sticky;
-        top: 0;
-        z-index: 20;
-    }
+    ${(props) => {
+        if (props.$stickyHeader) {
+            return css`
+                thead {
+                    position: sticky;
+                    top: 0;
+                    z-index: 20;
+                }
+            `;
+        }
+    }};
 `;
 
 export const TableBody = styled.tbody<TableBodyProps>`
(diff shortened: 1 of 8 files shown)


src/data-table/data-table.tsx
```diff
@@ -256,18 +277,37 @@ export const DataTable = ({
         );
     };
 
+    const renderSelectionBar = () => {
+        return (
+            <SelectionBar>
+                <Text.H5 weight="semibold">{`${selectedIds.length} items selected`}</Text.H5>
+                <ClearSelectionAction onClick={onClearSelectionClick}>
+                    Clear Selection
+                </ClearSelectionAction>
```
qroll, line 286: This should be a button as it's an interactive element. I think the closest DS component is `<Button.Small styleType="link">`

src/data-table/types.ts
```diff
@@ -9,13 +9,18 @@ export interface DataTableProps {
...
     className?: string | undefined;
     selectedIds?: string[] | undefined;
+    disabledIds?: string[] | undefined;
     /** enable checkboxes */
     enableMultiSelect?: boolean | undefined;
     /** display select all checkbox */
     enableSelectAll?: boolean | undefined;
+    /** display action bar */
+    enableSelectionBar?: boolean | undefined;
+    selectionBarContent?: (() => ReactNode) | undefined;
```
qroll, line 19: the standard JSX.Element should be sufficient in this case

```suggestion
    selectionBarContent?: ReactNode | undefined;
```

qroll (changes requested):
some comments

for the sticky bar implementation, let me think on it again and also check in with the designer. I realised that one of the mockups shows a floating bar with a box shadow, which requires a totally different approach 🙃 apologies for missing out on this during KO

src/data-table/types.ts
```diff
@@ -9,13 +9,19 @@ export interface DataTableProps {
...
     selectedIds?: string[] | undefined;
+    disabledIds?: string[] | undefined;
     /** enable checkboxes */
     enableMultiSelect?: boolean | undefined;
     /** display select all checkbox */
     enableSelectAll?: boolean | undefined;
+    /** display action bar */
+    enableSelectionBar?: boolean | undefined;
+    enableStickyHeader?: boolean | undefined;
+    selectionBarContent?: ReactNode | undefined;
```
qroll, line 20: following the terminology used in figma


```suggestion
    enableActionBar?: boolean | undefined;
    enableStickyHeader?: boolean | undefined;
    actionBarContent?: ReactNode | undefined;
```

src/data-table/data-table.tsx
```diff
@@ -28,27 +32,55 @@ export const DataTable = ({
...
     // ===========================================================================
     // HELPER FUNCTIONS
     // ===========================================================================
-    const isAllCheckBoxSelected = (): boolean => {
+    const isAllCheckboxSelected = (): boolean => {
         return selectedIds?.length === rows.length;
     };
 
+    const isIndeterminateCheckbox = (): boolean => {
+        return selectedIds?.length !== 0 && !isAllCheckboxSelected();
```
qroll, line 71: noticed that the indeterminate state was showing up when no selection is passed, just some reworking of the length check

```suggestion
        return selectedIds && selectedIds.length && !isAllCheckboxSelected();
```

src/data-table/data-table.tsx
```diff
@@ -256,18 +311,43 @@ export const DataTable = ({
         );
     };
 
+    const renderSelectionBar = () => {
+        return (
+            <SelectionBar $isFloating={!isAtBottom}>
+                <Text.H5 weight="semibold">{`${selectedIds.length} items selected`}</Text.H5>
```
qroll, line 317: to handle the singular/plural forms:

1 item selected
2 items selected

src/data-table/data-table.tsx
```diff
@@ -256,18 +311,43 @@ export const DataTable = ({
...
+                <Button.Small styleType="link" onClick={onClearSelectionClick}>
+                    Clear Selection
```
qroll, line 319:
```suggestion
                    Clear selection
```

src/data-table/data-table.tsx
```diff
@@ -28,27 +32,55 @@ export const DataTable = ({
...
+    const [isAtBottom, setIsAtBottom] = useState(false);
+    const [showLastBorder, setShowLastBorder] = useState(false);
+    const tableRef = useRef<HTMLInputElement>();
+
+    // =============================================================================
+    // EFFECTS
+    // =============================================================================
+    useEffect(() => {
+        scrollHandler();
+        checkLastBorder();
```
qroll, line 60: what does this do? also if row height changes, does this need to be recomputed?

qroll (commented):
some other issues:
- there’s still a jitter effect when scrolling to the bottom
- the sticky behaviour doesn’t work when height is not defined

src/data-table/data-table.tsx
SasithaRajapakasha: Yes. I believe, if rows get changed. This need to be  re computed

src/data-table/data-table.tsx
```diff
@@ -28,27 +34,72 @@ export const DataTable = ({
...
+    const tableRef = useRef<HTMLTableElement>();
+
+    const { ref: stickyRef, inView: isLastRowInView } = useInView({
+        threshold: 0,
+    });
+    const { ref: endRef, inView: atEnd } = useInView({
+        threshold: 0,
+    });
+    const { ref: containerRef, inView: isContainerInView } = useInView({
+        threshold: 0.2,
```
qroll, line 65: this threshold is percentage based so it might not work when the table does not have a fixed height or when it's very tall relative to the viewport

an alternative is using root margin to adjust the offset
```
threshold: 0,
rootMargin: "0px 0px -100px 0px", <--- adjust based on the header height
```

but note that this doesn't work in the Canvas tab due to iframe limitations in Storybook

qroll:
I noticed that there’s still a chance for the jitter to happen when scrolling between the rows Title 14 and 15

I think what’s happening is the position change of the action bar ends up moving the last row out of the view, so it keeps recalculating `isLastRowInView`

I don't have a good solution for this, but we could try
1. debounce so hopefully only one `isLastRowInView` becomes the winner
2. use `onScroll` to detect user has scrolled to end of table. this will still result in the jitter, but it can be worked around by tracking the user's scroll direction

src/data-table/data-table.tsx
```diff
@@ -60,15 +150,64 @@ export const DataTable = ({
...
+        } else {
+            actionBarRef.current.style.transform = `translateY(0)`;
+        }
+    };
+
+    const checkLastBorder = () => {
+        setShowLastBorder(
+            tableRef.current?.scrollHeight <
+                (tableRef.current?.childNodes[0] as HTMLElement).clientHeight
+        );
```
qroll, line 198: in which scenario is the table scrollHeight less than the height of the first row?

src/data-table/data-table.styles.tsx
```diff
@@ -49,11 +80,76 @@ export const Table = styled.table`
...
+    }};
+`;
+
+export const ActionBar = styled.div<ActionBarProps>`
+    overflow: hidden;
+    display: flex;
+    ${(props) =>
+        props.$float &&
+        `
+            transform: translateX(-0.5%) translateY(-32px);
```
qroll, line 138:
```suggestion
            transform: translateX(-0.5%) translateY(-2rem);
```

qroll (changes requested):

qroll (approved):

PR #305, opened 2023-09-07
Class name props

qroll:
**Changes**

- Ensure components receive the `className` attribute for custom styling
- Also noticed that some components do not take in the standard `id` and `data-testid` attributes, took the chance to add them and correct the prop table documentation
- [delete] branch

**Changelog entry**

Bug fix
- Pass `className`, `id` and `data-testid` props to all components

### DIFF
--- a/src/file-upload/dropzone.tsx
+++ b/src/file-upload/dropzone.tsx
@@ -90,7 +90,7 @@ const Component = (
                 accept={accept}
                 capture={capture}
                 multiple={multiple}
-                data-testid={`${testId}-input` || "dropzone-input"}
+                data-testid={testId ? `${testId}-input` : "dropzone-input"}
                 onChange={handleChange}
             />
             {children}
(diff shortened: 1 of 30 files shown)


qroll:
Some prop tables are still using the old documentation style. Shall update in another PR.

src/file-upload/dropzone.tsx
keithtxw, line 1: Actually this one is an internal component. But no harm anyway.

keithtxw (approved):

src/file-upload/dropzone.tsx
```diff
@@ -89,7 +90,7 @@ const Component = (
                 accept={accept}
                 capture={capture}
                 multiple={multiple}
-                data-testid="dropzone-input"
+                data-testid={`${testId}-input` || "dropzone-input"}
```
keithtxw, line 93: This causes a breakage in the test as it is looking for `dropzone-input`. Maybe we should check if `testId` is specified before setting it?

keithtxw (changes requested):

keithtxw (approved):

PR #310, opened 2023-09-12
Timeline variant enhancement

Wilker0725:
**Changes**
Description of changes...
* Enhance variants. `completed`, `current`, `upcoming-active` and `upcoming-inactive`.

- [delete] branch

Description of change in `Timeline Component`

**Additional information**

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/BOOKINGSG-4566)
- [Figma](https://www.figma.com/file/us90jOpWBahsqK2VAluPvD/Flagship-Design-System?node-id=16810%3A1919&mode=dev)

### DIFF
--- a/src/timeline/timeline.style.tsx
+++ b/src/timeline/timeline.style.tsx
@@ -1,5 +1,5 @@
 import styled, { css } from "styled-components";
-import { TickCircleFillIcon } from "@lifesg/react-icons";
+import { TickIcon } from "@lifesg/react-icons";
 import { Color } from "../color";
 import { MediaQuery } from "../media";
 import { Text } from "../text";
@@ -23,7 +23,9 @@ interface VariantStyleProps {
 // STYLE INTERFACES
 // =============================================================================
 export const CircleIndicator = styled.div<VariantStyleProps>`
-    overflow: hidden;
+    display: flex;
+    justify-content: center;
+    align-items: center;
     width: 1.5rem;
     height: 1.5rem;
     margin-top: 4px;
@@ -45,6 +47,7 @@ export const CircleIndicator = styled.div<VariantStyleProps>`
                 `;
             case "completed":
                 return css`
+                    background-color: ${Color.Validation.Green.Icon};
                     svg {
                         display: block;
                     }
@@ -53,15 +56,9 @@ export const CircleIndicator = styled.div<VariantStyleProps>`
     }}
 `;
 
-export const TickCircleIcon = styled(TickCircleFillIcon)`
+export const IconTick = styled(TickIcon)`
+    color: ${Color.Neutral[8]};
     display: none;
-    width: 100%;
-    height: 100%;
-    transform: scale(1.2);
-
-    path {
-        fill: ${Color.Validation.Green.Icon};
-    }
 `;
 
 export const LineIndicator = styled.div<VariantStyleProps>`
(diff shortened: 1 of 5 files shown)


src/timeline/timeline.tsx
```diff
@@ -68,21 +71,27 @@ export const Timeline = ({
...
     const renderItems = () =>
         items.map((item: TimelineItemProps, index) => {
-            const { title, content, statuses } = item;
+            const { title, content, statuses, variant: _variant } = item;
             const circleIndicatorTestId = baseIndicatorTestId
                 ? `circleindicator${index + 1}_div_${baseIndicatorTestId}`
                 : "circleindicator";
+            const variant =
+                (!_variant && index === 0 ? "current" : _variant) ||
+                "upcoming-active";
```
qroll, line 83: we could reorganise the condition checks here so the logic to set the default variant is kept together

```suggestion
            const variant = _variant || (index === 0 ? "current" : "upcoming-active");
```

src/timeline/timeline.style.tsx
```diff
@@ -1,60 +1,99 @@
...
+export const TickCircleIcon = styled(TickCircleFillIcon)`
+    display: none;
+    width: 100%;
+    height: 100%;
+    transform: scale(1.2);
+
+    path {
+        fill: ${Color.Validation.Green.Icon};
+    }
+`;
```
qroll, line 65: does `color` not work here?

Also I wonder if you can avoid the `transform` by using `TickIcon` instead. The circle indicator would still provide the green background, and the tick can be coloured white

src/timeline/timeline.tsx
```diff
@@ -68,21 +71,26 @@ export const Timeline = ({
...
                 <TimelineItem key={`timeline-item-${index}`}>
                     <TimelineIndicators>
                         <CircleIndicator
                             data-testid={circleIndicatorTestId}
-                            $filled={index === 0}
-                        />
-                        <LineIndicator />
+                            $variant={variant}
+                        >
+                            <IconTick />
```
qroll, line 91: instead of a fixed icon, can have a `renderIcon(variant)` with a switch case that returns `TickIcon` for completed, otherwise default null

in `timeline.style.tsx`, you can simply style all svg children with color, width and height without having to toggle the display

src/timeline/timeline.style.tsx
```diff
@@ -1,60 +1,96 @@
...
 
-export const LineIndicator = styled.div`
-    width: 0.25rem;
+export const IconTick = styled(TickIcon)`
+    color: ${Color.Neutral[8]};
+    display: none;
+`;
+
+export const LineIndicator = styled.div<VariantStyleProps>`
+    width: 4px;
```
qroll, line 65: did you mean to change this to px?

qroll (changes requested):

src/timeline/timeline.style.tsx
Wilker0725: ya. i thoughts we will use `px` if the value is small enough. haha

qroll: hmm I'd say usually spacing can be kept in rem even for 4px/8px

but in this case it needs to follow the border width which is in px, so yea this is fine

src/timeline/timeline.tsx
Wilker0725: i see. No needed to render into DOM if didn't use it! thanks!

qroll (approved):

PR #313, opened 2023-09-14
Changed shadow blur value from 3px to 4px

vivchong:
In preparation for dimension tokens.

**Changes**
- Changed shadow's blur value from `3px` to `4px` in preparation for introduction of the dimension tokens.
- [delete] branch


vivchong:
I just noticed some other box-shadows with 3px so let me confirm the styling of those first, and I'll make another commit!

keithtxw (approved):

PR #315, opened 2023-09-20
Hide nav branding and adjust navbar styles

qroll:
**Changes**

- Introduce prop `hideNavBranding` to hide brand logos in navbar and drawer. Nav items will be aligned to the left.
- Reduce logo size for mobileS to cater for wider logos
- Add data-ids for brand elements in drawer to support targeted styling (previously only navbar brand elements could be targeted)
- [delete] branch

**Changelog entry**

-  Add `hideNavBranding` prop to hide brand logos in `Navbar`

### DIFF
--- a/src/navbar/navbar-items.tsx
+++ b/src/navbar/navbar-items.tsx
@@ -121,9 +121,8 @@ export const NavbarItems = <T,>({
                         selectedIndex >= 0 &&
                         selectedIndex === index &&
                         showSubMenu;
-                    const alignLeft = index === 0 && hideNavBranding;
                     return (
-                        <LinkItem key={index} $alignLeft={alignLeft}>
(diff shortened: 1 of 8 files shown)


src/navbar/navbar-items.tsx
```diff
@@ -119,8 +121,9 @@ export const NavbarItems = <T,>({
                         selectedIndex >= 0 &&
                         selectedIndex === index &&
                         showSubMenu;
+                    const alignLeft = index === 0 && hideNavBranding;
```
keithtxw, line 124: Since we are already restricting only the first child to have the align left, this check is redundant? On the same note, then maybe can have a generic prop called `$hiddenBranding` or something

keithtxw (approved):

PR #326, opened 2023-10-09
[MOL-14046][HFAH] auto uppercase IC input field for eWills & HDB carparking

Hamiriza:
**Changes**
Added  a new prop to cater for the auto uppercase IC input field

- delete branch
- You may refer to this [MOL-14046](https://jira.ship.gov.sg/browse/MOL-14046)

### DIFF
--- a/src/masked-input/masked-input.tsx
+++ b/src/masked-input/masked-input.tsx
@@ -26,7 +26,7 @@ const Component = (
         onBlur,
         error,
         disableMask,
-        transformUppercase,
+        transformInput,
         ...otherProps
     }: MaskedInputProps,
     ref: React.Ref<HTMLInputElement>
@@ -60,8 +60,11 @@ const Component = (
     const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
         let value = event.target.value;
 
-        if (transformUppercase) {
+        if (transformInput == "uppercase") {
             value = value.toUpperCase();
+        }
+        if (transformInput == "lowercase") {
+            value = value.toLowerCase();
         }
         setUpdatedValue(value);
         onChange && onChange(event);
--- a/src/masked-input/masked-input.tsx
+++ b/src/masked-input/masked-input.tsx
@@ -60,12 +60,17 @@ const Component = (
     const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
         let value = event.target.value;
 
-        if (transformInput === "uppercase") {
-            value = value.toUpperCase();
+        if (transformInput) {
+            switch (transformInput) {
+                case "uppercase":
+                    value = value.toUpperCase();
+                    break;
+                case "lowercase":
+                    value = value.toLowerCase();
+                    break;
+            }
         }
-        if (transformInput === "lowercase") {
-            value = value.toLowerCase();
-        }
+
         setUpdatedValue(value);
         event.target.value = value;
         onChange && onChange(event);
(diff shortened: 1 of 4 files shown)


src/masked-input/types.ts
```diff
@@ -13,6 +13,7 @@ export interface MaskedInputProps extends InputProps {
     onMask?: (() => void) | undefined;
     onUnmask?: (() => void) | undefined;
     disableMask?: boolean | undefined;
+    transformUppercase?: boolean;
```
keithtxw, line 16: Thinking for more flexibility, let's make it a string literal type
```
transform?: "uppercase" | "lowercase" | undefined;
```

keithtxw (changes requested):
Do also document down this new prop in the stories too

src/masked-input/types.ts
thyAshley: m just thinking if we need these transform, cause our nric validator fails any non-uppercase input, should uppercase be the default?

Hamiriza: I was thinking of the possibility of this component being reused for other purposes such as password field, so one way is to just pass the props later on, depending on the validator behaviour. I think going for flexibility would make more sense now?

src/masked-input/masked-input.tsx
```diff
@@ -57,7 +58,15 @@ const Component = (
     };
 
     const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-        setUpdatedValue(event.target.value);
+        let value = event.target.value;
```
qroll, line 61: onChange still receives the original `event.target.value`, so you may want to assign the value on the event object

src/masked-input/types.ts
keithtxw: yup agree. I am seeing this more than just for nric. The user for nric fields should set the prop to `uppercase`

stories/form/form-masked-input/props-table.tsx
```diff
@@ -82,6 +82,12 @@ const DATA: ApiTableSectionProps[] = [
                 description: "The callback function when field is not masked",
                 propTypes: ["() => void"],
             },
+            {
+                name: "transformInput",
+                description:
+                    "Specifies whether the input is tranformed into uppercase or lowercase",
```
keithtxw, line 88:
```suggestion
                    "Specifies whether the input value is to be tranformed into uppercase or lowercase format",
```

stories/form/form-masked-input/form-masked-input.stories.mdx
```diff
@@ -40,6 +40,16 @@ import { Form } from "@lifesg/react-design-system/form";
                     value="S1234567D"
                     maskRange={[2, 5]}
                 />
+                <Form.MaskedInput
+                    label="This is the mask field with uppercase input"
```
keithtxw, line 44:
```suggestion
                    label="This is set to be transformed to uppercase"
```

stories/form/form-masked-input/form-masked-input.stories.mdx
```diff
@@ -40,6 +40,16 @@ import { Form } from "@lifesg/react-design-system/form";
...
+                    maskRange={[2, 5]}
+                    transformInput={"uppercase"}
+                />
+                <Form.MaskedInput
+                    label="This is the mask field with lowercase input"
```
keithtxw, line 49:
```suggestion
                    label="This is set to be transformed to lowercase"
```

keithtxw (changes requested):

keithtxw (approved):

PR #329, opened 2023-10-12
[WX] add collapsible feature in accordion

wx-uf:
**Changes**
Description of changes...

- [delete] branch

**Changelog entry**

- Add option for `Accordion.Item` to always be expanded

### DIFF
--- a/src/accordion/accordion-item.tsx
+++ b/src/accordion/accordion-item.tsx
@@ -27,6 +27,7 @@ function Component(
         children,
         expanded,
         type = "default",
+        collapsible = true,
         ...otherProps
     }: AccordionItemProps,
     ref: React.Ref<AccordionItemHandle>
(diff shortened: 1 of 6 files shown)


src/accordion/accordion-item.tsx
qroll, line 1: the main accordion has a "Show all"/"Hide all" button which also affects the expanded state. we should prevent an uncollapsible item from collapsing when this button is clicked

the check can be added to the existing useEffect
```
    useEffect(() => {
        if (hasFirstLoad) {
            setExpand(collapsible ? expandAll : true);
        }
    }, [expandAll]);
```

qroll, line 1: similarly, `expanded` and `expandAll` should be ignored when `collapsible` is true

this check should be added to line 40 where `expand` is initialised in `useState`

qroll (changes requested):

qroll (approved):

PR #332, opened 2023-10-17
[WX]add maxlines to Text

wx-uf:
**Changes**
Description of changes...

- [keep | delete] branch

**Changelog entry**

- add new props maxLines to Text in order to specify the number of lines visible, the additional lines will be truncated

**Additional information**

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-382)

### DIFF
--- a/src/text/helper.ts
+++ b/src/text/helper.ts
@@ -77,7 +77,7 @@ const getFontWeight = (weight: TextWeight) => {
 // =============================================================================
 // TEXT STYLE
 // =============================================================================
-const getMaxLinesLineStyle = (maxLines: number | undefined = undefined) => {
+const getMaxLinesLineStyle = (maxLines: number | undefined) => {
     if (maxLines > 0) {
         return css`
             display: -webkit-box !important;
--- a/src/text/helper.ts
+++ b/src/text/helper.ts
@@ -77,6 +77,18 @@ const getFontWeight = (weight: TextWeight) => {
 // =============================================================================
 // TEXT STYLE
 // =============================================================================
+const getMaxLinesLineStyle = (maxLines: number | undefined = undefined) => {
+    if (maxLines > 0) {
+        return css`
+            display: -webkit-box !important;
+            -webkit-box-orient: vertical;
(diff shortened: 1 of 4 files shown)


src/text/helper.ts
```diff
@@ -77,6 +77,18 @@ const getFontWeight = (weight: TextWeight) => {
 // =============================================================================
 // TEXT STYLE
 // =============================================================================
+const getMaxLinesLineStyle = (maxLines: number | undefined = undefined) => {
```
qroll, line 80: the internal function can be more explicit about the number of args required

```suggestion
const getMaxLinesLineStyle = (maxLines: number | undefined) => {
```

src/text/helper.ts
```diff
@@ -77,6 +77,18 @@ const getFontWeight = (weight: TextWeight) => {
...
+    if (maxLines > 0) {
+        return css`
+            display: -webkit-box !important;
```
qroll, line 83: is this `!important` required?

src/text/helper.ts
```diff
@@ -77,6 +77,18 @@ const getFontWeight = (weight: TextWeight) => {
...
+            -webkit-box-orient: vertical;
+            -webkit-line-clamp: ${maxLines};
+            line-clamp: ${maxLines};
```
qroll, line 86: I think `line-clamp` is not an official css property yet, is this more for future compatibility?

src/text/helper.ts
wx-uf: yes, need, otherwise display: 'block' overwrite it based on the testing.

src/text/helper.ts
wx-uf: yes, pls advice remain or remove

qroll: I noticed the styled components plugin recognises this property so it's fine

src/text/helper.ts
wx-uf: removed and code submitted

qroll (approved):

PR #334, opened 2023-10-18
Add Divider component

keithtxw:
**Changes**
- Add Divider component
- Had to introduce own props since the props for horizontal rule html component are deprecated
- Use svg for dashed style due to specific requirements that UX needs

- [delete] branch

**Changelog entry**

- Add `Divider` component to enable rendering of horizontal rules or dividers in pages

**Additional information**

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-491)
- [Figma](https://figma.fun/q4fP3U)

### DIFF
--- a/src/divider/divider.style.tsx
+++ b/src/divider/divider.style.tsx
@@ -1,6 +1,5 @@
 import styled, { css } from "styled-components";
 import { Color } from "../color/color";
-import { Layout } from "../layout";
 import { DividerLineStyleType } from "./types";
 
 // =============================================================================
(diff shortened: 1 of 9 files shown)


keithtxw:
New requirement came back for custom dashed gap. Will await designer to pass the assets

src/divider/divider.style.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+    return css`
+        display: block;
+        width: 100%;
+        background-color: transparent;
+        margin: 0;
+        height: ${thickness}px;
+        border: none;
+        background-repeat: repeat-x;
+        background-image: url('data:image/svg+xml,<svg width="8" height="${thickness}" viewBox="0 0 8 1" xmlns="http://www.w3.org/2000/svg"><line x1="2" y1="1" x2="8" y2="1" stroke="${encodedColor}" stroke-width="${strokeWidth}" stroke-dasharray="4 4" /></svg>');
```
qroll, line 54: instead of using different components for each line style, could this `styled.hr` and svg definition be reused? then I'm thinking the col div can just be conditionally rendered based on whether its layout is grid

qroll, line 54: is `x1="2"` intended to offset the starting position?

keithtxw: No longer relevant as the components have been merged into a single styled component

keithtxw: Yes correct as we observed some imbalance due to the 4px gap at the end of the dash. Hence we made it centered with 2px on each sides to achieve a more balanced look. Have amended the x2 to be 6 for correctness

qroll (approved):

PR #335, opened 2023-10-19
Refactor day cell

qroll:
**Changes**

- Introduce a base DayCell
- Migrate week calendar to use this component
- [delete] branch


qroll:
once the day calendar is migrated, will do clean up of unused files. also planning to put the variants under `week`, `days-of-month` and `fixed-range`

src/util/calendar-helper.ts
```diff
@@ -176,6 +176,34 @@ export namespace CalendarHelper {
...
+        const firstDayOfWeek = day.startOf("week").format("YYYY-MM-DD");
+        const lastDayOfWeek = day.endOf("week").format("YYYY-MM-DD");
+
+        return {
+            start: firstDayOfWeek,
+            end: lastDayOfWeek,
+        };
+    };
+
+    export const isDisabledDay = (
```
qroll, line 190: this was originally in `calendar-day-style-helper.ts`, moved it here as other variants will likely need this

src/shared/internal-calendar/week/week-day-cell.tsx
```diff
@@ -0,0 +1,136 @@
...
+            props.shadow = true;
+            props.circleShadow = isStart || isEnd;
+        } else if (isSelected) {
+            type = "selected-outline";
+        } else if (isHover) {
+            type = "hover-dash";
+        }
+
+        if (type) {
+            if (isStart) {
```
keithtxw, line 87: could this be 2 `if-else` statements?

keithtxw (changes requested):

src/shared/internal-calendar/week/week-day-cell.tsx
qroll: do you mean this?

```
if (isStart)
else if (isEnd)

if (!isStart)
else if (!isEnd)
```

the second one won't work because`!isStart` and `!isEnd` are not mutually exclusive, so Mon-Fri need both bgLeft and bgRight

keithtxw: no more like 
```
if (isStart) {
} else { }

if (isEnd) {
} else {}
```

src/shared/internal-calendar/day-cell/day-cell.style.tsx
```diff
@@ -0,0 +1,229 @@
...
+interface LabelStyleProps {
+    $type: LabelType;
+    $disabled: boolean;
+    $interactive: boolean;
+}
+
+// =============================================================================
+// HELPERS
+// =============================================================================
+const getCellStyle = (props, type: CellType) => {
```
keithtxw, line 23: Actually can just use the props and get `$type` from it? Were you trying to avoid dealing with the typings?

keithtxw (changes requested):
Sorry just spotted this.

src/shared/internal-calendar/day-cell/day-cell.style.tsx
qroll: oops this was a holdover from when the style props were distinct, the props arg can indeed access `$type` now

keithtxw (approved):

PR #343, opened 2023-10-26
Fix tests to use fake timers to support running tests in d…

keithtxw:
**Changes**
- Fix tests to use fake timers
- @daiproxomhoa to test locally

- [delete] branch


qroll:
any issues with `time-slot-bar-helper.spec.ts`? noticed that also uses dayjs for time formatting

tests/utils/calendar-helper.spec.ts
```diff
@@ -2,6 +2,15 @@ import dayjs, { Dayjs } from "dayjs";
 import { CalendarHelper } from "../../src/util/calendar-helper";
 
 describe("CalendarHelper", () => {
+    beforeAll(() => {
+        jest.useFakeTimers();
+        jest.setSystemTime(new Date("26 Oct 2023 12:00:00 GMT").getTime());
```
qroll, line 7: minor thing but a Date instance is also accepted so there's no need to convert it

```suggestion
        jest.setSystemTime(new Date("26 Oct 2023 12:00:00 GMT"));
```

keithtxw:
> any issues with `time-slot-bar-helper.spec.ts`? noticed that also uses dayjs for time formatting

Looking at the tests, it doesn't seem to have any time conversion elements involved.

qroll (approved):
tests passed on my side with a different timezone

qroll (approved):

qroll:
forcing a timezone so that the snapshot tests are generated consistently, referencing solutions from https://stackoverflow.com/questions/56261381/how-do-i-set-a-timezone-in-my-jest-config

note that we're not able to test timezones dynamically with this

PR #354, opened 2023-11-16
[CCUBE-681][DV] DateInput component - NaN display

daiproxomhoa:
Currently if the year has less than 4 digits, the input will display NaN, refer to image attached
To parse the date value so that the date will be displayed instead of NaN, even if the year has less than 4 digits


qroll:
as discussed previously, there are additional issues associated with handling three/two digit dates with `dayjs` that need follow up. the scope of this PR is to handle user selecting `31/12/999` from the calendar but does not resolve scenarios like `<Form.DateInput value="0099-01-01" />`

qroll (approved):

PR #361, opened 2023-11-21
Fix select onblur

Sharptail:
**Changes**
- moved `onBlur` into `handleWrapperBlur` 
- modified the typing of `onBlur`

- delete branch

**Changelog entry**

- Fixes onBlur behaviour for `Form.Select`

**Additional information**

- You may refer to this [MOL-13314](https://jira.ship.gov.sg/browse/MOL-13314)


src/input-select/input-select.tsx
```diff
@@ -69,6 +70,9 @@ export const InputSelect = <T, V>({
         }
         setShowOptions(!showOptions);
         triggerOptionDisplayCallback(!showOptions);
+        if (showOptions) {
+            onBlur(event);
+        }
```
qroll, line 75: onBlur is optional so optional chaining can be used here

also suggest applying the selector blur on the button instead of here so that it matches the visual focus

```
            <Selector
                onBlur={(e) => {
                    if (!showOptions) {
                        onBlur?.(e);
                    }
                }}
            >
```

src/input-select/types.ts
```diff
@@ -54,6 +54,7 @@ export interface InputSelectProps<T, V>
     valueToStringFunction?: ((value: V) => string) | undefined;
     /** Function to render selected custom component */
     renderCustomSelectedOption?: ((option: T) => JSX.Element) | undefined;
+    onBlur?: (e?: any) => void;
```
qroll, line 57: is the event arg still needed?

qroll (changes requested):
(mmoving my comments here for visibility)

one more thing, can this fix also be applied to the multi select? as it has the same dropdown behaviour

src/input-select/input-select.tsx
Sharptail: Actually this wont work because as mentioned `Selector` is a button and it's `onBlur` is the moment you release your click.

qroll: what behaviour did you observe with this handler logic?

was thinking it could cover the scenario where user tabs between form inputs

src/input-multi-select/input-multi-select.tsx
```diff
@@ -59,6 +60,9 @@ export const InputMultiSelect = <T, V>({
         if (!disabled) {
             setShowOptions(!showOptions);
             triggerOptionDisplayCallback(!showOptions);
+            if (showOptions) {
+                onBlur(event);
```
qroll, line 64: `onBlur` is an optional prop so optional chaining should be used here

src/input-select/types.ts
```diff
@@ -54,6 +54,7 @@ export interface InputSelectProps<T, V>
     valueToStringFunction?: ((value: V) => string) | undefined;
     /** Function to render selected custom component */
     renderCustomSelectedOption?: ((option: T) => JSX.Element) | undefined;
+    onBlur?: () => void;
```
qroll, line 57: to add the undefined typing per the conventions

```suggestion
    onBlur?: (() => void) | undefined;
```

qroll (approved):

PR #367, opened 2023-11-29
[MOL-13885][TK] add year limit to 1900 for dateinput

terrykkj:
**Changes**
Description of changes...
- add `const MIN_DATE_INPUT`
- delete branch

**Changelog entry**

- Fix year limit, user shouldn't be able to select to year before that

**Additional information**

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/MOL-13885)

### DIFF
--- a/src/date-input/date-input.tsx
+++ b/src/date-input/date-input.tsx
@@ -44,6 +44,7 @@ export const DateInput = ({
     const nodeRef = useRef<HTMLDivElement>(null);
     const calendarRef = useRef<InternalCalendarRef>();
     const inputRef = useRef<StandaloneDateInputRef>();
+    const MIN_DATE_INPUT = "1900-01-01";
     // =============================================================================
     // EFFECTS
     // =============================================================================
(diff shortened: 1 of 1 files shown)


src/date-input/date-input.tsx
```diff
@@ -176,7 +177,7 @@ export const DateInput = ({
                 withButton={withButton}
                 value={selectedDate}
                 disabledDates={disabledDates}
-                minDate={minDate}
+                minDate={MIN_DATE_INPUT}
```
keithtxw, line 180: What you are doing here replaces the `minDate` prop value entirely

src/date-input/date-input.tsx
```diff
@@ -44,6 +44,7 @@ export const DateInput = ({
     const nodeRef = useRef<HTMLDivElement>(null);
     const calendarRef = useRef<InternalCalendarRef>();
     const inputRef = useRef<StandaloneDateInputRef>();
+    const MIN_DATE_INPUT = "1900-01-01";
```
keithtxw, line 47: Instead of setting it as such, you should set "1900-01-01" over in the deconstruct of the props instead at Line 16

keithtxw (changes requested):

PR #379, opened 2023-12-18
Add red button variant

choo95dev:
**Changes**
Description of changes...

- [delete] branch

**Changelog entry**
- introduce a new prop `danger` to button
- introduce a new design-token prop called `DesignToken.Button.Danger` with the following props:

Props: 
1. backgroundColor
2. hover
3. primary
4. border

- You may refer to this [CCUBE-764](https://jira.ship.gov.sg/browse/CCUBE-764)


src/button/types.ts
```diff
@@ -2,13 +2,13 @@ import React from "react";
...
 
 type StyleType = "default" | "secondary" | "light" | "link";
-
 export interface ButtonProps
     extends React.ButtonHTMLAttributes<HTMLButtonElement> {
     /** The style type of the button. Values: "default" | "secondary" | "light" | "link" */
     styleType?: StyleType | undefined;
     /** Indicates if a loading spinner is to be displayed */
     loading?: boolean | undefined;
+    danger?: boolean | undefined;
```
keithtxw, line 11: Do you have any mockups that I can refer to? The reason this is not a variant is because there is the concept of light and secondary but in red is it?

stories/button/button.stories.mdx
```diff
@@ -73,6 +93,30 @@ The `Button` component comes in 2 sizes, a default and small size.
                 Loading
             </Button.Small>
         </Container>
+        <br />
+        <Container>
+            <Button.Default loading danger={true}>
```
keithtxw, line 98:
```suggestion
            <Button.Default loading danger>
```

stories/button/button.stories.mdx
```diff
@@ -35,6 +35,26 @@ The `Button` component comes in 2 sizes, a default and small size.
             <Button.Small styleType="link">Link</Button.Small>
             <Button.Small disabled>Disabled</Button.Small>
         </Container>
+        <br />
+        <Container>
+            <Button.Default danger={true}>Default</Button.Default>
```
keithtxw, line 40:
```suggestion
            <Button.Default danger>Default</Button.Default>
```

src/button/button.style.tsx
```diff
@@ -68,26 +76,32 @@ export const Main = styled.button<MainStyleProps>`
...
+                        : Color.Primary(props)};
                     border: 1px solid transparent;
 
                     ${MediaQuery.MaxWidth.mobileL} {
                         width: 100%;
                     }
 
                     span {
-                        color: ${Color.Neutral[8](props)};
+                        color: ${isDanger ? "white" : Color.Neutral[8](props)};
```
keithtxw, line 104: Neutral 8 is close to white. Any reason we don't use it?

keithtxw (changes requested):

src/button/button.style.tsx
choo95dev: updated to use Neutral 8.

 Was looking for the color , but saw there were some styles that uses "white" . So I went to follow it 

Realize ur right, Neutral 8 is #FFFFFF which is consider white already

src/button/button.style.tsx
```diff
@@ -25,14 +26,21 @@ export const Main = styled.button<MainStyleProps>`
     // BUTTON STYLE + TEXT COLOR
     // -----------------------------------------------------------------------------
     ${(props) => {
+        const isDanger = props.$buttonIsDanger;
+
         switch (props.$buttonStyle) {
             case "secondary":
                 return css`
                     background-color: ${Color.Neutral[8](props)};
```
keithtxw, line 34: Since it is wrapped by the `css` helper, you need not pass the `(props)` params already

src/button/types.ts
keithtxw, line 11: Could you add a jsdoc comment using the same description as the one in `props-table`. This is to allow the description to appear when one hovers the prop in code.

keithtxw (changes requested):

src/button/button.style.tsx
choo95dev: applied the suggestion

src/button/types.ts
choo95dev: applied

keithtxw (approved):

PR #381, opened 2023-12-21
Fix ErrorDisplay lighthouse warning

jiansheng-gt:
**Changes**
- add width and height for all preset images in ErrorDisplay
- export `imgAttributeHelper` to allow consumers to generate the same attributes

- [delete] branch

**Changelog entry**

- Add width and height attributes to ErrorDisplay image to fix lighthouse warning

**Additional information**
This PR fixes the following warning from lighthouse:
image

- this also prevents jumping of layouts when the image takes a longer time to load.
- do note that lighthouse does not show this warning if the image is in an iframe - use the [direct iframe url](https://www.life.gov.sg/react/iframe.html?args=&id=modules-errordisplay--default&viewMode=story) to trigger the warning

further reading here: [1](https://web.dev/articles/optimize-cls?utm_source=lighthouse&utm_medium=devtools#modern_best_practice_for_setting_image_dimensions) [2](https://www.smashingmagazine.com/2020/03/setting-height-width-images-important-again/) [3](https://jakearchibald.com/2022/img-aspect-ratio/#width--height-presentational-hints)

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/MOL-14604)

### DIFF
--- a/src/error-display/helper.ts
+++ b/src/error-display/helper.ts
@@ -1,22 +1,37 @@
 import { MediaWidths } from "../spec/media-spec";
 
-export interface ImagePathAttributes {
-    base: string;
-    md: string;
-    lg: string;
-    width: number;
-    height: number;
-}
+export namespace ErrorDisplayHelper {
+    export interface ImagePathAttributes {
+        /** base sized image path */
+        base: string;
 
-export const imgAttributeHelper = (
-    pathAttributes: ImagePathAttributes
-): React.ImgHTMLAttributes<HTMLImageElement> => {
-    const { base, md, lg, width, height } = pathAttributes;
-    return {
-        srcSet: `${base} 400w, ${md} 800w, ${lg} 1200w`,
-        src: lg,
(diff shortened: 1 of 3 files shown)


src/error-display/helper.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export interface ImagePathAttributes {
+    base: string;
+    md: string;
+    lg: string;
+    width: number;
+    height: number;
+}
+
+export const imgAttributeHelper = (
```
qroll, line 11: for the purpose of code organisation, can you move this type and util to a namespace called `ErrorDisplayHelper` and export that namespace instead? partly to avoid potential naming conflicts between public utility functions

qroll (changes requested):

src/error-display/helper.ts
jiansheng-gt: updated and added jsdoc

qroll (approved):

PR #390, opened 2024-01-09
Add callback for end of slider selection

qroll:
**Changes**

- To avoid triggering too many unnecessary change events, devs can choose to listen to the end of selection instead
- Also removed old `stories.mdx` files
- [delete] branch

**Changelog entry**

- Add `onChangeEnd` prop that is called when selection ends in `InputSlider`, `InputRangeSlider` and `HistogramSlider`


src/input-slider/input-slider.tsx
```diff
@@ -44,6 +51,7 @@ export const InputSlider = ({
             value={selection}
             numOfThumbs={1}
             onChange={handleChange}
+            onChangeEnd={handleChange}
```
weili-govtech, line 54: should be calling `handleChangeEnd()`

qroll: thanks for the catch, updated

weili-govtech (approved):

PR #391, opened 2024-01-10
Change TimeSlotBarWeek header to use DayCell

jhyeep:
**Changes**
- Added optional `onHoverOut` prop for `DayCell`, `null` type for `interactive` prop
- Modified `TimeSlotBarWeek` header to use `DayCell`
- [delete] branch

**Additional information**
-  Follow-up to the previous [PR](https://github.com/LifeSG/react-design-system/pull/384), to fix some minor discrepancies from the [Figma](https://www.figma.com/file/us90jOpWBahsqK2VAluPvD/Flagship-Design-System?type=design&node-id=12221%3A109215&mode=design&t=S9jp4nlUS6hTmYAm-1) design


src/time-slot-bar-week/time-slot-bar-week-days.tsx
```diff
@@ -152,27 +156,31 @@ export const TimeSlotBarWeekDays = ({
...
         }
 
-        styleCircleProps.$interactive = enableSelection && !disabled;
+        dayCellStyleProps.interactive = !enableSelection
+            ? null
+            : isHoverEnabled;
+
+        if (isHoverEnabled && day === hoverDay) {
+            dayCellStyleProps.circleLeft = "hover-dash";
+            dayCellStyleProps.circleRight = "hover-dash";
```
qroll, line 172: to use `hover-current` instead

src/time-slot-bar-week/time-slot-bar-week-days.tsx
```diff
@@ -152,27 +156,31 @@ export const TimeSlotBarWeekDays = ({
...
-            styleLabelProps.$disabledDisplay = true;
+            dayCellStyleProps.disabled = true;
         }
 
-        styleCircleProps.$interactive = enableSelection && !disabled;
+        dayCellStyleProps.interactive = !enableSelection
+            ? null
+            : isHoverEnabled;
+
+        if (isHoverEnabled && day === hoverDay) {
```
qroll, line 170: to use the built in comparator here
```suggestion
        if (isHoverEnabled && hoverDay && day.isSame(hoverDay, "day")) {
```

src/time-slot-bar-week/time-slot-bar-week-days.tsx
```diff
@@ -152,27 +156,31 @@ export const TimeSlotBarWeekDays = ({
     const generateStyleProps = (day: Dayjs) => {
```
qroll, line 156: to also include the greyed out styling for non-current months. it would be something like

```
 if (calendarDate.month() !== day.month()) {
    labelType = "unavailable";
}
```

`CellWeekText` should also apply this color

src/time-slot-bar-week/time-slot-bar-week-days.style.tsx
```diff
@@ -39,6 +26,14 @@ export const HeaderCellWeek = styled.div`
     flex: 1;
 `;
 
+export const CellWeekText = styled(Text.XSmall)<{ $disabled?: boolean }>`
```
qroll, line 29: extract the inline typing to a interface called `CellWeekTextStyleProps`

qroll (changes requested):

src/time-slot-bar-week/time-slot-bar-week-days.tsx
jhyeep: i think we should leave the greyed out style exclusive to disabled dates, having non-current month days greyed out can be confusing when they're actually available

qroll: I see, the current TimeSlotWeekView does indeed not grey out non-current month days. I may have confused it with the calendar

qroll (approved):

PR #415, opened 2024-02-28
Enhance status progress stepper

kokengthecoder:
**Changes**
- Display `Step x of x` with title in tablet and mobile view
- Only display the current step in tablet and mobile view
- Remove scrollable behaviour in mobile view

- [keep] branch

**Changelog entry**
- Display only the current `Step x of x` with title in tablet and mobile view without scrolling effect.

**Additional information**

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/MOL-14993)
- [Figma](https://www.figma.com/file/us90jOpWBahsqK2VAluPvD/Flagship-Design-System?type=design&node-id=17235-62640&mode=design&t=yemxIRAThM2umXNO-0)

### DIFF
--- a/src/progress-indicator/progress-indicator.style.tsx
+++ b/src/progress-indicator/progress-indicator.style.tsx
@@ -54,15 +54,17 @@ export const IndicatorBar = styled.div<IndicatorProps>`
     }};
 `;
 
-export const IndicatorTitle = styled(Text.BodySmall)<IndicatorProps>`
+export const IndicatorTitleDesktop = styled(Text.BodySmall)<IndicatorProps>`
     ${(props) => {
-        const { highlighted, isMobile } = props;
-        const color = isMobile
-            ? Color.Neutral[1]
-            : highlighted
-            ? Color.Primary
-            : Color.Neutral[4];
+        const { highlighted } = props;
+        const color = highlighted ? Color.Primary : Color.Neutral[4];
         return `color: ${color(props)};`;
+    }};
+`;
+
+export const IndicatorTitleMobile = styled(Text.BodySmall)`
+    ${(props) => {
+        return `color: ${Color.Neutral[1](props)};`;
     }};
 `;
 
--- a/src/progress-indicator/progress-indicator.style.tsx
+++ b/src/progress-indicator/progress-indicator.style.tsx
@@ -55,8 +55,6 @@ export const IndicatorBar = styled.div<IndicatorProps>`
 `;
 
 export const IndicatorTitle = styled(Text.BodySmall)<IndicatorProps>`
-    margin-right: 0.5rem;
-
     ${(props) => {
         const { highlighted, isMobile } = props;
         const color = isMobile
--- a/src/progress-indicator/progress-indicator.style.tsx
+++ b/src/progress-indicator/progress-indicator.style.tsx
@@ -57,7 +57,7 @@ export const IndicatorTitleDesktop = styled(Text.BodySmall)<IndicatorProps>`
     overflow-wrap: anywhere;
     ${(props) => {
         const { highlighted } = props;
-        const color = highlighted ? Color.Primary : Color.Neutral[4];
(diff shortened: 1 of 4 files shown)


qroll (changes requested):
code comments:
- `INDICATOR_BAR_WIDTH_MOBILE` and `INDICATOR_BAR_FADE_WIDTH_MOBILE` can be cleaned up as they are no longer in use
- could you add the deprecated flag to the fade props?
- similarly, add a notice in the prop table: "This attribute has no effect and will be removed in v3.0.0."

some styling comments:
- in mobile, the steps have a margin right of 0.5rem
- will be good to add a `overflow-wrap: anywhere` on the steps to allow long text to wrap nicely

src/progress-indicator/progress-indicator.style.tsx
```diff
@@ -109,22 +52,24 @@ export const IndicatorBar = styled.div<IndicatorProps>`
...
         const color = highlighted ? Color.Primary : Color.Neutral[4];
         return `color: ${color(props)};`;
     }};
 `;
 
+export const IndicatorTitleMobile = styled(Text.BodySmall)`
+    overflow-wrap: anywhere;
+    ${(props) => {
+        return `color: ${Color.Neutral[1](props)};`;
+    }};
```
qroll, line 70: since this is constant, we can simplify this to `color: ${Color.Neutral[1]};` without needing to nest a function

src/progress-indicator/progress-indicator.tsx
```diff
@@ -151,34 +74,56 @@ export const ProgressIndicator = <T,>({
...
-                    <Fade backgroundColor={fadeColor} position={"left"} />
-                )}
-                {showFadeRight && (
-                    <Fade backgroundColor={fadeColor} position={"right"} />
-                )}
-            </>
+            <Indicator
+                key={currentIndex}
+                aria-label={getAriaLabel(currentIndex, currentIndex)}
+                id={getId(currentIndex, currentIndex)}
```
qroll, line 109: noticed that the id is set on the bar and on the title text. should try to keep it unique, I would suggest to append a `-title` for the title

src/progress-indicator/progress-indicator.style.tsx
```diff
@@ -1,24 +1,18 @@
...
 // STYLE TYPES
 // =============================================================================
-interface FadeProps {
-    backgroundColor: string[];
-    position?: "left" | "right";
-}
 
 interface IndicatorProps {
     highlighted: boolean;
+    isMobile?: boolean;
```
qroll, line 15: this can be reverted as the prop is not used (since the components were split out)

src/progress-indicator/progress-indicator.style.tsx
```diff
@@ -109,22 +52,24 @@ export const IndicatorBar = styled.div<IndicatorProps>`
...
 `;
 
-export const IndicatorTitle = styled(Text.BodySmall)<IndicatorProps>`
-    margin-right: 0.5rem;
-
+export const IndicatorTitleDesktop = styled(Text.BodySmall)<IndicatorProps>`
+    overflow-wrap: anywhere;
     ${(props) => {
         const { highlighted } = props;
         const color = highlighted ? Color.Primary : Color.Neutral[4];
```
qroll, line 61: Figma shows N3 instead of N4, can confirm with UX again?

qroll (changes requested):

src/progress-indicator/progress-indicator.style.tsx
kokengthecoder: > The colour should be N3 as per figma. N4 on our white background fails colour contrast/accessibility standards.

Verified with UX, will update the change 👍

src/progress-indicator/progress-indicator.style.tsx
qroll: more of a code style issue but is functionally correct, will close for now

qroll (approved):

PR #420, opened 2024-03-04
Add masked field in uneditable section

keithtxw:
**Changes**
- Add mask and unmask features along with the loading and error states
- Extract mask function in `MaskedInput` to `StringHelper` to enable shared usage
- Extract mask attribute props in `MaskedInput` for shared usage
- [delete] branch

**Changelog entry**

- Implement mask and unmask features for items in the `UneditableSection`

**Additional information**

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-1097)
- [Figma reference](https://www.figma.com/file/GpnybGCMIH5YNs1Of3a6d8/ECDA-Subsidy-Application?type=design&node-id=4731-97442&mode=design&t=U9cCc6yBkVpMcqoZ-4)

### DIFF
--- a/src/uneditable-section/section-item.styles.tsx
+++ b/src/uneditable-section/section-item.styles.tsx
@@ -1,6 +1,10 @@
 import styled, { css } from "styled-components";
-import { UneditableSectionItemDisplayWidth } from "./types";
+import { Color } from "../color";
 import { MediaQuery } from "../media";
+import { ComponentLoadingSpinner } from "../shared/component-loading-spinner/component-loading-spinner";
+import { Text } from "../text";
+import { UneditableSectionItemDisplayWidth } from "./types";
+import { ExclamationTriangleIcon } from "@lifesg/react-icons/exclamation-triangle";
 
 // =============================================================================
 // STYLING INTERFACES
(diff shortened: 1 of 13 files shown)


src/uneditable-section/section-item.tsx
```diff
@@ -1,17 +1,166 @@
...
+        }
+    };
+
+    const renderContent = () => {
+        if (!displayMaskState) {
+            return <Text.Body>{value}</Text.Body>;
+        }
+
+        if (disableMaskUnmask) {
+            return <Text.Body>{getMaskedValue()}</Text.Body>;
```
qroll, line 145: based on the description of the `displayMaskState` prop, would still need to render the unmasked value if the `displayMaskState === "unmasked"`

src/uneditable-section/uneditable-section.tsx
```diff
@@ -15,15 +15,45 @@ export const UneditableSectionBase = ({
...
+                    <UneditableSectionItem
+                        key={index}
+                        {...item}
+                        {...(onMask ? { onMask: handleItemMask(item) } : {})}
+                        {...(onUnmask
+                            ? { onUnmask: handleItemUnmask(item) }
+                            : {})}
+                        {...(onTryAgain
+                            ? { onTryAgain: handleTryAgain(item) }
+                            : {})}
```
qroll, line 54: why not pass the callback props directly?

src/uneditable-section/section-item.styles.tsx
```diff
@@ -37,3 +41,63 @@ export const Container = styled.li<ContainerStyleProps>`
...
+    color: ${Color.Primary};
+    margin-left: 0.5rem;
+
+    svg {
+        width: 1.125rem;
+        height: 1.125rem;
+    }
+`;
+
+export const Clickable = styled.button`
```
qroll, line 59: we can apply a text overflow wrap here to handle long text
Screenshot 2024-03-06 at 9 23 01 AM

for such scenarios, is the icon expected to remain on the right?

src/uneditable-section/section-item.tsx
```diff
@@ -1,17 +1,166 @@
...
+                    </>
+                );
+            default:
+                return (
+                    <>
+                        <Text.Body>
+                            {displayMaskState === "masked"
+                                ? getMaskedValue()
+                                : value}
+                        </Text.Body>
```
qroll, line 126: this would render a p tag in the button which is not recommended, could cast to span or apply the text styling on the button

src/util/string-helper.ts
```diff
@@ -1,3 +1,5 @@
+import { MaskAttributeProps } from "../masked-input";
```
qroll, line 1: hmm feels odd to have a util import from a component, even if it's just typing. perhaps this component could expose it as an option instead? and let the components re-export as needed?

tests/uneditable-section/uneditable-section.spec.tsx
```diff
@@ -48,6 +48,169 @@ describe("UneditableSection", () => {
...
+                    items={ITEMS}
+                    title="Test"
+                    onUnmask={onUnmaskFn}
+                    onMask={onMaskFn}
+                />
+            );
+
+            act(() => {
+                fireEvent.click(screen.getByTestId("clickable-label"));
+                expect(onUnmaskFn).toHaveBeenCalledWith(ITEMS[0]);
```
qroll, line 179: typically assertions are done outside of the act

qroll (changes requested):

src/util/string-helper.ts
keithtxw: I found it awkward also but worried we would replicate. I think maybe it's more direct if i respecify the option types in the helper instead.

src/uneditable-section/section-item.styles.tsx
keithtxw: When you say expected to remain on the right, are u referring to alongside the overflowed text?

qroll: referring to columned layout vs inlined

qroll (approved):

PR #425, opened 2024-03-13
[MISC][KE] Fix custom timeline item title JSX element with styling

kokengthecoder:
**Changes**
Fixed timeline item title alignment if using custom JSX

- [delete] branch

**Changelog entry**
Fixed timeline item title alignment if using custom JSX

**Additional information**
It's very difficult to align the custom JSX title item due to the cascading styling of TimelineItemTitle in TimelineItem


qroll (approved):
changes look sound and the height of the line indicator is aligned more closely with the intended design. thanks!

PR #431, opened 2024-03-19
Image carousel magnifier and option to hide components

kokengthecoder:
**Changes**
- Added optional props to hide thumbnail, navigation, counter and magnifier, defaulting to false
- Added magnifier function to zoom image to fit device width or height based on the image relative size ratio to device screen size

- [delete] branch

**Changelog entry**
- Added optional props to hide thumbnail, navigation, counter and magnifier, defaulting to false
- Added magnifier function to zoom image to fit device width or height based on the image relative size ratio to device screen size

**Additional information**

- You may refer to this [ticket](https://jira.ship.gov.sg/browse/MOL-14812)
- [Figma](https://www.figma.com/file/CTn3KEPmSp7Pq3aFJKHC3l/%5BWeb%5D-%5BApp%5D--NS-Article-Templates?type=design&node-id=1247-104954&mode=design&t=dFqXsmYK0ylz5eJS-0)

**Solution explained:**

The `react-zoom-pan-pinch` library provides `centerView` function to zoom to the center of image, however we have to pass in the zoom ratio. 

Since the image is being scale down on load, we have to retrieve the true dimension of the image to compare to the device screen dimension. This is the reason why redux is implemented here, to keep the image dimension in state after loaded.

**Note:**
After checking with UX, the zoom position should be center center instead of top center stated in figma.


src/fullscreen-image-carousel/fullscreen-image-carousel.tsx
```diff
@@ -125,6 +142,36 @@ export const Component = (
...
+        const src = items[currentSlide].src;
+        const { width, height } =
+            fullscreenImageCarouselState.imagesDimension.find(
+                (i) => i.src == src
+            );
+
+        const isImgLandscapeRelativeToDevice =
+            height / width <
+            containerRef?.current.clientHeight /
+                containerRef?.current.clientWidth;
```
qroll, line 166: could check if `containerRef.current` is defined first

src/fullscreen-image-carousel/fullscreen-image-carousel.tsx
```diff
@@ -214,49 +261,75 @@ export const Component = (
...
+                <CloseButton
+                    aria-label="Close image carousel"
+                    onClick={onClose}
+                    focusHighlight={false}
+                >
+                    <CrossIcon aria-hidden />
+                </CloseButton>
+                {!hideMagnifier && (
+                    <MagnifierButton
+                        aria-label="Magnify image"
```
qroll, line 280: for UX to advise on the copy, but could the button aria label be `Zoom in` and `Zoom out` depending on the current zoom?

src/fullscreen-image-carousel/fullscreen-image-carousel.tsx
```diff
@@ -173,7 +220,7 @@ export const Component = (
                                         src={item.src}
                                         alt={item.alt ?? `Image ${index + 1}`}
                                         placeholder={<SlidePlaceholderImage />}
-                                        fit="scale-down"
```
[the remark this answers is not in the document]
kokengthecoder: Accidentally removed this line due to previous experiment 😅  will bring it back

kokengthecoder:
**Update:** 

- Migrate redux to local state
- Added zoom behaviour for small image (aka width and height is smaller than device's width and height)

qroll (approved):

PR #466, opened 2024-05-06
[RBS-1266] Update notification banner component

Mono143:
**Changes**

- added a new optional onClick field for NotificationBanner component
- added a new optional collapsedHeight field for NotificationBanner component
- [delete] branch

**Changelog entry**

- Added new optional onClick field for NotificationBanner component
- Added new optional collapsedHeight field for NotificationBanner component

**Additional information**

- You may refer to [RBS-1266](https://jira.ship.gov.sg/browse/RBS-1266)


src/notification-banner/notification-banner.tsx
```diff
@@ -40,23 +46,48 @@ export const NBComponent = ({
...
+        onClick();
+    };
+
     // =============================================================================
     // RENDER FUNCTIONS
     // =============================================================================
     if (!isVisible) return null;
 
+    const renderViewMore = () => (
+        <ViewMoreButton>
```
qroll, line 68: semantically, this should be a `styled.button`

this allows keyboard users or screen readers to interact with the overall banner

likely need a `event.stopPropagation` to prevent triggering the main banner again

src/notification-banner/notification-banner.tsx
```diff
@@ -40,23 +46,48 @@ export const NBComponent = ({
...
             <Container id={formatId("container", id)}>
                 <TextContainer>
                     <Content data-testid={formatId("text-content", testId)}>
-                        {children}
+                        <ContentWrapper
+                            $collapsedHeight={collapsedHeight}
+                        >
+                            {children}
+                        </ContentWrapper>
+                        {!isNil(collapsedHeight) && renderViewMore()}
```
qroll, line 90: the checks for `collapsedHeight` should be consistent - in styles, it's 
```
if (props.$collapsedHeight)
```
which doesn't allow for 0

checking for `!!collapsedHeight` should be sufficient here as setting 0 doesn't make sense

src/notification-banner/notification-banner.styles.tsx
```diff
@@ -94,3 +100,28 @@ export const StyledIcon = styled(CrossIcon)`
...
     color: ${Color.Neutral[8]};
 `;
+
+export const ContentWrapper = styled.div<ContentWrapperStyleProps>`
+    ${(props) => {
+        const gradient =
+            "linear-gradient(to bottom, black 50%, transparent 100%)";
+        if (props.$collapsedHeight)
+            return css`
+                height: ${props.$collapsedHeight}px;
```
qroll, line 110: should this be a `max-height` instead? if content is short, there will be empty space

src/notification-banner/types.ts
```diff
@@ -8,6 +8,13 @@ export interface NotificationBannerProps
     sticky?: boolean | undefined;
     onDismiss?: (() => void) | undefined;
     "data-testid"?: string | undefined;
+    /**
+     * Specifies the height to be displayed outside of
+     * the collapsible zone
+     */
+    collapsedHeight?: number | undefined;
```
qroll, line 15: since the constraint is on the content, would be good to be more explicit

```suggestion
    collapsedContentHeight?: number | undefined;
```

qroll (changes requested):

qroll:
not set → content can show fully, no `View more` button needed
`collapsedHeight` = 100px, content is 200px → gradient effect shown, `View more` button displayed

what's the expected behaviour here if `collapsedHeight` = 100px, content is 50px? it doesn't make sense to show the `View more` button if the content is fully visible, but if it's an outgoing content, user should be able to interact with the banner

it feels like the `View more` should be decoupled from the content height and banner click

src/notification-banner/notification-banner.tsx
```diff
@@ -40,35 +47,91 @@ export const NBComponent = ({
...
+                    {children}
+                </ContentWrapper>
+            );
+        }
+    };
+
     return (
-        <Wrapper ref={forwardedRef} $sticky={sticky} {...otherProps}>
+        <Wrapper
+            role={handleBannerClick ? "button" : "banner"}
```
qroll, line 117: in addition to role `button`, focus and key presses also need to be handled https://www.w3.org/WAI/ARIA/apg/patterns/button/examples/button/#kbd_label

hence it might be easier to render an invisible button instead

qroll (changes requested):

src/notification-banner/notification-banner.tsx
```diff
@@ -40,36 +47,94 @@ export const NBComponent = ({
...
             <Container id={formatId("container", id)}>
                 <TextContainer>
-                    <Content data-testid={formatId("text-content", testId)}>
-                        {children}
+                    <Content
+                        data-testid={formatId("text-content", testId)}
+                        $maxCollapsedHeight={maxCollapsedHeight} 
+                    >
+                        {renderChildren()}
+                        {actionButton && renderActionButton()}
```
qroll, line 132: the action button should be outside of `Content` as well, else it'll get collapsed too

you could move it to be a direct descendent of TextContainer, which then requires a `flex-direction: column`

src/notification-banner/notification-banner.tsx
```diff
@@ -40,36 +47,94 @@ export const NBComponent = ({
...
+            <StyledIcon />
+        </StyledIconButton>
+    );
+
+    const renderActionButton = () => (
+        <ActionButton
+            id={formatId("action-button", id)}
+            data-testid={formatId("action-button", testId)}
+            onClick={handleActionButtonOnClick}
+            {...actionButton}
```
qroll, line 93: the order should be swapped here so that the DS logic takes precedence

```suggestion
            {...actionButton}
            onClick={handleActionButtonOnClick}
```

qroll: after some testing, I think the click handlers can be simplified

```
// actionButton
if (!actionButton.onClick)
  return // do nothing, let it bubble to the banner
event.stopPropagation()
actionButton.onClick(event);

// accessible banner button
<AccessibleBannerButton /> // render without an onClick and the event bubbles to the banner as well

// handleBannerClick - not required, let the component accept `onClick` directly aka the native behaviour
```

qroll (approved):

PR #472, opened 2024-05-15
[RBS-1265][DS] minor styling fix for alert actionlink

Mono143:
**Changes**
- Minor styling fix to alert's actionLink when showMore is rendered

- [delete] branch

**Changelog entry**
- Minor styling fix to alert's actionLink when showMore is rendered

**Additional information**

- You may refer to [RBS-1265](https://jira.ship.gov.sg/browse/RBS-1265)

### DIFF
--- a/src/alert/alert.style.tsx
+++ b/src/alert/alert.style.tsx
@@ -15,6 +15,7 @@ interface StyleProps {
 }
 export interface TextWrapperContainerStyleProps {
     $showMore: boolean;
+    $hasActionLink: boolean;
     $maxCollapsedHeight?: number;
 }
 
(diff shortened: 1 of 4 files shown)


src/alert/alert.style.tsx
qroll, line 1: can you run prettier formatter on this page?

src/alert/alert.style.tsx
```diff
@@ -157,12 +158,17 @@ export const TextWrapperContainer = styled.div<TextWrapperContainerStyleProps>`
...
+            return css`
+            margin-bottom: 0.5rem;
+        `;
+    }}
     ${(props) => {
         const gradient =
             "linear-gradient(to bottom, black 50%, transparent 100%)";
         if (!props.$showMore && props.$maxCollapsedHeight)
-            return `
+            return css`
```
qroll, line 171: there's no real need to use `css` here as we're returning plain strings, we usually only need to do so for those containing nested functions or themes

qroll (changes requested):
one more thing that should be fixed: the `sizeType="small"` variant  should also be applied to the "Show more" button

qroll (approved):

PR #474, opened 2024-05-23
Create component to apply default HTML styles for JSX children

qroll:
**Changes**

Styles the common HTML tags `strong`, `a`, `ul`, and `ol`
Useful for components that accept JSX.Element as children

- [delete] branch

**Changelog entry**

- Introduce `Markup` to apply default styles for HTML markup

**Additional information**

- You may refer to this [CCUBE-1360](https://jira.ship.gov.sg/browse/CCUBE-1360)

### DIFF
--- a/src/shared/html-content/html-content.ts
+++ b/src/shared/html-content/html-content.ts
@@ -1,25 +1,27 @@
 import { css } from "styled-components";
 import { Color } from "../../color";
+import { FontFamily } from "../../spec/text-spec/font-spec";
 import { TextLinkSizeType, TextSizeType, TextStyleHelper } from "../../text";
 
-interface HtmlContentStyleOptions {
-    textColor?: string | ((props: any) => string);
+export interface HtmlContentStyleOptions {
+    textSize?: TextSizeType | TextLinkSizeType | undefined;
+    textColor?: string | ((props: any) => string) | undefined;
 }
 
-export const applyHtmlContentStyle = (
-    textStyle: TextSizeType | TextLinkSizeType,
-    options?: HtmlContentStyleOptions
-) => {
-    const { textColor = Color.Neutral[1] } = options || {};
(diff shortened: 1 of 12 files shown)


src/shared/html-content/html-content.ts
```diff
@@ -1,25 +1,27 @@
 import { css } from "styled-components";
 import { Color } from "../../color";
+import { FontFamily } from "../../spec/text-spec/font-spec";
 import { TextLinkSizeType, TextSizeType, TextStyleHelper } from "../../text";
 
-interface HtmlContentStyleOptions {
-    textColor?: string | ((props: any) => string);
+export interface HtmlContentStyleOptions {
+    textSize?: TextSizeType | TextLinkSizeType | undefined;
+    textColor?: string | ((props: any) => string) | undefined;
```
weili-govtech, line 8: why the need to pass textColor? technically we can style the color after `applyHtmlContentStyle()` like

```
const MyComponent = styled.div`
   ${applyHtmlContentStyle({ textSize: "Body"});
   color: Color.Neutral[1];
}
`;
```

similarly for `Markup`, if we want to change the colour, we will turn it into a styled-component

anyway if this is intended, we can introduce generics to the textColor function to allow devs to narrow the arguments `(<T = any>(props: T) => string)`

weili-govtech (changes requested):

src/shared/html-content/html-content.ts
qroll: agreed, it's not necessary for `applyHtmlContentStyle` now that it can inherit the text color

for the `Markup` component it's more for convenience

qroll: > we can introduce generics to the textColor function to allow devs to narrow the arguments

the function is for theme colors, so the idea is we can specify `Color.Primary` etc. and the props type doesn't matter

weili-govtech (approved):

PR #479, opened 2024-06-06
Add small variant to InputSelect

raymondtanwl:
**Changes**
Add small variant to InputSelect

**Changelog entry**
Add small variant to InputSelect

**Additional information**

- You may refer to this [[BOOKINGSG-5642](https://jira.ship.gov.sg/browse/BOOKINGSG-5642)]

### DIFF
--- a/src/shared/dropdown-list/dropdown-list.styles.tsx
+++ b/src/shared/dropdown-list/dropdown-list.styles.tsx
@@ -4,8 +4,13 @@ import styled, { css } from "styled-components";
 import { Checkbox } from "../../checkbox";
 import { Color } from "../../color";
 import { MediaQuery } from "../../media";
-import { Text, TextStyleHelper } from "../../text";
-import { LabelDisplayType, TruncateType } from "./types";
+import { TextStyleHelper } from "../../text";
+import {
+    DropdownVariantType,
+    IconProps,
+    LabelDisplayType,
+    TruncateType,
+} from "./types";
 
 // =============================================================================
 // STYLE INTERFACE
@@ -21,12 +26,18 @@ interface ListItemProps {
 interface ListItemSelectorProps {
     $multiSelect?: boolean;
     $hasNextLineLabel?: boolean;
+    $variant?: DropdownVariantType;
 }
 
 interface LabelProps {
     $truncateType?: TruncateType;
     $maxLines?: number;
     $labelDisplayType?: LabelDisplayType;
+    $variant?: DropdownVariantType;
+}
+
+interface ResultStateProps {
+    $variant?: DropdownVariantType;
 }
 
 // =============================================================================
@@ -87,6 +98,19 @@ export const ListItem = styled.li<ListItemProps>`
     }}
 `;
 
+const getMinHeight = (props: ListItemSelectorProps) => {
+    let minHeight = 3.5;
+    if (props.$variant === "small") {
+        minHeight = 3.25;
+    }
+
+    if (props.$hasNextLineLabel) {
+        minHeight = 4.255;
+    }
+
+    return minHeight;
+};
+
 export const ListItemSelector = styled.button<ListItemSelectorProps>`
     display: flex;
     ${(props) => {
@@ -97,7 +121,7 @@ export const ListItemSelector = styled.button<ListItemSelectorProps>`
         } else {
             return css`
                 padding: 0 1rem;
-                min-height: ${props.$hasNextLineLabel ? "4.255rem" : "3.5rem"};
+                min-height: ${getMinHeight(props)}rem;
                 align-items: center;
             `;
         }
@@ -134,7 +158,11 @@ const lineClampCss = css<LabelProps>`
 `;
 
 export const PrimaryText = styled.div<LabelProps>`
-    ${TextStyleHelper.getTextStyle("Body", "regular")}
+    ${(props) =>
+        TextStyleHelper.getTextStyle(
+            props.$variant === "small" ? "BodySmall" : "Body",
+            "regular"
+        )}
     color: ${Color.Neutral[1]};
     width: 100%;
 
@@ -223,8 +251,12 @@ export const SelectAllContainer = styled.div`
     padding: 1rem 0 0.5rem 0;
 `;
 
-export const DropdownCommonButton = styled.button`
-    ${TextStyleHelper.getTextStyle("Body", "semibold")}
+export const DropdownCommonButton = styled.button<LabelProps>`
+    ${(props) =>
+        TextStyleHelper.getTextStyle(
+            props.$variant === "small" ? "BodySmall" : "Body",
+            "semibold"
+        )}
     background-color: transparent;
     background-repeat: no-repeat;
     border: none;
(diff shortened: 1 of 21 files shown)


src/shared/dropdown-list/dropdown-list.styles.tsx
```diff
@@ -246,10 +272,20 @@ export const ResultStateContainer = styled.div`
 `;
 
 export const ResultStateText = styled(Text.Body)``;
+export const ResultStateTextSmall = styled(Text.BodySmall)``;
 
-export const LabelIcon = styled(ExclamationCircleFillIcon)`
+export const LabelIcon = styled(ExclamationCircleFillIcon)<IconProps>`
+    ${(props) => {
+        let _size = 1.5;
```
qroll, line 279: no need to prefix variables

```suggestion
        let size = 1.5;
```

qroll (changes requested):

src/shared/dropdown-list/dropdown-list.styles.tsx
```diff
@@ -87,6 +98,19 @@ export const ListItem = styled.li<ListItemProps>`
     }}
 `;
 
+const _getMinHeight = (props: ListItemSelectorProps) => {
```
qroll, line 101: do update the other instances, the prefix is not needed

```suggestion
const getMinHeight = (props: ListItemSelectorProps) => {
```

src/shared/dropdown-list/dropdown-list.styles.tsx
```diff
@@ -87,6 +98,19 @@ export const ListItem = styled.li<ListItemProps>`
...
+    let _minHeight = 3.5;
```
qroll, line 102:
```suggestion
    let minHeight = 3.5;
```

src/shared/dropdown-list/dropdown-search.styles.tsx
```diff
@@ -12,29 +18,53 @@ export const Container = styled.li`
     align-items: center;
 `;
 
-export const SearchInput = styled(BasicInput)`
-    height: 3rem;
+const _commonSearchInputStyle = `
```
qroll, line 21: can revert this change since it's only used in 1 place

src/shared/dropdown-wrapper/dropdown-wrapper.styles.tsx
```diff
@@ -18,49 +18,67 @@ export interface DropdownWrapperStyleProps {
 
 export interface ValueLabelStyleProps {
     truncateType?: TruncateType;
+    variant?: DropdownVariantType | undefined;
+}
+
+export interface SelectorStyleProps {
+    variant?: DropdownVariantType;
```
qroll, line 25: for non-HTML attributes, prefix them with `$` to mark them as transient props

please update for the other style interfaces

src/shared/dropdown-wrapper/dropdown-wrapper.styles.tsx
```diff
@@ -18,49 +18,67 @@ export interface DropdownWrapperStyleProps {
...
-    border-radius: ${BORDER_RADIUS};
-    border: none;
-    background: transparent;
-
-    :focus,
-    :active {
-        outline: none;
-    }
+export const baseSelectorCSS = (props?: SelectorStyleProps) => {
+    const height = _getHeight(props.variant);
```
qroll, line 50: instead of converting to a function, `css` can also take in props

```
export const baseSelectorCSS = css<SelectorStyleProps>`
  ${props => ...}
`
```

qroll (changes requested):

src/shared/dropdown-wrapper/dropdown-wrapper.styles.tsx
qroll: whoops sorrry if it wasn't clear

for component props, there's no need to prefix them with `$`

however for styled components, `$` is needed to mark them as transient, so that extra non-standard attributes are not passed down to the underlying HTML element

raymondtanwl: @qroll i have pushed a commit to remove `$` from the component props.
you can have a look again :)

src/shared/dropdown-list/dropdown-list.styles.tsx
```diff
@@ -245,11 +277,23 @@ export const ResultStateContainer = styled.div`
...
+        TextStyleHelper.getTextStyle(
+            props.$variant === "small" ? "BodySmall" : "Body",
+            "regular"
+        )}
+`;
 
-export const LabelIcon = styled(ExclamationCircleFillIcon)`
+export const LabelIcon = styled(ExclamationCircleFillIcon)<IconProps>`
+    ${(props) => {
+        // 1rem 16px, 1.5rem 24px
```
qroll, line 290: no need to add the comments on the unit conversion here, as the rem->px is a standard convention in this repo

src/shared/dropdown-list/dropdown-list.styles.tsx
```diff
@@ -245,11 +277,23 @@ export const ResultStateContainer = styled.div`
     align-items: center;
 `;
 
-export const ResultStateText = styled(Text.Body)``;
+export const ResultStateText = styled.text<ResultStateProps>`
```
qroll, line 280: I don't think text is a standard html element? render a div instead

```suggestion
export const ResultStateText = styled.div<ResultStateProps>`
```

src/shared/dropdown-list/dropdown-list.tsx
```diff
@@ -460,19 +468,29 @@ export const DropdownList = <T, V>({
...
+
             return (
                 <ResultStateContainer key="loading" data-testid="list-loading">
-                    <Spinner $buttonStyle="secondary" size={24} />
-                    <ResultStateText>Loading...</ResultStateText>
+                    <Spinner $buttonStyle="secondary" size={spinnerSize} />
+                    <ResultStateText $variant={variant}>
+                        Loading...
+                    </ResultStateText>
+                    ;
```
qroll, line 493: unintended semicolon

src/shared/dropdown-list/dropdown-search.styles.tsx
```diff
@@ -12,29 +24,43 @@ export const Container = styled.li`
...
     color: ${Color.Neutral[3]};
     cursor: pointer;
+    ${(props) => {
+        if (props.$variant === "small")
+            return css`
+                svg {
+                    ${getCssHeightAndWidth(props.$variant)}
+                }
+            `;
+    }}
```
qroll, line 58: actually this is redundant as the CancelIcon has its own css styling already

src/shared/dropdown-wrapper/dropdown-wrapper.styles.tsx
```diff
@@ -18,27 +18,42 @@ export interface DropdownWrapperStyleProps {
...
+
+export interface SelectorStyleProps {
+    $variant?: DropdownVariantType | undefined;
 }
 
 // =============================================================================
 // STYLING
 // =============================================================================
 const BORDER_RADIUS = "4px";
+const getHeight = (variant?: DropdownVariantType | undefined) => {
```
qroll, line 32: small nitpick but can you add a newline between 31-32? to separate the two items

qroll (changes requested):
just some minor code style comments

separately, can I trouble you to squash your commits in this PR? there were various reverts done here, so this is to keep the commit history clean and make it easier to compare diffs with older code

src/shared/dropdown-list/dropdown-search.styles.tsx
raymondtanwl: i added this because its ClickableIcon set default svg height and width to 1.125rem
https://github.com/LifeSG/react-design-system/blob/55f2a59308303af3298c7c44a1ad0ba05e713eed/src/shared/clickable-icon/clickable-icon.style.tsx#L23

hmm shall i modify ClickableIcon to have the variant prop too?

qroll: then in that case, CancelButton can handle the `svg` styling for both variants, and CancelIcon can be removed entirely (the component can use CrossIcon directly instead)

I'd say ClickableIcon doesn't need a variant for now because there's actually no standardised sizing for icons

qroll:
with the latest changes, the variant prop doesn't seem to be available

raymondtanwl:
> with the latest changes, the variant prop doesn't seem to be available

as discussed, I have removed variant prop from DropdownStyleProps and instead added it specifically on InputSelectProps, InputMultiSelectProps and DropdownListProps

qroll (approved):

PR #486, opened 2024-06-13
Give precedence to mobile elements when hiding menu button

qroll:
**Changes**
Fixes a minor bug where the menu button is still visible even though the mobile items and action buttons are correctly hidden in the drawer

```
export const Test: StoryObj<Component> = {
    render: () => {
        return (
            <Navbar
                items={{
                    desktop: [{ id: "home", children: "Home" }],
                    mobile: [],
                }}
                actionButtons={{
                    desktop: [{ type: "download" }],
                    mobile: [],
                }}
                fixed={false}
            />
        );
    },
};
```

- [delete] branch

**Changelog entry**

Bug fix
- Let mobile items and action buttons take precedence when hiding `Navbar` mobile menu button

### DIFF
--- a/src/navbar/navbar.tsx
+++ b/src/navbar/navbar.tsx
@@ -105,26 +105,18 @@ const Component = <T,>(
         );
     };
 
-    const hasUncollapsibleActionButtons = () => {
-        if (actionButtons.mobile) {
-            const hasUncollapsibleItems = actionButtons.mobile.some(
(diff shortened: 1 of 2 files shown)


qroll:
`hideNavElements` should probably be deprecated at this point as you can simply pass empty `[]`

weili-govtech (approved):

PR #489, opened 2024-06-19
Amended the touch area of the cross button & button styles for the toast enhancements

ShIrInRuTh:
**Changes**
- Amended the touch area of the cross button 
- Amended the button to prevent the button label from wrapping

**Additional information**

- You may refer to this [CCUBE-1400](https://jira.ship.gov.sg/browse/CCUBE-1400)
- This PR is to fix issues raised by UX

### DIFF
--- a/src/toast/toast.styles.tsx
+++ b/src/toast/toast.styles.tsx
@@ -69,6 +69,7 @@ export const ContentWrapper = styled.div`
     gap: 2rem;
     flex: 1;
     justify-content: space-between;
+    margin-right: 4rem;
 
     ${MediaQuery.MaxWidth.mobileL} {
         display: flex;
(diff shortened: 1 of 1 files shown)


src/toast/toast.styles.tsx
```diff
@@ -136,6 +138,7 @@ export const ActionButton = styled(Button.Small)`
 export const DismissButton = styled(ClickableIcon)<StyleProps>`
     padding: 0.75rem;
     margin: -0.75rem;
+    align-items: stretch;
```
qroll, line 141: `align-items` should be applied to the flex container, which in this case is `Wrapper`

`align-self` is used for individual flex children

src/toast/toast.styles.tsx
```diff
@@ -69,6 +69,7 @@ export const ContentWrapper = styled.div`
     gap: 2rem;
     flex: 1;
     justify-content: space-between;
+    margin-right: 4rem;
```
qroll, line 72: this margin doesn't seem to be needed? since the button still occupies space

qroll (changes requested):

qroll (approved):
lgtm ✨

PR #493, opened 2024-06-26
`DateInput`: fix dismiss behaviour, enhance focus and blur handling

qroll:
**Changes**

- Restores dismiss behaviour for variant `withButton={true}`
  - It's supposed to reset to the previous value if the Done button is not clicked
- Tweak blur event handling in date input
  - After date selection, transfer focus to the main element. This prevents the user from losing their place in the form
  - Only trigger `onBlur` when user moves away from the field, instead of whenever the calendar is dismissed
- Ensure main element still appears focused when calendar is active
- [delete] branch

**Changelog entry**

Bug fix
- `DateInput`
  - Restore behaviour to reset to the previous value if dismissed without confirmation
  - Only trigger `onBlur` when leaving the field instead of calendar dismissal
  - Ensure main input still appears focused when calendar is active

### DIFF
--- a/src/date-range-input/date-range-input.tsx
+++ b/src/date-range-input/date-range-input.tsx
@@ -32,6 +32,7 @@ interface DateRangeInputState {
     calendarOpen: boolean;
     isStartDirty: boolean;
     isEndDirty: boolean;
+    focused: boolean;
 }
 
 const INITIAL_STATE: DateRangeInputState = {
@@ -43,6 +44,7 @@ const INITIAL_STATE: DateRangeInputState = {
     calendarOpen: false,
     isStartDirty: false,
     isEndDirty: false,
+    focused: false,
 };
 
 export const DateRangeInput = ({
@@ -85,6 +87,7 @@ export const DateRangeInput = ({
             calendarOpen,
             isStartDirty,
             isEndDirty,
+            focused,
         },
         actions,
     ] = useStateActions({
@@ -92,6 +95,16 @@ export const DateRangeInput = ({
         initialState: INITIAL_STATE,
         reducers: {
             blur: (state) => ({
+                ...state,
+                selectedStart: state.initialStart,
+                selectedEnd: state.initialEnd,
+                currentFocus: "none",
+                calendarOpen: false,
+                isStartDirty: false,
+                isEndDirty: false,
+                focused: false,
+            }),
+            dismiss: (state) => ({
                 ...state,
                 selectedStart: state.initialStart,
                 selectedEnd: state.initialEnd,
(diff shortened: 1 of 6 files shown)


stories/filter/doc-elements.tsx
```diff
@@ -41,18 +41,7 @@ export const SearchFilter = ({ mode, value, onChange }: Props<string>) => {
...
-        <div style={{ height: isFocused ? "33rem" : undefined }}>
-            <Form.DateInput
-                value={value}
-                onChange={(date) => onChange(date)}
-                onFocus={() => setIsFocused(true)}
-                onBlur={() => setIsFocused(false)}
-            />
-        </div>
-    );
+    return <Form.DateInput value={value} onChange={(date) => onChange(date)} />;
```
qroll, line 44: custom height handling is no longer needed as the date input can escape the filter container

src/shared/input-wrapper/input-wrapper.tsx
```diff
@@ -36,9 +57,9 @@ export const InputWrapper = styled.div<InputWrapperStyleProps>`
         props.$position === "right" ? "row-reverse" : "row"};
 
     :focus-within {
-        border: 1px solid ${Color.Accent.Light[1]};
-        box-shadow: ${DesignToken.InputBoxShadow};
+        ${defaultFocusCss}
     }
+    ${(props) => props.$focused && defaultFocusCss}
```
qroll, line 63: need to handle focus programmatically, as the floating element is not necessarily a child in the actual HTML DOM hierarchy

src/shared/dropdown-wrapper/element-with-dropdown.tsx
```diff
@@ -76,13 +77,11 @@ export const ElementWithDropdown = ({
         duration: 300,
     });
 
-    const click = useClick(context, { enabled });
-    const focus = useFocus(context, { enabled });
```
qroll, line 80: removed to give control to the parent

there may be cases where the main element is focused but popover should remain closed (e.g. toggling behaviour)

src/date-range-input/date-range-input.tsx
```diff
@@ -526,6 +528,7 @@ export const DateRangeInput = ({
         return (
             <Container
```
weili-govtech, line 529: shouldnt it have the same `onBlur` / `onFocus` / `$focused` updates like how you did for `DateInput`?

weili-govtech (changes requested):

src/date-range-input/date-range-input.tsx
weili-govtech: apologies to expand on this, what about all the consumers of the `InputWrapper`? should they follow the same behaviour? this includes
- `InputGroup`
- `TimePicker`
- `TimeRangePicker`

qroll: yup, blur/focus handling needs to be reworked across these components. however these can be handled separately, as I'm targeting to migrate them to also use floating-ui and standardise the behaviour

weili-govtech (approved):

PR #495, opened 2024-06-30
Date picker tests

qroll:
**Changes**

Add tests for `DateInput`
- Uses `user-event` library to simulate browser interaction more closely in the unit tests
- Introduce Playwright to verify actual browser behaviour with `floating-ui` library and to perform visual regression testing

Also took the chance to migrate Jest to v29 for better fake timers support.

- [delete] branch


qroll:
random test suites failing even though they are passing locally. investigating

qroll:
unit tests started to fail after installing playwright dependencies

PR #507, opened 2024-07-10
Fix breadcrumb flash

### LINKED ISSUE #465: Breadcrumb fade-to-white thing appears briefly on page load under 1200px screen width
**Component**

`Breadcrumb`

**Describe the bug**

The "fade-to-white" things on the left edge of the breadcrumb seems to be appearing / not hiding in time, when the page is first loaded, but only when the screen width is under 1200px (corresponding to the `MaxWidth.Tablet` breakpoint).

**To Reproduce**

Visit https://www.life.gov.sg/guides/marriage-support and refresh the page, or create a minimal example with just the `Breadcrumb` component. Refresh at different screen sizes.

**Screen Recordings**

https://github.com/LifeSG/react-design-system/assets/61947377/08ece4b8-4ca1-4cc0-8560-af633a68b5ae

https://github.com/LifeSG/react-design-system/assets/61947377/8ed7aa78-16ab-45fc-954c-3687b725c3c9

qroll:
Resolves #465

There are two different pieces of logic running here:
1. calculation of fade effect
2. scrolling to the last breadcrumb

Both can cause an initial flash of content as the component has to set these on mount

**Changes**

- disable fade effect at the start, only enable if content overflows
- rework fade logic
- repurpose `fadePosition` (not 100% sure what it did previously as it was always overridden)
- use `useLayoutEffect` instead of `useEffect` - which requires SSR handling and hence the introduction of `useIsomorphicLayoutEffect`
- add E2E test for scroll behaviour
- [delete] branch

**Changelog entry**

- Fix initial flash of content in `Breadcrumb` on tablet/mobile


nobeltws (approved):
lgtm~ thanks for the fix!

PR #510, opened 2024-07-12
Make it easier to reset spacing in `Toggle` composite section

qroll:
**Changes**

- There's a default padding on the `Toggle` composite section, but sometimes the children might need to take up the full width of the Toggle or remove the padding, which is not possible without negative margin hacks
- Moved the top-level padding to individual elements in the composite section and added a `data-id` to target the children container
- [delete] branch

**Changelog entry**

Feature
- Add `data-id` for `Toggle` composite section children and make it easier to reset spacing

### DIFF
--- a/src/toggle/toggle.tsx
+++ b/src/toggle/toggle.tsx
@@ -6,7 +6,6 @@ import { SimpleIdGenerator } from "../util";
 import {
     AlertContainer,
     Children,
-    CompositeSectionContainer,
     Container,
     ErrorContainer,
     ErrorList,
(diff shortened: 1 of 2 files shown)


src/toggle/toggle.tsx
```diff
@@ -304,11 +307,11 @@ export const Toggle = ({
     const renderCompositeSection = () => {
         return (
             compositeSectionChildren && (
-                <CompositeSectionContainer>
+                <div>
```
weili-govtech, line 310: this div does not seem to have any use anymore, unless we want to add a `data-id` to make it target-able too

qroll: hmm the flex column does lay the items out vertically, but thought it could group the items together

weili-govtech: I see, ok that's fine then, let's keep it

weili-govtech (approved):

PR #518, opened 2024-07-19
E2E tests A-B

qroll:
**Changes**

- add e2e tests for accordion, alert, box container
- add missing `data-testid` and `id` selectors for accordion props
- [delete] branch


e2e/box-container/__screenshots__/chrome-mobile/with-action-button--mount.png
qroll, line 1: first example is indeed wonky, to be addressed separately

weili-govtech:
I suppose we testing for features that cannot be covered in unit tests? 

I think we can also consider some granular features like size types in `Accordion.Item` and `Alert` and the visual treatment to max collapsed height in `Alert`.

qroll:
will be regenerating the screens for v3

Issue #443, opened 2024-04-11
Floating elements are cut off by parent with `overflow: hidden`


qroll:
**Describe the bug**
When a floating element is rendered in a parent container that has `overflow: hidden`, it gets cut off when it extends beyond the parent bounds

<img width="562" alt="Screenshot 2024-04-11 at 4 58 23 PM" src="https://github.com/LifeSG/react-design-system/assets/16853062/00fbf966-38e2-434e-b3c0-22eb314a9891">

**Expected behavior**
- The floating element should be able to display in full
- The floating element should be automatically repositioned if there is no space in the preferred position
- The animation should be replaced with a fade effect

**Affected components**

- [x] DateInput https://github.com/LifeSG/react-design-system/pull/444
- [x] DateRangeInput https://github.com/LifeSG/react-design-system/pull/444
- [x] Select https://github.com/LifeSG/react-design-system/pull/509
- [x] MultiSelect https://github.com/LifeSG/react-design-system/pull/520
- [x] NestedSelect https://github.com/LifeSG/react-design-system/pull/637
- [x] NestedMultiSelect https://github.com/LifeSG/react-design-system/pull/637
- [x] PredictiveTextInput https://github.com/LifeSG/react-design-system/issues/815
- [x] RangeSelect https://github.com/LifeSG/react-design-system/pull/948
- [x] InputGroup, PhoneNumberInput https://github.com/LifeSG/react-design-system/pull/825
- [x] Timepicker https://github.com/LifeSG/react-design-system/pull/952
- [x] TimeRangePicker https://github.com/LifeSG/react-design-system/pull/958
- [x] Popover https://github.com/LifeSG/react-design-system/pull/401
- [ ] ~~Table action bar~~

PR #520, opened 2024-07-24
Fix multiselect dropdown getting cut off

qroll:
**Changes**

Fix for MultiSelect in #443

- Updates `MultiSelect` to be similar to the new `InputSelect`
- Enhance list behaviour - only 1 item should be visually hovered at a time, so mouse hover + keyboard + focus should be considered together
- Further style adjustments to match the new Figma specs
- [delete] branch

**Changelog entry**

- Fix dropdown in `MultiSelect` getting cut off when it exceeds the parent container bounds

### DIFF
--- a/src/input-multi-select/input-multi-select.tsx
+++ b/src/input-multi-select/input-multi-select.tsx
@@ -217,7 +217,7 @@ export const InputMultiSelect = <T, V>({
                     ref={selectorRef}
                     disabled={disabled}
                     expanded={showOptions}
-                    id={internalId}
+                    listboxId={internalId}
                     readOnly={readOnly}
                     variant={variant}
                 >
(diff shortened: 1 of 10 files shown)


src/input-multi-select/input-multi-select.tsx
weili-govtech, line 1: is it intended that the `Select all` / `Clear all` cannot be reached via keyboard navigation?

on a separate note, for the search field, once use navigates away, it also cannot be reached via keyboard

qroll: it's still tab focusable. I have tweaked the button styles so it's more obvious when it has keyboard focus

but you'll need to Shift+Tab to return to the `Select all` button/search input from the list items

qroll: up arrow on the first list item will now focus the search field. thanks for highlighting the keyboard interaction!

weili-govtech (approved):

PR #532, opened 2024-08-08
[BOOKINGSG-5979][WX]add new component SelectHistogram

wx-uf:
1. add new component SelectHistogram in DS
2. add storybook and test cases accordingly

### DIFF
--- a/src/select-histogram/select-histogram.tsx
+++ b/src/select-histogram/select-histogram.tsx
@@ -15,7 +15,7 @@ import {
     PlaceholderLabel,
     Separator,
 } from "./select-histogram.styles";
-import { SelectHistogramSliderProps } from "./types";
+import { SelectHistogramProps } from "./types";
 
 export const SelectHistogram = ({
     alignment = "left",
@@ -36,7 +36,7 @@ export const SelectHistogram = ({
     renderRangeLabel,
     value,
     ...otherProps
-}: SelectHistogramSliderProps): JSX.Element => {
+}: SelectHistogramProps): JSX.Element => {
     const { interval, bins = [], renderEmptyView } = histogramSlider;
     // =============================================================================
     // CONST, STATE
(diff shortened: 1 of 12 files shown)


choo95dev (approved):

src/select-histogram/types.ts
```diff
@@ -0,0 +1,16 @@
+import { HistogramSliderProps } from "../histogram-slider";
+import { TruncateType } from "../shared/dropdown-list-v2/types";
+import { DropdownAlignmentType } from "../shared/dropdown-wrapper";
+
+export interface SelectHistogramSliderProps {
```
qroll, line 5: to also include the following optional props:
- `className`
- `readOnly`
- `onBlur` - can be typed without the event arg as `() => void`

src/select-histogram/types.ts
```diff
@@ -0,0 +1,16 @@
...
+    error?: boolean | undefined;
+    alignment?: DropdownAlignmentType | undefined;
+    dropdownZIndex?: number | undefined;
+    histogramSlider: HistogramSliderProps;
```
qroll, line 9: can pick specific props instead as not all are used or relevant

i.e.
`Pick<HistogramSliderProps, "bins" | "interval" | "renderEmptyView"> `

and move `onChange`, `rangeLabelPrefix` and `rangeLabelSuffix` to `SelectHistogramSliderProps`

src/select-histogram/select-histogram.tsx
qroll, line 1: to also include the `focused` handling with `onBlur` callback

src/select-histogram/select-histogram.tsx
```diff
@@ -0,0 +1,201 @@
...
+            {getDisplayValue()}
+        </LabelContainer>
+    );
+
+    const renderElement = () => {
+        return (
+            <InputBox
+                className={className}
+                data-testid={"select-histogram-selector"}
+                id="select-histogram-selector"
```
qroll, line 146: don't hardcode, use the props instead

qroll (changes requested):

qroll (approved):

PR #533, opened 2024-08-15
[MOL-15934][AT] Add support for an isReady state to handle situation when download is still being generated or not available to the user.

thyAshley:
**Changes**

**FileDownload Component**
Add a `isReady` prop to fileItemDownload which is defaulted to `true`
Add a check in handleDownload that prevents user from downloading the file when either isLoading  is true or when isReady is false

- delete branch

**Changelog entry**
Add support for an `isReady` state to handle situation when download is still being generated or not available to the user.

**Additional information**

- You may refer to this [MOL-15934](https://jira.ship.gov.sg/browse/MOL-15934)

**demo**

[Demo.webm](https://github.com/user-attachments/assets/0dd4fa4b-58fc-4422-bfc4-09b38a36dd2e)

### DIFF
--- a/src/file-download/types.ts
+++ b/src/file-download/types.ts
@@ -14,6 +14,8 @@ export interface FileItemDownloadProps {
     thumbnailImageDataUrl?: string | undefined;
     /** Indicates if text should be truncated */
     truncateText?: boolean | undefined;
+    /** Indicate if file is ready for download, defaults to true */
+    isReady?: boolean;
 }
 
 export type FileDownloadStyle = "bordered" | "no-border";
@@ -30,4 +32,6 @@ export interface FileDownloadProps {
     id?: string | undefined;
     /** Called when file item is clicked  */
     onDownload: (file: FileItemDownloadProps) => void | Promise<void>;
+    /** controlled loading state*/
+    isLoading?: boolean;
 }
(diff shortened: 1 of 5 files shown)


src/file-download/types.ts
```diff
@@ -14,6 +14,8 @@ export interface FileItemDownloadProps {
     thumbnailImageDataUrl?: string | undefined;
     /** Indicates if text should be truncated */
     truncateText?: boolean | undefined;
+    /** Indicate if file is ready for download, defaults to true */
+    isReady?: boolean;
```
qroll, line 18:
```suggestion
    ready?: boolean | undefined;
```

stories/file-download/file-download.stories.tsx
```diff
@@ -52,7 +52,7 @@ export const Default: StoryObj<Component> = {
 
         return (
             <FileDownload
-                fileItems={fileItems}
+                fileItems={[fileItems[2]]}
```
qroll, line 55: is this change intended? as it's for a separate story

stories/file-download/file-download.stories.tsx
qroll, line 1: the story should be included in the main `file-download.mdx` as well, with a short explanation describing the use case

src/file-download/types.ts
qroll, line 1: update `stories/file-download/props-table.tsx` with the new prop as well

qroll (changes requested):

stories/file-download/file-download.stories.tsx
thyAshley: oh shit, copy pasta lol

qroll (approved):

PR #534, opened 2024-08-15
[CCUBE-1515][MAHI]Deprecate theme to V2

mahidhar-reddy09:
Deprecate current theme version to V2


package.json
```diff
@@ -41,6 +41,7 @@
         "fabric": "^5.3.0",
         "immer": "^10.0.2",
         "react-slider": "^2.0.6",
+        "@lifesg/react-design-system": "^2.7.0-canary.3",
```
qroll, line 44: to revert this change

qroll (changes requested):

PR #536, opened 2024-08-15
[MOL-16390][Tung] Update touch area of accordion

QuocTung95:
**Changes**
move onClick={handleExpandCollapseClick} from ExpandCollapseButton to TitleContainer
create css cursor: pointer for TitleContainer

- delete branch

- You may refer to this [[MOL-16390](https://jira.ship.gov.sg/browse/MOL-16390)]
- **Demo**

https://github.com/user-attachments/assets/c439bc0d-b1c9-433f-960b-18f768f54e30

### DIFF
--- a/src/accordion/accordion-item.tsx
+++ b/src/accordion/accordion-item.tsx
@@ -148,7 +148,10 @@ function Component(
             $isCollapsed={expand}
             ref={elementRef}
         >
-            <TitleContainer onClick={handleExpandCollapseClick}>
+            <TitleContainer
+                $isCollapsed={collapsible}
+                onClick={collapsible ? handleExpandCollapseClick : undefined}
+            >
                 {renderTitle()}
                 {collapsible && (
                     <ExpandCollapseButton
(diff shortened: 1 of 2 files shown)


src/accordion/accordion-item.tsx
qroll, line 1: this needs to account for the `collapsible` prop

if `collapsible` is false, then the title shouldn't be interactive, and the cursor should remain as the default

qroll, line 1: can I confirm if this behaviour should apply to all accordions? or only in specific scenarios?

if the latter, then it should be configurable through a prop

qroll (changes requested):

src/accordion/accordion-item.tsx
QuocTung95: it should be apply for all, confirmed form UX

QuocTung95: But this morning QE said "now the scope has changed. I would like to clarify this with UX on exact accordions that requires this change" in the chat. So wait for me uppdate

QuocTung95: With new this Accordion we make it as default ( touch area whole title - confirmed ). And with BoxContainer we keep it as old version

QuocTung95: checked collapsible

qroll (approved):

PR #544, opened 2024-08-22
[BOOKINGSG-6086][JH] add optional timestamp prop for CountdownTimer

jhyeep:
**Changes**
Add optional `timestamp` prop for `CountdownTimer`. `timestamp` signifies the time at which the countdown should end, and takes precedence over `timer`

- delete branch

**Changelog entry**

- Add optional `timestamp` prop for `CountdownTimer` component

**Additional information**

- You may refer to this [BOOKINGSG-6086](https://jira.ship.gov.sg/browse/BOOKINGSG-6086)

### DIFF
--- a/src/countdown-timer/types.ts
+++ b/src/countdown-timer/types.ts
@@ -12,16 +12,11 @@ interface MobileOffset {
     top?: number | undefined;
 }
 
-export interface CountdownTimerProps
-    extends React.HTMLAttributes<HTMLDivElement> {
+interface CountdownTimerBaseProps extends React.HTMLAttributes<HTMLDivElement> {
     "data-testid"?: string | undefined;
     /** To show/play the countdown timer */
     show: boolean;
     fixed?: boolean | undefined;
-    /** Specifies the countdown timer (in seconds) */
-    timer: number;
-    /** Specifies the timestamp at which the countdown ends (milliseconds since Jan 1, 1970) */
-    timestamp?: number;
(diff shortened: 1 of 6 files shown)


src/countdown-timer/types.ts
```diff
@@ -20,6 +20,8 @@ export interface CountdownTimerProps
     fixed?: boolean | undefined;
     /** Specifies the countdown timer (in seconds) */
     timer: number;
+    /** Specifies the timestamp at which the countdown ends (milliseconds since Jan 1, 1970) */
+    timestamp?: number;
```
weili-govtech, line 24: as `timer` and `timestamp` are mutually exclusive, `timer` should not be mandatory. but we need to ensure either of them is specified at any time. can we use discriminated union to ensure either `timer` or `timestamp` is specified at any given time?

src/countdown-timer/use-timer.tsx
```diff
@@ -1,6 +1,10 @@
 import { useEffect, useState } from "react";
 
-export const useTimer = (seconds: number, isPlaying: boolean) => {
+export const useTimer = (
+    seconds: number,
+    isPlaying: boolean,
+    endTime?: number // Takes precedence over seconds
```
weili-govtech, line 6:
```suggestion
    endTime?: number | undefined // Takes precedence over seconds
```

weili-govtech (changes requested):

weili-govtech (approved):

PR #549, opened 2024-08-28
[BOOKINGSG-6086][JH] Adjust CountdownTimer logic to improve accuracy for timestamp

jhyeep:
**Changes**
- Changed `CountdownTimer` logic from a regular `setInterval` to `setTimeout`, to dynamically adjust the delay between executions based on drift time
- Also fixed a typo in the storybook props table for `Filter`. `toggleFilterButtonStyle` was incorrectly named as `toggleFilterButtonType`

- [delete] branch

**Additional information**

- You may refer to this [BOOKINGSG-6086](https://jira.ship.gov.sg/browse/BOOKINGSG-6086)

### DIFF
--- a/src/countdown-timer/use-timer.tsx
+++ b/src/countdown-timer/use-timer.tsx
@@ -8,7 +8,10 @@ export const useTimer = (
     // =============================================================================
     // CONST, STATE, REF
     // =============================================================================
-    const [remainingSeconds, setRemainingSeconds] = useState<number>(seconds);
+    // -1 to match Math.floor behavior (initial state only affects first displayed number)
+    const [remainingSeconds, setRemainingSeconds] = useState<number>(
+        seconds - 1
+    );
 
     // =============================================================================
     // EFFECTS
@@ -37,7 +40,7 @@ export const useTimer = (
     // HELPER FUNCTIONS
     // =========================================================================
     const calculateRemainingSecondsFromTimestamp = (timestamp: number) =>
-        Math.max(Math.ceil((timestamp - Date.now()) / 1000), 0);
+        Math.max(Math.floor((timestamp - Date.now()) / 1000), 0);
 
     return [remainingSeconds] as const;
 };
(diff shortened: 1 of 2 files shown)


src/countdown-timer/use-timer.tsx
weili-govtech, line 1: I dont quite get this change actually, if it's meant to count down for 10s, it should start from 10s? where does the extra 1s come from?

I did a simple console.log with setTimeout and can see the countdown running properly without this fix, what am I missing?

```
useEffect(() => {
    if (!isPlaying) return;
    if (endTime)
        setRemainingSeconds(
            calculateRemainingSecondsFromTimestamp(endTime)
        );

    const startTime = Date.now();
    const intervalId = setInterval(() => {
        const countdown = calculateRemainingSecondsFromTimestamp(
            endTime ?? startTime + seconds * 1000
        );

        setRemainingSeconds(countdown);
        if (countdown <= 0) {
            console.log("countdown ended");
            clearInterval(intervalId);
        }
    }, 1000);

    setTimeout(() => {
        console.log("15s");
    }, 15000);

    return () => clearInterval(intervalId);
}, [endTime, isPlaying, seconds]);
```

weili-govtech (changes requested):

src/countdown-timer/use-timer.tsx
jhyeep: you're right - i somehow mixed up ceilings, floors and timers that display milliseconds, sorry for the confusion. 

the behavior of the CountdownTimer is currently fine on the surface, but for `timestamp` due to the current 1000 interval logic the timer will likely have inaccuracies up to 1s. i'll be changing the PR to focus on fixing that instead

weili-govtech: ok no issues with the new logic. 

just want to point out that, I think by now you probably should be aware that the countdown will never be 100% accurate because it is based on the end user's computer performance. a slower computer will take a longer time to complete the setTimeout.

weili-govtech (approved):

PR #554, opened 2024-09-04
Border width style token

mahidhar-reddy09:
Added Border tokens

### DIFF
--- a/src/theme/border/specs/lifesg-border-set.ts
+++ b/src/theme/border/specs/lifesg-border-set.ts
@@ -1,0 +1,16 @@
+import { css } from "styled-components";
+import { BorderSet } from "../types";
+import { dashedBorderStyle } from "../helper";
+
+export const LifeSgBorderSet: BorderSet = {
+    "width-sm": 1,
+
+    solid: "solid",
+
+    "dashed-default": (
+        thickness: number | ((props: any) => string) = 1,
+        color: string | ((props: any) => string) = "black"
+    ) => {
+        return dashedBorderStyle(thickness, color);
+    },
+};
--- a/src/theme/border/specs/lifesg-border-set.ts
+++ b/src/theme/border/specs/lifesg-border-set.ts
@@ -1,9 +1,9 @@
-import { css } from "styled-components";
 import { BorderSet } from "../types";
-import { dashedBorderStyle } from "../helper";
+import { dashedBorderStyle } from "../theme-helper";
 
 export const LifeSgBorderSet: BorderSet = {
-    "width-sm": 1,
+    "width-010": 1,
+    "width-020": 2,
 
     solid: "solid",
 
--- a/src/theme/border/specs/lifesg-border-set.ts
+++ b/src/theme/border/specs/lifesg-border-set.ts
@@ -1,7 +1,7 @@
 import { BorderSet } from "../types";
 import { dashedBorderStyle } from "../theme-helper";
 import { StyledComponentProps } from "../../helpers";
-import { Colour } from "../../index";
+import { ColourSemantic } from "../../colour-semantic/theme-helper";
 
 export const LifeSgBorderSet: BorderSet = {
     "width-010": 1,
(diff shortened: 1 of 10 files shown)


src/theme/border/theme-helper.ts
```diff
@@ -0,0 +1,94 @@
...
+    },
+    defaultValue: "lifesg",
+};
+
+export const dashedBorderStyle =
+    (
+        thickness: number | ((props: any) => string),
+        color: string | ((props: any) => string)
+    ) =>
+    (props: StyleProps) => {
```
qroll, line 19: if you log the arguments, this is the styled component prop obj, so this should be typed to `StyledComponentProps`

src/theme/border/theme-helper.ts
```diff
@@ -0,0 +1,94 @@
...
+const BorderSpec: ThemeCollectionSpec<BorderCollectionMap, BorderScheme> = {
+    collections: {
+        lifesg: LifeSgBorderSet,
+    },
+    defaultValue: "lifesg",
+};
+
+export const dashedBorderStyle =
+    (
+        thickness: number | ((props: any) => string),
```
qroll, line 16: we can enforce thickness to be returned as a pixel number value. this way, we don't have to parse the resolved thickness

```suggestion
        thickness: number | ((props: any) => number),
```

src/theme/border/theme-helper.ts
```diff
@@ -0,0 +1,94 @@
...
+                    ${result}
+                `;
+            }
+
+            // If number make it return a pixel string
+            if (typeof borderValue === "number") {
+                const pixelBorderValue = `${borderValue}px`;
+                return css`
+                    ${pixelBorderValue}
+                `;
```
qroll, line 80: no need to wrap in `css`, a string can also be interpolated

qroll (changes requested):

src/theme/border/specs/lifesg-border-set.ts
```diff
@@ -0,0 +1,16 @@
...
+    "width-010": 1,
+    "width-020": 2,
+
+    solid: "solid",
+
+    "dashed-default": (
+        thickness: number | ((props: any) => string) = 1,
+        color: string | ((props: any) => string) = "black"
+    ) => {
+        return dashedBorderStyle(thickness, color);
```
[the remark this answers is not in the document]
mahidhar-reddy09: Hi rou ling I think i would need it because the default values might differ for differnt themes. So i left it for now.

src/theme/border/specs/lifesg-border-set.ts
```diff
@@ -1,5 +1,7 @@
 import { BorderSet } from "../types";
 import { dashedBorderStyle } from "../theme-helper";
+import { StyledComponentProps } from "../../helpers";
+import { Colour } from "../../index";
```
qroll, line 4: import from colour file instead to avoid circular dependency

src/theme/border/theme-helper.ts
```diff
@@ -10,31 +10,35 @@ const BorderSpec: ThemeCollectionSpec<BorderCollectionMap, BorderScheme> = {
     },
     defaultValue: "lifesg",
 };
+// to check if hex colour
+const isHexColor = (color: string): boolean => {
```
qroll, line 14: is there a need to check for hex format?

mahidhar-reddy09: The reason I did that is if I were to use the semantic colour straight then there is no need to encode it because the return is a hex string but when I were to pass in arguments like (1, "red") for the border then we need to encode the red to fit the svg. So thats why i added a check to see if its a hex.

src/theme/index.ts
```diff
@@ -16,32 +18,39 @@ export const Typography = TypographyValues;
 
 export const Animation = AnimationValues;
 
+export const BorderSet = BorderValues;
```
qroll, line 21:
```suggestion
export const Border = BorderValues;
```

src/theme/index.ts
```diff
@@ -1,4 +1,6 @@
+import { Border } from "../time-slot-bar/time-slot-bar.styles";
```
qroll, line 1: unintended import?

qroll (approved):

PR #559, opened 2024-09-11
Breakpoint, MediaQuery, spacing, radius tokens

mahidhar-reddy09:
Set Up V3 Breakpoint and MediaQuery.


src/theme/spacing/specs/lifesg-spacing-set.ts
qroll, line 1: probably a controversial choice, but after some discussion with designers, we'll go with `px` for spacing tokens as well. the rationale is that we generally want layouts to follow the device size and not scale with font size

https://www.joshwcomeau.com/css/surprising-truth-about-pixels-and-accessibility/ is a good read on `rem` vs `px`

src/theme/types.ts
```diff
@@ -22,6 +25,16 @@ export type { AnimationSet, AnimationSetOptions } from "./animation/types";
 
 export type { BorderSet, BorderSetOptions } from "./border/types";
 
+export type {
+    RadiusSet,
+    RadiusSetOptions as RadiusSetSetOptions,
+} from "./radius/types";
```
qroll, line 31: was this export intended to be renamed?

src/theme/types.ts
```diff
@@ -22,6 +25,16 @@ export type { AnimationSet, AnimationSetOptions } from "./animation/types";
...
+
+export type {
+    SpacingSet as SpacingSetType,
```
qroll, line 34: was this export intended to be renamed?

qroll (changes requested):

src/theme/breakpoint/types.ts
```diff
@@ -0,0 +1,45 @@
...
+    "lg-margin": number;
+    "xl-margin": number;
+    "xxl-margin": number;
+}
+
+export type BreakPointCollectionsMap = {
+    [key in BreakpointScheme]: BreakpointSet;
+};
+
+export type BreakPointSetOptions = Partial<BreakpointSet>;
```
qroll, line 45: can keep the naming consistent. Break**p**oint instead of Break**P**oint

src/theme/radius/theme-helper.ts
```diff
@@ -0,0 +1,36 @@
...
+    collections: {
+        lifesg: LifeSgRadiusSet,
+    },
+    defaultValue: "lifesg",
+};
+
+export const getRadius = (key: keyof RadiusSet) => {
+    return (props: StyledComponentProps): string => {
+        const theme = props.theme;
+        const radiusset: RadiusSet = getCollection(
```
qroll, line 16:
```suggestion
        const radiusSet: RadiusSet = getCollection(
```

qroll (approved):

PR #561, opened 2024-09-12
[BOOKINGSG-6120][JH] Add TimeRangePicker variant component

jhyeep:
**Changes**
-  `DropdownListV2`: Allow enter key to select items and added an optional `topScrollItem` prop for automatic positioning of scrollbar
- Moved the original `TimeRangePicker` code to `DialPicker` and added the `ComboboxPicker` component, which allows users to select a time from a dropdown list of options, or manually type in a time which is automatically parsed and converted into a valid time format (similar to google calendar)
- `TimeRangePicker` now loads `DialPicker` (default) or `Combobox` through a `variant` prop (`dial`/`combobox`)

- [delete] branch

**Changelog entry**

- Add `combobox` variant for `TimeRangePicker`

**Additional information**

- You may refer to this [BOOKINGSG-6120](https://jira.ship.gov.sg/browse/BOOKINGSG-6120)
- [Figma](https://www.figma.com/design/us90jOpWBahsqK2VAluPvD/Flagship-Design-System?node-id=33362-303074&node-type=section&t=HYL2DoxBnMNS5O59-0)


qroll:
the components can be used standalone without the `Form` wrapper, so the logic for toggling between variants should sit in the main `TimeRangePicker` component

can I propose this structure instead?

```
src
  time-range-picker
    dropdown-picker
      dropdown-picker.tsx
      dropdown-picker.styles.ts
    card-picker
      card-picker.tsx
      card-picker.styles.ts
    time-range-picker.ts
    common.styles.ts
    index.ts
```

src/time-range-picker/combobox-picker/combobox-picker.tsx
```diff
@@ -0,0 +1,389 @@
...
+            $readOnly={readOnly}
+        >
+            <RangeInputInnerContainer
+                error={error || !!validationError}
+                currentActive={
+                    activeTimeSelector === null ? "none" : activeTimeSelector
+                }
+            >
+                {/* From */}
+                <SelectorInput
```
qroll, line 315: do include `aria-label` on the inputs to clearly indicate the purpose of each field

src/shared/dropdown-list-v2/dropdown-list.tsx
```diff
@@ -203,6 +205,21 @@ export const DropdownList = <T, V>({
...
 
+    useEffect(() => {
+        if (topScrollItem === undefined) return;
+
+        // Delay to ensure render is complete
+        setTimeout(() => {
+            const index = listItems.indexOf(topScrollItem);
+            const focusedItem = listItemRefs.current[index];
+            // Align the item to top of scrollable container
+            if (nodeRef.current) {
```
qroll, line 216: should also check if `focusedItem` is defined - else you would get NaN

qroll (changes requested):

tests/time-range-picker/time-range-picker.spec.tsx
```diff
@@ -0,0 +1,555 @@
...
+                        errorMessage={"Error Message"}
+                    />
+                );
+
+                // "Error Message" should overwrite internal error
+                expect(screen.queryByText("Error Message")).toBeInTheDocument();
+            });
+        });
+
+        describe("parseInput tests", () => {
```
qroll, line 209: can split this out into its own file under `tests/utils/time-helper.spec.ts`

qroll:
found an additional edge case for the focus/blur behaviour:

1. focus end input
2. press Escape to dismiss the dropdown
3. press Tab

the expected behaviour is for the indicator to disappear and the blur event to fire

currently the indicator remains active even after moving to another input, and `onBlur` did not fire

qroll (approved):

tests/time-range-picker/time-range-picker.spec.tsx
qroll, line 209: can split this out into its own file under `tests/utils/time-helper.spec.ts`

qroll (changes requested):

qroll (approved):
the blur handling will need further fine-tuning to match the behaviour of the date input and select components, but nothing looks critical, so I'm approving the changes

thanks for the work on this component!

PR #566, opened 2024-09-16
Modal-aware inputs

qroll:
**Changes**

related to https://github.com/LifeSG/react-design-system/pull/565

- apply similar changes to `ElementWithDropdown` which is the base component for the floating-enhanced inputs
- [delete] branch

**Changelog entry**

- Enhance inputs to be modal-aware by default

### DIFF
--- a/src/shared/dropdown-wrapper/element-with-dropdown.tsx
+++ b/src/shared/dropdown-wrapper/element-with-dropdown.tsx
@@ -17,6 +17,7 @@ import {
 } from "@floating-ui/react";
 import { useRef } from "react";
 import { useResizeDetector } from "react-resize-detector";
+import { useFloatingChild } from "../../overlay/use-floating-context";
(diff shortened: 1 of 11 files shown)


src/shared/dropdown-wrapper/element-with-dropdown.tsx
```diff
@@ -59,7 +61,8 @@ export const ElementWithDropdown = ({
     onDismiss,
     renderElement,
     renderDropdown,
-    zIndex = 50,
+    customZIndex,
+    defaultZIndex = 50,
```
choo95dev, line 65: `defaultZIndex`, this can be define as a const in the file right? 

If they ever need their own zIndex, can just use the `customZIndex`. 

And if default can just declare in the file, custom can be reverted to zIndex but optional

choo95dev (approved):

PR #567, opened 2024-09-20
Accessibility improvements to PopoverTrigger and new inline popover

qroll:
**Changes**

- Enhancement to `PopoverTrigger`
  - Switch implementation to use floating-ui's built in interactions
  - This comes with keyboard support and focus trapping
- New `PopoverInline` component
  - Displays text/icon to trigger a popover, intended to be used inline within a block of text
  - Built on top of `PopoverTrigger` and so exposes similar props
  - Supports underline styling
  - The dashed style is to cater for MyLegacy's usage https://dev.designsystem.lifesg.io/storybook/mylegacy/index.html?path=/docs/components--tooltip
- [delete] branch

**Changelog entry**

- New `PopoverInline` to support text and/or icon as a popover trigger
- Improve `PopoverTrigger` accessibility with keyboard support and automatic focus trapping
  -  [WARNING] If you use `PopoverTrigger` for custom popovers, verify that your interaction is not affected

### DIFF
--- a/src/popover-v2/popover-trigger.tsx
+++ b/src/popover-v2/popover-trigger.tsx
@@ -1,24 +1,29 @@
 import {
+    FloatingFocusManager,
     FloatingPortal,
     autoUpdate,
     flip,
     limitShift,
     offset,
     shift,
+    useClick,
+    useDismiss,
     useFloating,
+    useHover,
+    useInteractions,
 } from "@floating-ui/react";
-import { useEffect, useRef, useState } from "react";
+import { useRef, useState } from "react";
 import { useMediaQuery } from "react-responsive";
 import { MediaWidths } from "../media";
 import { useFloatingChild } from "../overlay/use-floating-context";
 import { PopoverV2 } from "./popover";
 import { TriggerContainer } from "./popover-trigger.styles";
-import { PopoverV2TriggerProps } from "./types";
+import { PopoverV2TriggerProps, PopoverV2TriggerType } from "./types";
 
 export const PopoverTrigger = ({
     children,
     popoverContent,
-    trigger = "click",
+    trigger: _trigger = "click",
     position = "top",
     zIndex,
     rootNode,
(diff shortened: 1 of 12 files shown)


src/popover-v2/popover-trigger.tsx
```diff
@@ -46,63 +51,46 @@ export const PopoverTrigger = ({
                 limiter: limitShift(),
             }),
         ],
+        onOpenChange: (isOpen) => {
+            setVisible(isOpen);
+
+            // this callback is triggering twice on hover, the check here prevents extra calls
+            if (visible !== isOpen) {
+                handleVisibilityChange(isOpen);
+            }
```
weili-govtech, line 60: why not handle the visibility change in a useEffect? this way you wont need to call it at multiple places / check for previous state

stories/popover-v2/popover-inline/props-table.tsx
```diff
@@ -0,0 +1,50 @@
...
+            },
+            {
+                name: "icon",
+                description: "The icon to display",
+                propTypes: ["React.ReactNode"],
+            },
+            {
+                name: "underlineStyle",
+                description: "The underline style of the text",
+                propTypes: ["React.ReactNode"],
```
weili-govtech, line 36:
```suggestion
                propTypes: ["default", "underline", "underline-dashed"],
```

stories/popover-v2/popover-inline/props-table.tsx
```diff
@@ -0,0 +1,50 @@
...
+                defaultValue: `"default"`,
+            },
+            {
+                name: "underlineHoverStyle",
+                description: "The underline style of the text when hovered",
+                propTypes: ["React.ReactNode"],
```
weili-govtech, line 42:
```suggestion
                propTypes: ["default", "underline", "underline-dashed"],
```

weili-govtech (changes requested):

src/popover-v2/popover-trigger.tsx
qroll: oh it will work for this case, just that I generally prefer not to use `useEffect` for events. in a more complex components 1) I might not always want to trigger the effect and 2) it gets hard to trace the originating event

`useEffect` also triggers on mount which requires tracking the first render in this case

stories/popover-v2/popover-inline/props-table.tsx
qroll: thanks for catching this 😢

weili-govtech (approved):

PR #569, opened 2024-09-20
Support interactive elements in toggle label

qroll:
**Changes**

- The current implementation applies an `input` over the toggle label + sublabel to preserve the native accessibility and keyboard behaviour. But this prevents elements in the toggle from being targeted, e.g. tooltips
- This PR does the following:
  - Style input to be invisible and out-of-the-way. It can still be tabbed to via keyboard, and the screen reader is able to read out its associated label
  - Let clicks on the toggle container trigger the input change event
- [delete] branch

**Changelog entry**

- [WARNING] Modify click behaviour in `Toggle` to support interactive elements in label


qroll:
putting this on hold as Android Talkback is not detecting the checkbox/radio input, even though similar implementations on the web are working fine

PR #570, opened 2024-09-23
Set up V3 Typography component

mahidhar-reddy09:
Typography component with its test cases.


src/typography/helper.ts
```diff
@@ -0,0 +1,98 @@
...
+export const createTypographyStyles = (
+    textStyle: TypographySizeType,
+    props: TypographyProps
+) => css`
+    ${getTypographyStyle(
+        textStyle,
+        props.weight || "regular",
+        props.paragraph
+    )(props)}
+    color: ${Colour.Primitive["neutral-20"]};
```
qroll, line 96: use the semantic token here

src/typography/helper.ts
```diff
@@ -0,0 +1,98 @@
...
+            return css`
+                margin-bottom: ${fontSizeValue * marginBottomScale}${fontSizeUnit};
+            `;
+        };
+
+        return css`
+            font-size: ${fontSize};
+            line-height: ${attrs.lineHeight};
+            letter-spacing: ${attrs.letterSpacing || 0};
+            font-weight: ${resolvedWeight};
```
qroll, line 64: instead of individually picking out the styles, we can reuse the `Typography` tokens instead?

for example, HeaderXL will use `header-xl-<weight>` where weight comes from the prop

src/typography/types.ts
```diff
@@ -0,0 +1,51 @@
...
+    // For consumer to choose if they want the text to be inline for example
+    inline?: boolean;
+    // For consumer to choose for block level style
+    paragraph?: boolean;
+}
+
+export interface LinkProps extends TypographyProps {
+    // If the link is external
+    external?: boolean;
+    textStyle?: TypographySizeType;
```
qroll, line 50: how is this text style being used by the Link component?

src/typography/types.ts
```diff
@@ -0,0 +1,51 @@
...
+    | "regular"
+    | "semibold"
+    | "bold"
+    | "light"
+    | 400
+    | 600
+    | 700
+    | 300;
+
+export type TextStyleSetType = {
```
qroll, line 34: this seems to be unused

qroll (changes requested):

src/typography/types.ts
```diff
@@ -0,0 +1,51 @@
...
+    [key in TypographySizeType]: TypographyStyleSpec;
+};
+
+export interface TypographyProps extends React.HTMLAttributes<HTMLElement> {
+    // Can be any weight such as regular or 400
+    weight?: TypographyWeight;
```
qroll, line 40: set `| undefined` for the optional props

src/typography/helper.ts
mahidhar-reddy09: Just to check if i remove this if i am not wrong i would have to create a helper function or within the code itself do like a iteration to convert the header-xl etc values to css properties. Is that intended or is there another approach ?

qroll: you can simplify the current implementation of the Typography namespace, I am envisioning something like

```
export const HeaderXXL = createHeader("h1", "header-xxl");

// given HeaderXL + `weight="bold"`, it results in this:
styled(tag)`
  ${Typography["header-xl-bold"]}
`
```

tests/typography/typography-component.spec.tsx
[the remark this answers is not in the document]
qroll: can further trim the test cases - for example Body just requires the last unit test to assert the rendering behaviour. the other two are not required

src/typography/helper.ts
```diff
@@ -0,0 +1,101 @@
...
+        const getMarginBottomStyle = () => {
+            const marginBottomScale = paragraph ? 1.05 : 0;
+            return css`
+                margin-bottom: ${fontSizeValue * marginBottomScale}${fontSizeUnit};
+            `;
+        };
+
+        return css`
+            ${Typography[typographyKey]}
+            ${getMarginBottomStyle()}
```
qroll, line 68: there will be some changes to the paragraph styling (potentially a token), for now we can simplify the implementation

```suggestion
            margin-bottom: 1.05em;
```

qroll (changes requested):

src/typography/types.ts
```diff
@@ -0,0 +1,42 @@
...
+export interface TypographyProps extends React.HTMLAttributes<HTMLElement> {
+    // Can be any weight such as regular or 400
+    weight?: TypographyWeight | undefined;
+    // For consumer to choose if they want the text to be inline for example
+    inline?: boolean | undefined;
+    // For consumer to choose for block level style
+    paragraph?: boolean | undefined;
+}
+
+export interface LinkProps extends TypographyProps {
```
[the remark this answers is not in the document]
mahidhar-reddy09: Hi Ruo ling it is using the weight prop though so If i dont extend TypographyProps  I would need to express weight again. So should I extend TypographyProps or no need.

src/typography/helper.ts
mahidhar-reddy09: So there is no need for the function which calculates the margin bottom?

mahidhar-reddy09: Or should i leave it for future implementations!

src/typography/types.ts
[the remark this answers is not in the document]
qroll: extend `React.HTMLAttributes<HTMLAnchorElement>`, and copy the weight prop

src/typography/helper.ts
qroll: the function can be removed, basically no need to manually calculate it anymore

src/typography/helper.ts
```diff
@@ -0,0 +1,54 @@
+import { css } from "styled-components";
+import { TypographyProps, TypographyWeight } from "./types";
+import { Colour, Typography } from "../theme";
+import { StyledComponentProps } from "../theme/helpers";
+import { TypographySizeType } from "../theme/typography/types";
+
+export const getTypographyStyle = (
+    type: TypographySizeType,
+    weight: TypographyWeight | ((props: StyledComponentProps) => string),
```
qroll, line 9: when do we expect to pass in weight as a function?

src/typography/types.ts
```diff
@@ -0,0 +1,16 @@
+export type TypographyWeight = "regular" | "semibold" | "bold" | "light";
+
+export interface TypographyProps extends React.HTMLAttributes<HTMLElement> {
+    //Weight prop such as regular, bold
+    weight?: TypographyWeight | undefined;
```
qroll, line 5: comments should be in jsdoc format `/** <description here> */` so that they are visible in vscode

src/typography/typography.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    export const LinkBL = createLinkComponent("body-baseline");
+    export const LinkMD = createLinkComponent("body-md");
+    export const LinkLG = createLinkComponent("body-lg");
+    export const LinkSM = createLinkComponent("body-sm");
+}
+
+// FOR LINK :
+export const StyledExternalIcon = styled(ExternalIcon)`
+    height: 1em;
```
qroll, line 88: can set height to `1lh` instead so that the icon can be aligned vertically (notice that it's slightly misaligned with the surrounding text)

qroll (changes requested):

qroll (approved):

PR #571, opened 2024-09-23
Layout component-section container content

mahidhar-reddy09:
Set up layout component for section,container and content.


src/layout/container.tsx
```diff
@@ -0,0 +1,98 @@
...
+};
+
+export const Container = React.forwardRef(Component);
+
+interface StyleProps {
+    $type: ContainerType;
+    $stretch: boolean;
+}
+
+const StyledContainer = styled.div<StyleProps>`
```
qroll, line 35: could you help to move the styled component to a separate file?

src/layout/types.ts
```diff
@@ -0,0 +1,19 @@
+export interface CommonLayoutProps
+    extends React.HTMLAttributes<HTMLDivElement> {
+    children: React.ReactNode;
+    "data-testid"?: string | undefined;
+    stretch?: boolean;
```
qroll, line 5:
```suggestion
    stretch?: boolean | undefined;
```

tests/layout/layout-container.spec.tsx
qroll, line 1: unit tests are not required for these components as they have minimal behaviour and the logic is mostly for visual styling

qroll (changes requested):

src/layout/container.tsx
```diff
@@ -0,0 +1,98 @@
...
+    }
+    ${MediaQuery.MaxWidth.sm} {
+        width: 100%;
+        padding: 0;
+        max-width: unset;
+    }
+
+    ${(props) => {
+        switch (props.$type) {
+            case "grid":
```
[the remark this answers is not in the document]
mahidhar-reddy09: Hi Roo Ling I am not sure what does comment means. In the sense that what or how is the theme and breakpoint used other than the current mediaquery breakpoint in the context of this grid case?

src/layout/section.tsx
```diff
@@ -0,0 +1,55 @@
...
+            `;
+        } else {
+            return css`
+                padding: 0 1.5rem;
+
+                padding-left: 1.5rem !important;
+                padding-right: 1.5rem !important;
+            `;
+        }
+    }}
```
qroll, line 54: since the margin will be in the container, this doesn't seem to be in use any more

qroll (changes requested):

src/layout/container.styles.tsx
```diff
@@ -0,0 +1,129 @@
...
+        padding: 0 ${Breakpoint["sm-margin"]};
+    }
+
+    ${MediaQuery.MaxWidth.xs} {
+        padding: 0 ${Breakpoint["xs-margin"]};
+    }
+
+    ${MediaQuery.MaxWidth.xxs} {
+        padding: 0 ${Breakpoint["xxs-margin"]};
+    }
```
qroll, line 53: this would take override the styles above. only apply this if `$stretch` is false.

qroll (changes requested):

src/layout/container.styles.tsx
```diff
@@ -0,0 +1,129 @@
...
+    height: auto;
+
+    ${(props) => {
+        if (props.$stretch) {
+            return css`
+                padding: 0 32px;
+            `;
+        } else {
+            return css`
+                padding: 0 32px;
```
qroll, line 25: should be based on xxl-margin token

qroll (approved):

PR #581, opened 2024-09-26
[BOOKINGSG-6090-timetable][JZ|FY] Added TimeTable component and enhanced popover

xyukianesa:
**Changes**
- Added TimeTable component

- [delete] branch

**Changelog entry**
Added TimeTable component to facilitate listing displays and its statuses across a time range, with lazy loading, empty content, custom popovers, and click functionalities.
Updated Popover to allow offset/delay props to be customisable, defaults to the existing fixed values of offset=16, delay={open:0, close: 500}.

**Additional information**

- Ticket at [BOOKINGSG-6090](https://jira.ship.gov.sg/browse/BOOKINGSG-6090)


src/timetable/types.ts
```diff
@@ -0,0 +1,75 @@
...
+    maxDate?: string | undefined;
+    totalRecords?: number | undefined;
+    width?: string | undefined;
+    height?: string | undefined;
+    outsideOpHoursCellCustomPopover?: CustomPopoverProps | undefined;
+    onRefresh?: () => void;
+    onPage?: () => void;
+    onLeftArrowClick?: (currentDate: string) => void;
+    onRightArrowClick?: (currentDate: string) => void;
+    onNameClick?: (rowData: RowData, e: React.MouseEvent) => void;
```
qroll, line 61:
```suggestion
    onRowNameClick?: (rowData: RowData, e: React.MouseEvent) => void;
```

src/timetable/types.ts
```diff
@@ -0,0 +1,75 @@
...
+export const ROW_INTERVAL = 15;
+export type RowBarColors = (typeof ROW_BAR_COLOR_SEQUENCE)[number];
+export type CellType = "filled" | "blocked" | "default";
+
+interface EmptyContentProps {
+    description: string;
+    illustrationScheme: ResourceScheme;
+}
+
+export interface RowData {
```
qroll, line 21: public types need to be named more explicitly to avoid collision

```suggestion
export interface TimeTableRowData {
```

src/timetable/types.ts
```diff
@@ -0,0 +1,75 @@
...
+    rowMinTime?: string;
+    // HH:mm format, defaults to maxTime
+    rowMaxTime?: string;
+}
+
+export interface CustomPopoverProps {
+    trigger: PopoverV2TriggerType;
+    content: string | JSX.Element;
+    width?: string | undefined;
+    padding?: string | undefined;
```
qroll, line 37: what's the use case for custom padding?

src/timetable/types.ts
```diff
@@ -0,0 +1,75 @@
...
+export const ROW_BAR_COLOR_SEQUENCE = [
+    "#FFE6BB",
+    "#D8EFEB",
+    "#E6EAFE",
+    "#FAE4E5",
+    "#D3EEFC",
+] as const; // Assert to be a readonly tuple
+export const ROW_CELL_GAP = 2;
+export const ROW_INTERVAL = 15;
+export type RowBarColors = (typeof ROW_BAR_COLOR_SEQUENCE)[number];
```
qroll, line 13: for internally used consts and types, move them to another file to avoid exposing them publicly

src/timetable/types.ts
```diff
@@ -0,0 +1,75 @@
...
+    // HH:mm format
+    minTime?: string;
+    // HH:mm format
+    maxTime?: string;
+    minDate?: string | undefined;
+    maxDate?: string | undefined;
+    totalRecords?: number | undefined;
+    width?: string | undefined;
+    height?: string | undefined;
+    outsideOpHoursCellCustomPopover?: CustomPopoverProps | undefined;
```
qroll, line 56: this was one of my concerns with automatically computing slots. for any additional rendering behaviour, new props need to be introduced for these special slots

qroll (commented):
some initial comments on the component interface

stories/timetable/timetable-data-even-days.json
qroll, line 1: for the mock data, could it be simplified to match `RowData` without requiring the mapper? this makes it easier for devs to view the shape of data from the examples

src/timetable/types.ts
```diff
@@ -0,0 +1,75 @@
...
+    "#FFE6BB",
+    "#D8EFEB",
+    "#E6EAFE",
+    "#FAE4E5",
+    "#D3EEFC",
+] as const; // Assert to be a readonly tuple
+export const ROW_CELL_GAP = 2;
+export const ROW_INTERVAL = 15;
+export type RowBarColors = (typeof ROW_BAR_COLOR_SEQUENCE)[number];
+export type CellType = "filled" | "blocked" | "default";
```
qroll, line 14: is it possible to include another type for the auto-filled slot?

src/timetable/types.ts
xyukianesa: will create a const.ts file separately for this

stories/timetable/timetable-data-even-days.json
xyukianesa: done for odd/even days mock data files, also added another file for default render, keeping lazy load as-is if thats ok 🙏🏻

src/timetable/types.ts
xyukianesa: this was not explicitly mentioned in the Figma, but for the intended facilities use case, the filled slots popover needs to have a fixed width of 400px, and required some padding also

https://www.figma.com/design/sjlRZcuzHyFefY9uLjxZdJ/%5BBSG2.0%5D-Facilities-%E2%80%94-General-users?node-id=8920-143145&node-type=frame&t=VQFzI7xc2VfLzBlM-0

alternatively, it could be done up in such a way so that this component just renders the content, instead of wrapping it over the PopoverCard component?

qroll: how about directly reusing the PopoverTrigger's `popoverContent`?

if `string | JSX.Element`, it renders PopoverV2 (which comes with mobile handling)

if `(() => React.ReactNode)`, it renders the custom element (which in your use case is a card of fixed width 400px)

src/timetable/timetable-row/row-cell.style.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+export const Gap = styled.div`
+    width: 2px;
+    height: 100%;
+`;
+
+export const Block = styled.div<BlockStyleProps>`
+    height: 100%;
+    width: ${({ $width }) => `${$width}px`};
+    font-size: 11px;
```
qroll, line 44: font size not needed

src/timetable/timetable-navigator/timetable-navigator.tsx
```diff
@@ -0,0 +1,107 @@
...
+    // EVENT HANDLERS
+    // =============================================================================
+
+    const scrollToTop = () => {
+        if (tableContainerRef.current) {
+            tableContainerRef.current.scrollTop = 0;
+        }
+    };
+
+    const onRefresh = () => {
```
qroll, line 39:
```suggestion
    const handleRefresh = () => {
```

src/timetable/timetable-navigator/timetable-navigator.style.tsx
```diff
@@ -0,0 +1,40 @@
+import styled, { css } from "styled-components";
+import { Color } from "../../color";
+import { IconButton } from "../../icon-button";
+import { Text } from "../../text";
+
+export const StyledRefreshButton = styled(IconButton)<{ $isLoading: boolean }>`
+    color: ${Color.Neutral[3]};
+    @keyframes spin {
```
qroll, line 8: to use styled components keyframe to ensure the name is unique https://styled-components.com/docs/basics#animations

qroll (approved):

PR #582, opened 2024-09-29
[BOOKINGSG-6120][JH] additional timeRangePicker combobox fixes

jhyeep:
**Changes**
- fixed `errorMessage` prop to hide dropdown as well (only internal validation was doing this previously)
- modified invalid time range validation behavior to keep values instead of auto clearing end value
- fixed bug where clicking outside the input, on the error message was not triggering the blur event
- split tab behavior test case to (hopefully) fix test case sometimes failing on gitlab pipeline

- [delete] branch

### DIFF
--- a/src/time-range-picker/combobox-picker/combobox-picker.tsx
+++ b/src/time-range-picker/combobox-picker/combobox-picker.tsx
@@ -16,7 +16,6 @@ import { TimeRangePickerProps, TimeRangePickerValue } from "../types";
 type TimeRangeInputType = "start" | "end";
 interface TimeChangeOptions {
     goToNextInput?: boolean | undefined;
-    skipRangeCheck?: boolean | undefined;
     triggerOnBlur?: boolean | undefined;
 }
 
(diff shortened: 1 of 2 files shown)


src/time-range-picker/combobox-picker/combobox-picker.tsx
```diff
@@ -248,9 +241,8 @@ export const ComboboxPicker = ({
         if (
             nodeRef.current &&
             !nodeRef.current.contains(event.relatedTarget) &&
-            !dropdownOpen
+            !dropdownOpen // Necessary because dropdown floating ui is not a child
```
qroll, line 244: `onBlur` fires twice with shift + tab due to overlap with `handleClose`

suggest to:
1. check `activeTimeSelector` is truthy here (similar to the focus check in the `Select` component)
2. trigger `onBlur` in `handleClose`

qroll (changes requested):

qroll (approved):

PR #596, opened 2024-10-16
[MOL-16856][CWH] Update dropdown field align to center in mobile scre…

CWH-UF:
**Changes**
- add in middleware to set the x value to align the dropdown selection to center when in mobile & it no gap between the dropdown selection and the screen
- delete branch

**Additional information**

- You may refer to this [MOL-16856](https://jira.ship.gov.sg/browse/MOL-16856)

### DIFF
--- a/src/shared/dropdown-wrapper/element-with-dropdown.tsx
+++ b/src/shared/dropdown-wrapper/element-with-dropdown.tsx
@@ -1,6 +1,7 @@
 import {
     FloatingFocusManager,
     FloatingPortal,
+    Middleware,
     OpenChangeReason,
     Placement,
     autoUpdate,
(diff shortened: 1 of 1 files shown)


src/shared/dropdown-wrapper/element-with-dropdown.tsx
```diff
@@ -77,6 +79,20 @@ export const ElementWithDropdown = ({
         targetRef: elementRef,
         handleHeight: false,
     });
+    const center: Middleware = {
+        name: "center",
+        fn: ({ x, rects }) => {
+            const noGapInBetween =
```
qroll, line 85: also consider the case where dropdown is right aligned. `x` will be 0

qroll (changes requested):

src/shared/dropdown-wrapper/element-with-dropdown.tsx
CWH-UF: updated, add in x===0

qroll (approved):

PR #601, opened 2024-10-24
[MOL-16900][CWH] Set color to initial to avoid been overwrite by appl…

CWH-UF:
**Changes**
- add color initial to avoid been overwrite by the apple-system-blue in IOS
- delete branch

**Additional information**

- You may refer to this [MOL-16900](https://jira.ship.gov.sg/browse/MOL-16900)


qroll:
hello! this style is applied by default in https://assets.life.gov.sg/react-design-system/v2/css/main.css. is it still not working?

in any case, we should apply the actual component color as dictated by the Figma design, as it's not guaranteed what is being inherited

CWH-UF:
will update to use v2

PR #610, opened 2024-11-06
V3 docs clean up

qroll:
**Changes**

- Reorganise components (will be updating as components get migrated)
- Update stories to render cleaner JSX and move wrappers to decorators
  - Done by adding `_args` to the CSF3 render function
  - However for stories using hooks like `useState` I opted to keep the full source as functions get rendered as `() => {}` which is not useful
- [delete] branch

will be easier to view the changes based on commits 🙏 

biggest changes are for `Toggle` in f1b06297ed176c9b366e35daf7a07d1c2f791a86
- simplifies the display to be row-based on bigger viewports and column-based on smaller viewports
- use a common decorator to render the variants


weili-govtech:
I noticed a few stories pointing to our prod's storybook directly, e.g. https://github.com/LifeSG/react-design-system/blob/v3-docs-clean-up/stories/otp-input/props-table.tsx#L12

now that we are reorganising the stories and serving a v3 storybook, we need to update them as well

qroll:
> I noticed a few stories pointing to our prod's storybook directly, e.g. https://github.com/LifeSG/react-design-system/blob/v3-docs-clean-up/stories/otp-input/props-table.tsx#L12
> 
> now that we are reorganising the stories and serving a v3 storybook, we need to update them as well

thanks for the spot.

previously had issues with getting links to work on both local and ci, taking the chance now to resolve this using a helper component

weili-govtech (approved):

PR #613, opened 2024-11-08
Make changes to Divider and Radio and Col Div

mahidhar-reddy09:
Use theme default for Divider, added displaysize to radio and allowed for full width span for col div.

### DIFF
--- a/src/divider/divider.tsx
+++ b/src/divider/divider.tsx
@@ -1,7 +1,9 @@
+import { useTheme } from "styled-components";
 import { ColDiv } from "../layout/col-div";
 import { V2_ColDiv } from "../v2_layout/col-div";
 import { Line } from "./divider.style";
 import { DividerProps } from "./types";
+import { Breakpoint } from "../theme";
 
 export const Divider = ({
     thickness = 1,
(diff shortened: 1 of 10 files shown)


src/layout/col-div.tsx
qroll, line 1: col-div typings will also need to allow `-1` as the end value

src/radio-button/types.ts
```diff
@@ -1,2 +1,5 @@
+export type RadioboxSize = "small" | "default";
```
qroll, line 1:
```suggestion
export type RadioButtonSize = "small" | "default";
```

src/radio-button/types.ts
qroll, line 1: please update the documentation under `stories/radio-button/props-table.tsx` to include this new prop

do also help to migrate this file to use the `ApiTable` component, similar to `stories/checkbox/props-table.tsx`

src/divider/divider.tsx
```diff
@@ -16,16 +18,27 @@ export const Divider = ({
...
 }: DividerProps) => {
+    const theme = useTheme();
     const isV2Layout =
         mobileCols !== undefined ||
         tabletCols !== undefined ||
         desktopCols !== undefined;
 
+    const colProps = {
+        xxsCols: xxsCols ?? Breakpoint["xxs-column"]({ theme }),
+        xsCols: xsCols ?? Breakpoint["xs-column"]({ theme }),
```
qroll, line 34: now that the ColDiv can accept `-1`, we can use that as the default value instead of having to pull from the theme

qroll (changes requested):

PR #614, opened 2024-11-08
[BOOKINGSG-6442][RT] Fix disabled color for RangeSelect and related components

raymondtanwl:
**Changes**
Fix disabled color for RangeSelect and related components

**Changelog entry**
- Fix disabled color for RangeSelect and related components

**Additional information**
Screenshot 2024-11-08 at 4 10 07 PM

### DIFF
--- a/src/shared/dropdown-wrapper/dropdown-wrapper.styles.tsx
+++ b/src/shared/dropdown-wrapper/dropdown-wrapper.styles.tsx
@@ -25,6 +25,10 @@ export interface SelectorStyleProps {
     $variant?: DropdownVariantType | undefined;
 }
 
+export interface LabelContainerStyleProps {
+    $disabled?: boolean | undefined;
+}
+
 // =============================================================================
 // STYLING
 // =============================================================================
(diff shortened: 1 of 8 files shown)


src/shared/dropdown-wrapper/dropdown-wrapper.styles.tsx
```diff
@@ -189,10 +189,17 @@ export const Divider = styled.div`
     margin: 0 0.5rem;
 `;
 
-export const LabelContainer = styled.div`
+export const LabelContainer = styled.div<{ disabled?: boolean }>`
```
qroll, line 192: declare the prop in an interface under the `STYLE INTERFACE` section instead of inlining them per the repo conventions

qroll, line 192: non-html attributes should be prefixed with `$` to mark them as [transient](https://styled-components.com/docs/api#transient-props)

```suggestion
export const LabelContainer = styled.div<{ $disabled?: boolean }>`
```

qroll (changes requested):

src/shared/dropdown-wrapper/dropdown-wrapper.styles.tsx
raymondtanwl: done and shifted into style prop

qroll (approved):

qroll:
thanks for the fix!

PR #617, opened 2024-11-22
[CCUBE-1628][MAHI]Migrate accordian and text list components

mahidhar-reddy09:
Migrate accordian and text list components

### DIFF
--- a/codemods/migrate-text-list/index.ts
+++ b/codemods/migrate-text-list/index.ts
@@ -1,0 +1,78 @@
+import { API, FileInfo, JSCodeshift } from "jscodeshift";
+
+// ======= Constants ======= //
+
+const IMPORT_PATHS = {
+    TEXT_LIST: "@lifesg/react-design-system/text-list",
+    V2_TEXT_LIST: "@lifesg/react-design-system/v2_text-list",
+    DESIGN_SYSTEM: "@lifesg/react-design-system",
+};
+
+const IMPORT_SPECIFIERS = {
+    TEXT_LIST: "TextList",
+    V2_TEXT_LIST: "V2_TextList",
+};
+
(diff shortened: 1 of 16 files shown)


src/accordion/accordion-item.style.tsx
```diff
@@ -1,11 +1,11 @@
 import { ChevronUpIcon } from "@lifesg/react-icons/chevron-up";
 import { animated } from "react-spring";
 import styled, { css } from "styled-components";
-import { V2_Color } from "../v2_color";
-import { V2_MediaQuery } from "../v2_media";
 import { ClickableIcon } from "../shared/clickable-icon";
-import { V2_Text } from "../v2_text/text";
 import { Transition } from "../transition";
```
qroll, line 5: to switch from `Transition` to use the new motion tokens

src/accordion/accordion-item.style.tsx
qroll, line 1: update `ExpandCollapseButton` to use motion tokens for transition property

src/text-list/types.ts
qroll, line 1: this would be a breaking change. could a `V2_TextList` be added along with the corresponding codemod migration for `deprecate-v2-tokens`?

qroll (changes requested):

codemods/migrate-text-list/index.ts
qroll, line 1: there's a breaking change in the `size` prop, could this new codemod help to map over the typography size accordingly?

qroll, line 1: to also include in the main `codemods/run-codemod.ts` list

qroll (changes requested):

qroll (approved):

PR #619, opened 2024-11-29
Deprecate text list to V2 and update eslint 

mahidhar-reddy09:
Deprecate text list to V2 and update eslint to remove restriction of pascal case for folders with v2


qroll:
could the codemod `deprecate-v2-tokens` also be updated to include the `TextList` type exports in the mapping?

qroll (approved):

PR #621, opened 2024-12-03
Migrate toggle composite section with design changes

mahidhar-reddy09:
Update toggle composite section styling with the latest design and logic changes.

### DIFF
--- a/src/toggle/toggle.styles.tsx
+++ b/src/toggle/toggle.styles.tsx
@@ -102,9 +102,7 @@ export const Container = styled.div<ContainerStyleProps>`
 
                         &:has(${HeaderContainer}:hover) {
                             background: ${Colour["bg-selected-hover"]};
-                        }
 
-                        &:hover {
                             & ${TextContainer} {
                                 color: ${Colour["text-selected-hover"]};
                             }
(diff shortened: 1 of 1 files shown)


src/toggle/toggle.styles.tsx
```diff
@@ -118,7 +118,11 @@ export const Container = styled.div<ContainerStyleProps>`
             default: {
                 if (props.$error) {
                     return css`
-                        border-color: ${Colour["border-error-strong"]};
+                        border-color: ${Colour["border-error"]};
+
+                        &:has(${HeaderContainer}:hover) {
+                            background: ${Colour["bg-hover-subtle"]};
+                        }
```
qroll, line 125: to also apply this hover behaviour to `no-border` case

src/toggle/toggle.styles.tsx
qroll, line 1: `TextContainer` and `StyledToggleIcon` also apply color styling based on parent hover - can move this into `Container` instead and let `Container` override the children styling

qroll (changes requested):

src/toggle/toggle.styles.tsx
```diff
@@ -148,6 +158,16 @@ export const Container = styled.div<ContainerStyleProps>`
                         &:has(${HeaderContainer}:hover) {
                             background: ${Colour["bg-selected-hover"]};
                         }
+
+                        &:hover {
```
qroll, line 162: this should also follow the `:has` selector like the background since the hover trigger behaviour is supposed to be the same. they can be merged together

src/toggle/toggle.styles.tsx
qroll: for `no-border` error case as well

qroll (approved):

PR #625, opened 2024-12-04
[MOL-17154][AM] update handle change behaviour in textarea

AnuMutyala:
**Changes**
[MOL-17154][AM] update handle change behaviour in textarea

- delete branch

**Additional information**

- You may refer to this [ticket](https://sgtechstack.atlassian.net/browse/MOL-17154)

### DIFF
--- a/src/form/form-textarea.tsx
+++ b/src/form/form-textarea.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
 import { TextareaBase } from "../input-textarea/textarea";
 import { TextareaCounter } from "../input-textarea/textarea-counter";
 import { TextareaRef } from "../input-textarea/types";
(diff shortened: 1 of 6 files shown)


src/form/form-textarea.tsx
```diff
@@ -32,17 +32,6 @@ const FormTextareaComponent = (
         ...otherProps
     } = props;
 
-    const [stateValue, setStateValue] = useState<
```
qroll, line 35: realised that the counter no longer works if using this as a uncontrolled component (check the behaviour of the stories with counter). hence the state is still needed to track the current input length

changes required:
1. handleChange calls `setStateValue()` and `onChange()`
2. `TextAreaCounter` should use `stateValue`
3. TextareaBase should use `value`
4. restore `useEffect` to sync `stateValue` with `value`

qroll (changes requested):

src/form/form-textarea.tsx
qroll: there's still the problem of 4 not being synced if value is undergoes transformation

qroll (approved):

PR #631, opened 2024-12-23
Nested select input

qroll:
**Changes**

follow up to https://github.com/LifeSG/react-design-system/pull/529

- rework the `NestedSelect` component to use the new floating ui-based helper components
- use the new `NestedDropdownList`
- align focus/blur behaviour and keyboard interactions with the `InputSelect` component
- tweaks to aria markup to match the intended semantics

Todo:
- optimise performance
- persist the focus state

### DIFF
--- a/src/input-multi-select/input-multi-select.tsx
+++ b/src/input-multi-select/input-multi-select.tsx
@@ -218,6 +218,7 @@ export const InputMultiSelect = <T, V>({
                     disabled={disabled}
                     expanded={showOptions}
                     listboxId={internalId}
+                    popupRole="listbox"
                     readOnly={readOnly}
                     variant={variant}
                 >
--- a/src/input-nested-select/input-nested-select.tsx
+++ b/src/input-nested-select/input-nested-select.tsx
@@ -235,6 +235,7 @@ export const InputNestedSelect = <V1, V2, V3>({
                     disabled={disabled}
                     expanded={showOptions}
                     listboxId={internalId}
+                    popupRole="tree"
                     readOnly={readOnly}
                     variant={variant}
                 >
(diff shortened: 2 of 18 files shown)


src/input-multi-select/input-multi-select.tsx
```diff
@@ -218,6 +218,7 @@ export const InputMultiSelect = <T, V>({
                     disabled={disabled}
                     expanded={showOptions}
                     listboxId={internalId}
+                    popupRole="listbox"
```
nobeltws, line 221: out of curiousity, what does this prop do?

qroll: this translates to the `aria-haspopup` [role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-haspopup) which is used in combobox style of inputs to hint to accessibility tools that there is dropdown content

it should match the `role` of the dropdown content, in the case of nested select input, it should be `tree`, for basic lists, it should be `listbox`

src/input-nested-select/input-nested-select.tsx
```diff
@@ -45,93 +51,114 @@ export const InputNestedSelect = <V1, V2, V3>({
...
+    // =========================================================================
     useEffect(() => {
-        const newKeyPath = _selectedKeyPath ? [_selectedKeyPath] : [];
-
-        setSelectedKeyPaths(newKeyPath);
-        updateSelectedItemFromKey(options, _selectedKeyPath || []);
+        setSelectedKeyPaths(_selectedKeyPath ? [_selectedKeyPath] : []);
+        const selectedItem = findItemByKeyPath(options, _selectedKeyPath || []);
+        if (selectedItem) {
+            const { label, value } = selectedItem;
```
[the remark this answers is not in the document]
qroll: yeah it doesn't hurt to do so since the `SelectedItem` typing is a subset of `selectedItem`, have updated to simplify this

nobeltws (approved):
looks good~

PR #635, opened 2025-01-23
Nested multi select input

qroll:
**Changes**

follow up to https://github.com/LifeSG/react-design-system/pull/529

- rework the `NestedMultiSelect` component to use the new floating ui-based helper components
- use the new `NestedDropdownList`
- align focus/blur behaviour and keyboard interactions with the `InputSelect` component


weili-govtech (changes requested):

weili-govtech:
noticed in the default nested-multi-select story, the prefill example is not getting prefilled

qroll:
> noticed in the default nested-multi-select story, the prefill example is not getting prefilled

thanks! this was due to referencing the keys belonging to the previous options. have updated the stories to match the single nested select

weili-govtech (approved):

PR #644, opened 2025-02-25
[MOL-17821][JAZ] Add new loading spinner

Jazsley:


src/animations/loading-spinner/loading-spinner.tsx
```diff
@@ -17,5 +17,5 @@ export const LoadingSpinner = (props: BaseAnimationProps) => {
 
 const Container = styled.div`
     margin: 0 auto;
-    padding: 1rem 2rem 4rem;
+    padding: 3rem 2rem 2rem;
```
qroll, line 20: the original padding was meant to visually center the spinner vertically. I think this can be reduced with

`padding: 0 1rem 0;`

src/animations/loading-spinner/data.json
qroll, line 1: could the white background be removed from the animation?
Screenshot 2025-02-27 at 9 10 11 AM

qroll (changes requested):

src/animations/loading-spinner/data.json
qroll: noted that this will be here temporarily while UX looks into removing the white background

qroll (approved):

PR #645, opened 2025-02-26
Enable Typescript strict mode

qroll:
**Changes**

- Enable [strict mode](https://www.typescriptlang.org/tsconfig/#strict) and fix type errors that cropped up
- The main motivation is to make sure access to the design tokens have type safety. When `noImplicitAny` is disabled, `Colour["wrong"]` does not get flagged out
- [delete] branch

sorry for the huge amount of code changes, in hindsight the individual strict mode flags could have been enabled incrementally. please view each of the commits individually - I have left some comments to hopefully give more info about certain changes

commits of interest:
b69ea7c61b47c9f1dd478eea887e8951a4f369d2
7df91b4bafe77d0593562972d223eec44ba846bb
ffd7fc3932bf37b037b2b2070a7ab2a95f582886
0d9668ed3f7340cb1ce12144a591bc549732edc3


.vscode/settings.json
```diff
@@ -12,5 +12,6 @@
         "mdx"
     ],
     "editor.rulers": [80],
-    "typescript.preferences.importModuleSpecifier": "relative"
+    "typescript.preferences.importModuleSpecifier": "relative",
+    "typescript.tsdk": "node_modules/typescript/lib"
```
qroll, line 16: this ensures the TS server uses the workspace version of TS installed in node modules instead of VSCode's

this avoid scenarios where VSCode and rollup show different errors

package.json
```diff
@@ -76,6 +76,7 @@
         "@testing-library/jest-dom": "^5.16.3",
         "@testing-library/react": "^12.1.4",
         "@testing-library/user-event": "^14.5.2",
+        "@types/dompurify": "^2.4.0",
```
qroll, line 79: used in NotificationBanner

src/alert/alert.tsx
```diff
@@ -126,7 +128,7 @@ export const Alert = ({
             $hasActionLink={!!actionLink}
         >
             <div ref={contentRef}>{children}</div>
-            {actionLink && renderLink()}
+            {renderLink()}
```
qroll, line 131: Typescript can't recognise that we have already performed the type check

in most cases I resolve this by:
1. passing the checked prop to be an argument
2. moving the check into the render function

src/date-input/date-input.tsx
```diff
@@ -166,7 +168,7 @@ export const DateInput = ({
     // =============================================================================
     // HELPER FUNCTIONS
     // =============================================================================
-    const performOnChangeHandler = (changeValue?: string) => {
+    const performOnChangeHandler = (changeValue: string) => {
```
qroll, line 171: unnecessary `?`

src/file-upload/dropzone.tsx
```diff
@@ -14,7 +14,7 @@ export interface DropzoneElement extends HTMLInputElement {
 }
 
 interface Props extends FileInputProps {
-    children: JSX.Element[];
+    children: React.ReactNode;
```
qroll, line 17: switched because JSX.Element has type error when conditional rendering in this manner:
```
{selected && <div />}
```

src/phone-number-input/phone-number-input-helper.tsx
```diff
@@ -79,7 +81,8 @@ export namespace PhoneNumberInputHelper {
                     };
                 }
 
-                const [head, ...tail] = acc.remainingText;
```
qroll, line 82: replaced as `...tail` is an array and doesn't match the argument type

src/shared/dropdown-list-v2/dropdown-list.tsx
```diff
@@ -491,9 +494,10 @@ export const DropdownList = <T, V>({
             return;
         }
 
+        // FIXME: implement onDismiss handling
```
qroll, line 492: based on the prop description, `renderCustomCallToAction` is supposed to pass down a helper function that allows the consumer to manually close the dropdown

currently it just calls `onDismiss` which doesn't actually do anything???

src/timetable/timetable-row/row-bar.tsx
```diff
@@ -26,11 +26,6 @@ export const RowBar = ({
     outOfRangeCellPopover,
     ...otherProps
 }: RowBarProps) => {
-    // =============================================================================
-    // CONST, STATE, REF
-    // =============================================================================
-    const testId = otherProps["data-testid"] || "timetable-row";
```
qroll, line 32: this data-testid prop doesn't actually exist

src/input-nested-select/input-nested-select.tsx
```diff
@@ -156,10 +156,11 @@ export const InputNestedSelect = <V1, V2, V3>({
     // HELPER FUNCTION
     // =========================================================================
     const getDisplayValue = (): string => {
```
qroll, line 158: generally:
1. custom stringify takes precedence
2. else stringify value
3. fallback to `""` if value is undefined/null

src/input-range-select/input-range-select.tsx
```diff
@@ -151,41 +151,35 @@ export const InputRangeSelect = <T, V>({
...
+    const truncateValue = (type: RangeType, value: string) => {
         if (optionTruncationType === "middle") {
             let widthOfElement = 0;
             if (labelContainerRef[type] && labelContainerRef[type].current) {
                 widthOfElement =
                     labelContainerRef[type].current.getBoundingClientRect()
                         .width;
             }
-            return StringHelper.truncateOneLine(
-                convertValueToString(value),
```
qroll, line 184: originally, the custom stringify would only be called when truncation type is middle

src/input-range-slider/input-range-slider.tsx
```diff
@@ -62,19 +62,21 @@ export const InputRangeSlider = ({
                 // skip unnecessary update when dragging the start thumb across the end thumb
                 return;
             }
-            setSelection(value);
-            onChange?.(value);
+            const newSelection = [...value];
```
qroll, line 65: forcing `readonly number[]` to `number[]`

tests/codemods/migrate-color/transformer.spec.tsx
```diff
@@ -24,7 +24,7 @@ describe("Codemod Transformer for V2_Color to Colour", () => {
 
         // Execute the jscodeshift command for the codemod
         execSync(
-            `jscodeshift --parser=tsx --verbose=2 -t ./codemods/migrate-colour --mapping=lifesg ${outputPath}`
+            `jscodeshift --parser=tsx --verbose=2 -t ./codemods/migrate-colour/index.ts --mapping=lifesg ${outputPath}`
```
qroll, line 27: there's a babel error that stops execution, if not pointing to the `index.ts`

src/timetable/timetable-row/row-bar.tsx
```diff
@@ -125,7 +120,7 @@ export const RowBar = ({
     ]);
 
     return (
-        <RowCellContainer data-testid={testId} {...otherProps}>
+        <RowCellContainer data-testid="timetable-row" {...otherProps}>
```
weili-govtech, line 123: getting this error:
```Warning: Unknown event handler property `onRowNameClick`. It will be ignored.```

I believe it originates from here, we'll need to exclude passing the custom props to the component

**Review by weili-govtech — CHANGES_REQUESTED** [review]

File: src/file-download/file-list-card/file-list-card.tsx [hunk 14]
```diff
@@ -104,7 +104,7 @@ const Component = ({ fileItem, onDownload }: FileListItemProps) => {
         </>
     );
 
-    const renderWithThumbnail = () => (
+    const renderWithThumbnail = (thumbnailImageDataUrl: string) => (
```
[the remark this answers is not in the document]
qroll: `Thumbnail`  src is expecting `thumbnailImageDataUrl` to be defined. this was checked in renderContents before calling renderWithThumbnail, but in renderWithThumbnail if it still references the prop directly, the undefined typing is still there

i thought of the following options:
1. non null assertion
2. check for `!thumbnailImageDataUrl` again in renderWithThumbnail
3. pass the narrowed type as an argument

was considering between 2 and 3 but ultimately did not choose 2 as it involved another runtime check

src/masked-input/masked-input.tsx
```diff
@@ -49,14 +49,14 @@ const Component = (
     // CONST, STATE, REFS
     // =============================================================================
     const isEmptyReadOnlyState = readOnly && isEmpty(value);
-    const [isMasked, setIsMasked] = useState(!disableMask);
-    const [updatedValue, setUpdatedValue] = useState(value || "");
+    const [isMasked, setIsMasked] = useState<boolean>(!disableMask);
+    const [updatedValue, setUpdatedValue] = useState<string>(value || "");
```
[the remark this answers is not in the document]
qroll: you're right! it's not needed for these two cases. the fallback to `|| ""` properly cast the type for `value`, so Typescript can infer this nicely

but other cases where generics are needed:
1. in the case of `useRef`, we need to be explicit about the element
2. for some values, they need `| undefined` if it's derived from a prop, can be initialised as undefined, or at some point can be set to undefined
3. using a string literal union where typescript would default to `string`

will need to decide on implicit vs explicit generics and make it consistent across the codebase. I am planning to document this in the conventions doc, and do a one-time clean up

src/timetable/timetable-row/row-bar.tsx
qroll: noted, have removed `otherProps` entirely as it was also unintentionally passing in things like `name` and `subtitle`

src/file-download/file-list-card/file-list-card.tsx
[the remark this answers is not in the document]
weili-govtech: yeah that's a fair assessment however won't it be clearer if we rename the argument to avoid shadowing the real `thumbnailImageDataUrl` declared outside the function?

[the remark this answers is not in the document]
qroll: thanks for pointing out, have fixed this particular instance. I think there are other components with this issue, will enable the `no-shadow` lint rule and tackle them separately

weili-govtech (approved):

weili-govtech:
notice the `build-dist-job` in the pipelines for this branch has been failing often, presumably the job has timed out. I found a bunch of circular dependency warnings with the theme in the logs, perhaps that contributed to it?

qroll:
> notice the `build-dist-job` in the pipelines for this branch has been failing often, presumably the job has timed out. I found a bunch of circular dependency warnings with the theme in the logs, perhaps that contributed to it?

thanks for pointing out! I have a fix for the circular dependency planned, will monitor the build

PR #650, opened 2025-03-06
Fix showActiveMonthDaysOnly when minDate is passed in

benjaminLeongSK:
**Changes**
Added additional styling when showActiveMonthsDaysOnly is present

- [delete] branch


qroll (dismissed):

qroll:
noticed that the date inputs which extend `CommonCalendarProps` accept `showActiveMonthDaysOnly` but doesn't do anything with it

will address this separately in the future, either by
1. moving out the prop from `CommonCalendarProps`, or
2. implementing the prop in the date inputs: updating the cells or `DayCell` itself to handle the flag

qroll (approved):

PR #652, opened 2025-03-07
[BOOKINGSG-6625][JZ]

xyukianesa:
**Changes**
- Add `react-virtuoso` to dependencies
- Implement virtualisation to both dropdown-list and nested-dropdown-list

- [keep] branch

**Changelog entry**
- Added Virtualisation example into Select, MultiSelect, NestedSelect, NestedMultiSelect stories.

**Additional information**
- You may refer to this [[ticket](url)](https://sgtechstack.atlassian.net/browse/BOOKINGSG-6624)
- NOTE: Further optimisation required for large lists with these dropdown components.


src/shared/dropdown-list-v2/nested-dropdown-list.tsx
```diff
@@ -190,21 +206,21 @@ export const NestedDropdownList = <T,>({
...
 
-    const handleListItemClick = (index: number) => {
-        setFocusedIndex(index);
-
-        const activeList = searchActive
-            ? filteredListItems
-            : unfilteredListItems;
-        onSelectItem?.(activeList[index]);
+    const handleListItemClick = (itemIndex: number, vIndex: number) => {
+        setVirtuosoIndex(vIndex);
```
qroll, line 210: is it intended for `setFocusedIndex` to be left out here for the click and hover?

src/shared/dropdown-list-v2/nested-dropdown-list.tsx
```diff
@@ -298,10 +314,6 @@ export const NestedDropdownList = <T,>({
     });
 
     const toggleCategory = (index: number, nextExpanded: boolean) => {
-        const activeList = searchActive
-            ? filteredListItems
-            : unfilteredListItems;
-
         const list = toggleSubtree(activeList, index, nextExpanded);
         setFocusedIndex(index);
```
qroll, line 318: should run `setVirtuosoIndex` here too - I think this affects mobile behaviour

src/shared/dropdown-list-v2/nested-dropdown-list.tsx
```diff
@@ -353,16 +370,24 @@ export const NestedDropdownList = <T,>({
...
+            setVirtuosoIndex(0);
             setTimeout(() => listItemRefs.current[0]?.focus(), 200);
         }
-    }, [focusedIndex, mounted, setFocusedIndex]);
+    }, [
+        focusedIndex,
+        virtuosoIndex,
+        mounted,
+        setFocusedIndex,
+        setVirtuosoIndex,
```
qroll, line 389: actually i think both of these setters can be removed as they come from useState and have a stable reference

src/shared/dropdown-list-v2/types.ts
qroll, line 1: my local is formatting the files differently - could you run prettier again on the files here? in particular, the whitespace on the left in this file should be correct one

src/shared/dropdown-list-v2/nested-dropdown-list-helpers.ts
```diff
@@ -230,6 +222,7 @@ export const updateSelectedState = <T>(
             }
         }
     });
+    return res;
 };
 
 export const findIndexFromStart = <T>(
```
qroll, line 228: `findIndexFromStart` and `findIndexFromEnd`  can be removed if they're no longer in use

src/shared/dropdown-list-v2/nested-dropdown-list-helpers.ts
```diff
@@ -198,16 +192,14 @@ export const toggleSubtree = <T>(
...
 ) => {
-    return produce(list, (draft) => {
+    const res = produce(list, (draft) => {
         for (let i = draft.length - 1; i >= 0; i--) {
             const item = draft[i];
-            item.checked = !!selectedKeyPaths.find((keyPath) =>
-                isEqual(keyPath, item.keyPath)
-            );
-
+            const itemKeyPathString = item.keyPath.join(',');
```
qroll, line 201: can have a common helper `buildKeyPath` that takes in the keyPath array and performs the join, mainly to keep the separator consistent

src/input-nested-select/input-nested-select.tsx
```diff
@@ -129,6 +128,7 @@ export const InputNestedSelect = <V1, V2, V3>({
     };
 
     const handleOpen = () => {
+        // console.log(">>>hmm", listOptions);
```
qroll, line 131: unintended console log

src/input-nested-select/input-nested-select.tsx
```diff
@@ -68,8 +68,8 @@ export const InputNestedSelect = <V1, V2, V3>({
     // CONST, STATE
     // =========================================================================
     const options = _options as NestedDropdownListItemProps<V1 | V2 | V3>[];
-    const [selectedKeyPaths, setSelectedKeyPaths] = useState<string[][]>(
-        _selectedKeyPath ? [_selectedKeyPath] : []
+    const [selectedKeyPaths, setSelectedKeyPaths] = useState<Set<string>>(
+        _selectedKeyPath ? new Set<string>([_selectedKeyPath.join(",")]) : new Set<string>()
```
qroll, line 72: also suggest to have a common helper function to convert the key path array to Set

qroll (changes requested):
comments for the nested selects inputs, still reviewing the normal dropdown list

tests/shared/dropdown-list-v2/nested-dropdown-list.spec.tsx
```diff
@@ -30,16 +36,16 @@ describe("NestedDropdownList", () => {
...
+                <NestedDropdownList listItems={options} selectedKeyPaths={new Set<string>()} />
             );
 
             expect(screen.queryByText("Parent 1 item")).toBeVisible();
             expect(screen.queryByText("Child 1.1 item")).toBeVisible();
             expect(screen.queryByText("Child 1.2 item")).toBeVisible();
             expect(screen.queryByText("Parent 2 item")).toBeVisible();
-            expect(screen.queryByText("Child 2.1 item")).not.toBeVisible();
-            expect(screen.queryByText("Child 2.2 item")).not.toBeVisible();
+            expect(screen.queryByText("Child 2.1 item")).toBeNull();
```
qroll, line 47: can use `not.toBeInTheDocument()` instead

tests/test-mocks/virtuoso/virtuoso-test-mocks.tsx
qroll, line 1: can rename this to `__test-utils__/virtuoso/virtuoso-test-mocks.tsx`. this makes it easier to distinguish the component tests from the test-specific utils

src/shared/dropdown-list-v2/dropdown-list.tsx
```diff
@@ -352,37 +346,36 @@ export const DropdownList = <T, V>({
         );
     };
 
-    const renderItems = () => {
+    const renderItems = (item: T, index: number) => {
```
qroll, line 349:
```suggestion
    const renderItem = (item: T, index: number) => {
```

src/shared/dropdown-list-v2/dropdown-list.tsx
```diff
@@ -486,6 +479,24 @@ export const DropdownList = <T, V>({
...
+    const renderVirtualisedList = () => {
+        return (
+            <Virtuoso
+                ref={virtuosoRef}
+                style={{ height: "100%" }}
+                data={displayListItems}
+                customScrollParent={nodeRef.current}
+                itemContent={(index, item) => {
+                    return (
+                        <Listbox role="listbox" id={listboxId}>
```
qroll, line 491: listbox should be rendered as a parent, not per item

src/shared/dropdown-list-v2/dropdown-list.tsx
```diff
@@ -242,25 +238,23 @@ export const DropdownList = <T, V>({
 
         if (disableItemFocus) return;
 
-        const selectedIndex = listItems.findIndex((item) =>
+        const index = listItems.findIndex((item) =>
             checkListItemSelected(item)
         );
-
-        // Focus search input if there is one
```
qroll, line 249: can restore these comment?

src/shared/dropdown-list-v2/dropdown-list.tsx
```diff
@@ -352,37 +346,36 @@ export const DropdownList = <T, V>({
...
+            return (
+                <ListItem
+                    aria-selected={selected}
+                    aria-multiselectable={multiSelect}
+                    data-testid="list-item"
+                    key={getItemKey(item, index)}
+                    onClick={() => handleListItemClick(item, index)}
+                    onMouseEnter={() => handleListItemHover(index)}
+                    ref={(element) => {
+                        if (!element) return;
```
qroll, line 362: why skip the assignment here?

src/shared/dropdown-list-v2/dropdown-list.tsx
```diff
@@ -242,25 +238,23 @@ export const DropdownList = <T, V>({
...
-        } else if (listItemRefs.current[focusedIndex]) {
-            // Else focus on the specified element
+            setTimeout(() => searchInputRef.current?.focus(), 200); // Wait for animation
+        } else if (focusedIndex > 0) {
             setTimeout(() => listItemRefs.current[focusedIndex]?.focus(), 200);
-        } else if (listItemRefs.current[selectedIndex]) {
-            // Else focus on the selected element
-            setFocusedIndex(selectedIndex);
-            setTimeout(() => listItemRefs.current[selectedIndex]?.focus(), 200);
+            virtuosoRef.current.scrollToIndex({ index: focusedIndex, align: "center" });
```
qroll, line 249: should the `scrollToIndex` be invoked together with the `focus` in the setTimeout?

and checking if the order matters - does the element need to be scrolled into visibility before it can be focused?

src/shared/dropdown-list-v2/dropdown-list.tsx
```diff
@@ -215,20 +215,16 @@ export const DropdownList = <T, V>({
     useEventListener("keydown", handleKeyboardPress);
 
     useEffect(() => {
-        if (topScrollItem === undefined) return;
-
+        if (!topScrollItem && virtuosoRef.current) {
+            virtuosoRef.current.scrollTo({ top: 0 });
+        }
```
qroll, line 220: could you explain why the scroll is needed here?

qroll (changes requested):

src/shared/dropdown-list-v2/dropdown-list.tsx
xyukianesa: this is to achieve the same behaviour for the TimeRangePicker component with the combobox variant keyboard tabbing interaction

src/shared/dropdown-list-v2/dropdown-list.tsx
xyukianesa: hmm im not sure if its a behaviour from Virtuoso or something but i noticed the element can sometimes be null hence i added a check to skip them if its null

src/shared/dropdown-list-v2/dropdown-list.tsx
xyukianesa: ill change the order of execution so that its more logical, but from my brief experimenting, the order doesnt really matter and the focusing only works thanks to the timeout

src/shared/dropdown-list-v2/dropdown-list.tsx
qroll: this probably happens because the component gets unmounted when it's offscreen, hence the ref becomes null

by right the optional chaining with `listItemRefs.current[upcomingIndex]?.focus();` should handle it correctly though? or is this causing any issues?

src/shared/dropdown-list-v2/dropdown-list.tsx
qroll: does this need to `return` early as with the previous code? otherwise the remaining code gets executed

qroll:
sorry, you'll need to run prettier again 💀 I'm not sure why the vscode config doesn't kick in by default

src/shared/dropdown-list-v2/dropdown-list.tsx
xyukianesa: dont think its causing any issues, just thought to not store null items into the ref when i was working on it haha

i will revert this part back!

qroll (approved):

qroll:
thanks for making this enhancement! 🙏

PR #656, opened 2025-03-14
[MISC][SW] improve modal typing

ryhuz:
**Changes**
Use discriminated union for `Modal` `enableOverlayClick`

- [delete] branch

### DIFF
--- a/src/modal/types.ts
+++ b/src/modal/types.ts
@@ -2,19 +2,27 @@ import React from "react";
 
 export type ModalAnimationDirection = "top" | "bottom" | "left" | "right";
 
-export interface ModalProps extends React.HTMLAttributes<HTMLDivElement> {
+type OverlayClickProps =
+    | {
+          enableOverlayClick: false;
+          onOverlayClick?: (() => void) | undefined;
+      }
+    | {
+          enableOverlayClick?: true | undefined;
+          onOverlayClick: () => void;
+      };
+
+export type ModalProps = React.HTMLAttributes<HTMLDivElement> & {
     show: boolean;
     children: JSX.Element | JSX.Element[];
     /** Animation direction of appearance and dismissal. Values: "top" | "bottom" | "left" | "right" */
     animationFrom?: ModalAnimationDirection | undefined;
-    enableOverlayClick?: boolean | undefined;
     /** The identifier of the element to inject the Modal into */
     rootComponentId?: string | undefined;
     zIndex?: number | undefined;
-    onOverlayClick?: (() => void) | undefined;
     /** Dismiss keyboard to keep modal in fullscreen */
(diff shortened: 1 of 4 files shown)


src/overlay/types.ts
```diff
@@ -1,4 +1,14 @@
-export interface OverlayProps {
+export type OverlayClickProps =
+    | {
+          enableOverlayClick?: false | undefined;
+          onOverlayClick?: (() => void) | undefined;
```
qroll, line 4: shouldn't this be
```suggestion
          onOverlayClick?: undefined;
```
to prevent the callback from being set without the flag?

src/modal/types.ts
```diff
@@ -1,20 +1,19 @@
...
     animationFrom?: ModalAnimationDirection | undefined;
-    enableOverlayClick?: boolean | undefined;
     /** The identifier of the element to inject the Modal into */
     rootComponentId?: string | undefined;
     zIndex?: number | undefined;
-    onOverlayClick?: (() => void) | undefined;
     /** Dismiss keyboard to keep modal in fullscreen */
     dismissKeyboardOnShow?: boolean | undefined;
-}
+} & OverlayClickProps;
```
qroll, line 16: unfortunately ModalProps can't inherit `OverlayClickProps` directly because `enableOverlayClick` is true by default in Modal and false by default in Overlay

qroll (changes requested):

src/overlay/types.ts
ryhuz: if there ever is a use case of conditional enabling of `overlayClick`, typescript will complain when you pass in a function, since it expects undefined for the `false` case

src/modal/types.ts
ryhuz: but it's now only inheriting the typing which should be the same..?

src/overlay/types.ts
qroll: okay so this is just to ensure `onOverlayClick` is definitely set if `enableOverlayClick` is true?

src/modal/types.ts
qroll: I was thinking Modal would be like this instead since it defaults to true, i.e. `onOverlayClick` is needed by default. haven't actually tried this though

```
    | {
          enableOverlayClick: false;
          onOverlayClick?: (() => void) | undefined;
      }
    | {
          enableOverlayClick?: true  | undefined;
          onOverlayClick: () => void;
      };
```

src/overlay/types.ts
ryhuz: yup

src/modal/types.ts
ryhuz: ah ok i get it. you're right. let me update

PR #660, opened 2025-03-20
[BOOKINGSG-6938][GC] V3 Update ligatures for Plus Jakarta Sans

gohguancheng:
**Changes**
Add font-variant property for Plus Jakarta Sans

- delete branch

**Changelog entry**

- Add font-variant property for Text common to handle Plus Jakarta Sans ligatures

**Additional information**

- You may refer to this ticket: [BOOKINGSG-6938](https://sgtechstack.atlassian.net/browse/BOOKINGSG-6938)


stories/v2_text/props-table.tsx
```diff
@@ -54,6 +54,7 @@ const TEXT_DATA: ApiTableSectionProps[] = [
                     "Specifies the number of lines visible, the additional lines will be truncated",
                 propTypes: ["number"],
             },
+            // TODO: Ask Ruo Ling whether necessary to expose fontVariant
```
qroll, line 57: can remove this liner. I think you'll need to document the setting in `stories/getting-started/themes-props-table.tsx` instead

qroll (changes requested):
thanks for porting the changes to V3! I've taken the liberty to document the recommended usage in css

stories/v2_text/props-table.tsx
qroll: the new path in v2 is `stories/v2/themes-props-table.tsx`

gohguancheng: have added the necessary props to stories/v2/themes-props-table.tsx, as ligatures are hardcoded based on `themes` in v2.

Will not put document here since we won't be exposing this to users of the `Text` component.

qroll (approved):

PR #664, opened 2025-03-26
Disable virtualisation in test environment

qroll:
**Changes**

- Virtuoso doesn't work in unit tests as it depends on actual DOM measurements
- you can normally [mock the context](https://virtuoso.dev/mocking-in-tests/) but this only works if you have a direct dependency on the library. but due to the way the DS is packaged, the affected components bundle their own instance of the virtuoso mock context, so this method doesn't work for DS consumers
- alternatives
  1. move virtuoso to peer dependency - exposing an internal implementation to consumers which is not ideal
  2. expose a mock for consumers - they'll need to update their unit test code to use this mock
  3. disable virtualisation in unit tests: this can be done by setting `initialItemCount`

have gone with 3 to simplify usage for consumers

- [delete] branch


weili-govtech (approved):
noticed a test failed in the pipelines but I am able to run it locally

PR #670, opened 2025-04-10
[MOL-18376][AM] Update the masthead to cater for correct text link color

AnuMutyala:
**Changes**
the color of the text link is shown in 6B4FEB

- [delete] branch

**Changelog entry**
update masthead component

**Additional information**

- You may refer to this [MOL-18376](https://sgtechstack.atlassian.net/browse/MOL-18376)


qroll:
thanks for looking into the theming options for the V2 and V3 masthead

latest update is that SGDS have reverted the link colour to blue in V3 so this change won't be needed

there's a slight mismatch in the background colour but will tackle that separately

PR #678, opened 2025-04-29
[BOOKINGSG-7232-v3][JZ] Update color and allow 24 hour rows

xyukianesa:
Changes

Interpret 23:59 as 24:00 or 00:00 of the next day for time comparisons, and also interpret as ending on the 0th minute

Amend color of the left border for each row to use N5 color

Auto block the entire row from start to end when row cell provided is an empty array and minTime/maxTime are not defined for that row

[delete] branch

Changelog entry

Update row left border color to N5
Interpret 23:59 to be start of next day timing
Auto block entire row if no rowcell provided and no minTime/maxTime provided for that row
Additional information

You may refer to this [BOOKINGSG-7232](https://sgtechstack.atlassian.net/browse/BOOKINGSG-7232)

### DIFF
--- a/src/timetable/timetable-row/row-bar.tsx
+++ b/src/timetable/timetable-row/row-bar.tsx
@@ -96,6 +96,7 @@ export const RowBar = ({
 
         // Handle non-op after hours
         if (
+            rowMaxTime !== "23:59" &&
             dayjs(timetableMaxTime, "HH:mm").isAfter(dayjs(rowMaxTime, "HH:mm"))
         ) {
             rowCellArray.push({
@@ -107,6 +108,16 @@ export const RowBar = ({
             });
         }
 
+        // Handle empty row cells
+        if (sortedRowCells.length === 0) {
+            rowCellArray.push({
+                id,
+                startTime: timetableMinTime,
+                endTime: timetableMaxTime,
+                status: "blocked",
+                customPopover: outOfRangeCellPopover,
+            });
+        }
         return rowCellArray;
     }, [
         id,
--- a/src/timetable/timetable-row/row-bar.tsx
+++ b/src/timetable/timetable-row/row-bar.tsx
@@ -17,8 +17,8 @@ export const RowBar = ({
     id,
     timetableMinTime,
     timetableMaxTime,
-    rowMinTime = timetableMinTime,
-    rowMaxTime = timetableMaxTime,
+    rowMinTime,
+    rowMaxTime,
     rowCells,
     rowBarColor,
     intervalWidth,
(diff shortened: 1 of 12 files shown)


src/timetable/timetable-row/row-bar.tsx
```diff
@@ -107,6 +108,16 @@ export const RowBar = ({
             });
         }
 
+        // Handle empty row cells
+        if (sortedRowCells.length === 0) {
```
qroll, line 112: should this check if rowMinTime/rowMaxTime are set too? as those add no-op cells

src/timetable/timetable.tsx
```diff
@@ -57,7 +57,10 @@ export const TimeTable = ({
     // =============================================================================
     const testId = otherProps["data-testid"] || "timetable";
     const timetableMinTime = TimeHelper.roundToNearestHour(minTime);
-    const timetableMaxTime = TimeHelper.roundToNearestHour(maxTime, true);
+    const timetableMaxTime =
+        maxTime === "23:59"
```
qroll, line 61: why couldn't we handle 24:00 as exactly 24:00 instead? (since I assume the problem here is that the TimeHelper converts 24:00 to 00:00)

stories/timetable/timetable-default-data.tsx
```diff
@@ -227,4 +227,8 @@ export const timetableDefaultData: TimeTableRowData[] = [
             alert(`Clicked on row header for ${rowData.name}`);
         },
     },
+    {
+        name: "sit",
```
qroll, line 231:
```suggestion
        name: "No cells",
```

qroll (commented):
initial thoughts on the 24 hr handling 🤔

src/timetable/timetable-row/row-bar.tsx
xyukianesa: thanks! ill check if they are not set instead, so that it will just generate 1 whole block instead, bcos from facilities consuming this component, during the mapping from data to component props, if theres no row cells, it usually means theres insufficient information to derive the min/max time for that row 🙇🏻

src/timetable/timetable.tsx
xyukianesa: do u mean to get consumer side to pass in 24:00 instead? 

thats because dayjs can differentiate 24:00 and 00:00 without specifically providing a date, it parses 24:00 as 00:00 of the next day so it can tell which is later than the other

however on facilities side we are using js-joda which cannot parse 24:00 and there will be extra checks needed to handle 24:00 (e.g. check for specific string 23:59, convert to 00:00 of the next day for time comparison..) 

so we thought chances of a consumer wanting 23:59 as-is is quite low, and we could reserve it to signify the beginning. of the next day (00:00 of the next day)

cmiiw!

qroll (changes requested):

qroll (approved):

PR #691, opened 2025-05-22
Multi Select - add in no result found desc & selectable limit

CWH-UF:
**Changes**
Description of changes...

- delete branch

**Changelog entry**
**noResultsDesc**
- add in new optional props `noResultsDesc` to -> input-multi-select, input-nested-multi-select, input-nested-select, input-select
- to show up another description under the no result found under the select

**multiSelectLimit**
- add in new optional props `multiSelectLimit` to  input-multi-select, dropdown-list
- to limit the maximum selectable option under multi select
- value should set to > 1 but less than the options length to be consider as a valid limit
- if it been limit
         - Select All and Clear All will be replace by Clear
         - other available option will be disabled if selected length === limit

**Additional information**

- You may refer to this [COMMPORTAL-436](https://sgtechstack.atlassian.net/browse/COMMPORTAL-436) figma

### DIFF
--- a/src/shared/dropdown-list-v2/dropdown-list.styles.tsx
+++ b/src/shared/dropdown-list-v2/dropdown-list.styles.tsx
@@ -108,13 +108,13 @@ export const ListItem = styled.li<ListItemStyleProps>`
     outline: none;
 
     ${(props) => {
-        if (props.$active && props.$selected) {
-            return css`
-                background: ${Colour["bg-hover"]};
-            `;
-        } else if (props.$disabled) {
+        if (props.$disabled) {
             return css`
                 cursor: not-allowed;
+            `;
+        } else if (props.$active && props.$selected) {
+            return css`
+                background: ${Colour["bg-hover"]};
             `;
         } else if (props.$active) {
             return css`
--- a/src/shared/dropdown-list-v2/dropdown-list.tsx
+++ b/src/shared/dropdown-list-v2/dropdown-list.tsx
@@ -376,7 +376,7 @@ export const DropdownList = <T, V>({
                 <ListItem
                     aria-selected={selected}
                     aria-multiselectable={multiSelect}
-                    aria-disabled={hasSelectedMax}
+                    aria-disabled={!selected && hasSelectedMax}
                     data-testid="list-item"
                     key={getItemKey(item, index)}
                     onClick={() => handleListItemClick(item, index)}
@@ -388,7 +388,7 @@ export const DropdownList = <T, V>({
                     tabIndex={active ? 0 : -1}
                     $active={active}
                     $selected={selected}
-                    $disabled={hasSelectedMax}
+                    $disabled={!selected && hasSelectedMax}
                 >
                     {renderListItem ? (
                         renderListItem(item, { selected })
@@ -435,11 +435,9 @@ export const DropdownList = <T, V>({
                         type="button"
                         $variant={variant}
                     >
-                        {maxSelectable
-                            ? "Clear"
-                            : selectedItems.length === 0
-                            ? "Select all"
-                            : "Clear all"}
+                        {maxSelectable || selectedItems.length !== 0
+                            ? "Clear all"
+                            : "Select all"}
                     </SelectAllButton>
                 </SelectAllContainer>
             );


src/input-multi-select/types.ts
```diff
@@ -22,6 +22,7 @@ export interface InputMultiSelectProps<T, V>
     variant?: DropdownVariantType | undefined;
     alignment?: DropdownAlignmentType | undefined;
     dropdownZIndex?: number | undefined;
+    multiSelectLimit?: number | undefined;
```
ryhuz, line 25: name suggestion:  `maxSelected`

src/input-multi-select/input-multi-select.tsx
```diff
@@ -59,6 +61,14 @@ export const InputMultiSelect = <T, V>({
     const nodeRef = useRef<HTMLDivElement>(null);
     const selectorRef = useRef<HTMLButtonElement>(null);
 
+    const isMultiSelectLimit = useMemo(
```
ryhuz, line 64: name suggestion: `hasSelectedMax`

src/input-multi-select/input-multi-select.tsx
```diff
@@ -70,7 +80,7 @@ export const InputMultiSelect = <T, V>({
     // EVENT HANDLERS
     // =============================================================================
     const handleSelectAllClick = () => {
-        if (selected && selected.length > 0) {
+        if ((selected && selected.length > 0) || isMultiSelectLimit) {
```
ryhuz, line 83: what scenario do we need to check that user has selected the max? because if they have selected the max, then `selected.length` should also be >0

src/shared/dropdown-list-v2/dropdown-list.styles.tsx
```diff
@@ -144,6 +145,13 @@ export const CheckboxUnselectedIndicator = styled(SquareIcon)`
     color: ${Colour["icon-primary-subtlest"]};
 `;
 
+export const CheckboxDisabledIndicator = styled(SquareFillIcon)`
+    flex-shrink: 0;
+    height: 1lh;
+    width: 1lh;
+    color: ${Colour["icon-disabled"]};
```
qroll, line 152: we can go with `icon-disabled-subtle` to match the Checkbox component - it's a lighter shade of gray

src/shared/dropdown-list-v2/dropdown-list.tsx
qroll, line 1: include `aria-disabled={hasSelectedMax}` on `ListItem` for accessibility

src/shared/dropdown-list-v2/dropdown-list.styles.tsx
```diff
@@ -188,3 +196,12 @@ export const Spinner = styled(ComponentLoadingSpinner)`
...
     color: ${Colour["icon"]};
 `;
+
+export const NoResultDescContainer = styled.div<ContainerStyleProps>`
+    padding: 0 ${Spacing["spacing-16"]} ${Spacing["spacing-12"]}
+        ${Spacing["spacing-16"]};
+    ${(props) =>
+        props.$variant === "small"
+            ? Font["body-sm-regular"]
+            : Font["body-md-regular"]}
```
qroll, line 206: from DS Figma it will apparently inherit the existing font size, so this custom font style is not needed

src/shared/dropdown-list-v2/dropdown-list.tsx
```diff
@@ -84,6 +89,15 @@ export const DropdownList = <T, V>({
     const searchInputRef = useRef<HTMLInputElement>(null);
     const virtuosoRef = useRef<VirtuosoHandle>(null);
 
+    const hasSelectedMax = useMemo(
+        () =>
+            maxSelected &&
+            listItems &&
+            maxSelected < listItems.length &&
+            maxSelected > 1,
```
qroll, line 97: hmm doesn't `maxSelected > 1` just mean at least 2 items are selected?

if tracking whether user has selected the upper limit, I thought the comparison would be something like `selectedItems.length === maxSelected`

src/shared/dropdown-list-v2/dropdown-list.tsx
```diff
@@ -178,6 +192,16 @@ export const DropdownList = <T, V>({
     };
 
     const handleListItemClick = (item: T, upcomingIndex: number) => {
+        if (
+            hasSelectedMax &&
+            selectedItems &&
+            selectedItems?.length === maxSelected &&
+            !selectedItems.some((selectedItem) => {
+                return selectedItem === item;
+            })
```
qroll, line 201: could reuse `checkListItemSelected` here

src/shared/dropdown-list-v2/types.ts
```diff
@@ -40,6 +40,7 @@ export interface DropdownSearchProps<T> {
     enableSearch?: boolean | undefined;
     /** If specified, the default no results display will not be rendered */
     hideNoResultsDisplay?: boolean | undefined;
+    noResultsDesc?: string | undefined;
```
qroll, line 43: this can also support JSX elements for custom descriptions, like bold text etc.

```suggestion
    noResultsDesc?: React.ReactNode | undefined;
```

src/input-multi-select/types.ts
qroll: suggestion: `maxSelectable`

src/shared/dropdown-list-v2/dropdown-list.tsx
```diff
@@ -84,6 +89,15 @@ export const DropdownList = <T, V>({
     const searchInputRef = useRef<HTMLInputElement>(null);
     const virtuosoRef = useRef<VirtuosoHandle>(null);
 
+    const hasSelectedMax = useMemo(
```
qroll, line 92: just curious why the need for useMemo as it looks like a pretty straightforward computation

src/shared/dropdown-list-v2/dropdown-list.styles.tsx
```diff
@@ -188,3 +196,12 @@ export const Spinner = styled(ComponentLoadingSpinner)`
     margin-right: ${Spacing["spacing-8"]};
     color: ${Colour["icon"]};
 `;
+
+export const NoResultDescContainer = styled.div<ContainerStyleProps>`
```
qroll, line 200: suggest to include `applyHtmlContentStyle` so that the description can automatically have HTML styling (see [Markup](https://designsystem.life.gov.sg/react/index.html?path=/docs/core-markup--docs))

src/theme/colour-semantic/specs/pa-semantic-tokens.ts
```diff
@@ -36,7 +36,7 @@ export const PAColourSet: SemanticColourSet = {
     "icon-selected": getPrimitiveColour("brand-20"),
     "icon-selected-hover": getPrimitiveColour("brand-10"),
 
-    "icon-disabled": getPrimitiveColour("neutral-50"),
+    "icon-disabled": getPrimitiveColour("neutral-70"),
```
qroll, line 39: by right this should be darker than `icon-disabled-subtle`

src/shared/dropdown-list-v2/types.ts
qroll, line 43:
```suggestion
    noResultsDescription?: string | undefined;
```

qroll (changes requested):
some comments are applicable to the nested dropdown list as well

src/shared/dropdown-list-v2/dropdown-list.tsx
```diff
@@ -359,6 +375,7 @@ export const DropdownList = <T, V>({
                 <ListItem
                     aria-selected={selected}
                     aria-multiselectable={multiSelect}
+                    aria-disabled={hasSelectedMax as boolean}
```
qroll, line 378: instead of doing type casting here, you can do `hasSelectedMax = !!maxSelectable && !!selectedItems ...` to cast the variables to boolean

qroll (changes requested):

src/shared/dropdown-list-v2/dropdown-list.tsx
```diff
@@ -359,6 +376,7 @@ export const DropdownList = <T, V>({
                 <ListItem
                     aria-selected={selected}
                     aria-multiselectable={multiSelect}
+                    aria-disabled={hasSelectedMax}
```
qroll, line 379: on further thought, for this and line 391, the item should only be disabled if it is `hasSelectedMax` and is NOT selected

because by right the selected items are still interactable

qroll (changes requested):

qroll (approved):

PR #692, opened 2025-05-28
[MISC][LS] Accessibility annotations for standard calendar 

lionel-sim:
**Changes**
Description of changes...

- keep branch

**Changelog entry**

**Calendar**

- Add new `isFocusable` flag

**calendarManager**

- Add aria attributes
- Add keyboard navigations for month and year dropdowns
- Add keyboard navigations for chevrons

**standardCalendarDayView**

- Changed to grid format
- Add keyboard navigations

**dayCell**

- Update cursor and point event styling
- Add new `onFocus`, `onKeyDown`, `tabIndex`, and `role` props

**Additional information**

- You may refer to [ds figma](https://www.figma.com/design/wDAxbRmlq4xOn4UJ8zPuEF/%F0%9F%A7%91%F0%9F%8F%BB%E2%80%8D%F0%9F%A6%AF-Flagship-A11y-Annotations?node-id=1-21881&t=FsH5eKcod9610Vvq-4)


src/shared/internal-calendar/internal-calendar-year.tsx
```diff
@@ -98,11 +173,14 @@ export const InternalCalendarYear = ({
             ? "current-year"
             : "default";
 
+        const tabIndex = yearIndex === 0 ? 0 : -1;
```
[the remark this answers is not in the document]
lionel-sim: updated the logic to be aligned with the month selection. if the view calendar date year is within the current decade view the tab spot should land on the selected year, else it will be the first year in the decade (not the first year item)

src/shared/internal-calendar/standard/standard-calendar-day-view.tsx
```diff
@@ -74,17 +138,34 @@ export const StandardCalendarDayView = ({
...
+                    onMouseLeave={handleMouseLeaveCell}
+                    onBlur={handleMouseLeaveCell}
+                >
                     {week.map((day, dayIndex) => {
+                        const isDateToFocus = day.isSame(dateToFocus, "day");
+
+                        // First day of the month, should always be in tab order
+                        const isFirstDayOfMonth =
+                            day.date() === firstDay &&
+                            day.month() === calendarDate.month();
```
qroll, line 167: this seems inconsistent with month/year as there could be 2 different tab spots in 1 grid (the first day of month + selected date). though I might have missed a requirement

if there's only 1 tab spot, it would be cleaner to keep the logic for the focusable date in one place (i.e. the `dateToFocus` constant that you introduced) and we can simplify to `tabIndex={day.isSame(dateToFocus, "day") ? 0 : 1}`

src/shared/internal-calendar/standard/standard-calendar-day-view.tsx
qroll, line 1: noticed that the day view does not perform focusing after key navigation? this means that the screen reader doesn't read out the current visually focused day

if this is handled, would "Enter" or " " still need to check for the `hoverDate` or can the day cell directly select itself?

qroll, line 1: another inconsistency I noticed is that navigating to a day that belongs to the previous/next month does not change the month

The `Show Active Month Days` might be behaving weirdly due to this (navigating with the down arrow starting from Jun 1 results in the focus disappearing)

qroll (approved):

PR #693, opened 2025-05-28
[CCUBE-1784][XY] Add A11y Playground Theme (Colours + Font Tokens)

ninjasheeps:
**Changes**
- Added A11y Playground Theme's colour and font tokens

- keep branch


src/theme/index.ts
```diff
@@ -153,3 +153,15 @@ export const PATheme: ThemeSpec = {
     resourceScheme: "pa",
     componentScheme: "pa",
 };
+
+export const A11yTheme: ThemeSpec = {
```
qroll, line 157: we might want to keep the naming consistent - is the product called A11y Playground with a A11yPlaygroundTheme and `a11yPlayground` schemes, or A11y with a `A11yTheme`?

qroll (changes requested):

qroll (approved):

PR #697, opened 2025-06-03
[MISC][JH] Fix missing background-color for timeline indicator (current variant)

jhyeep:
**Changes**
Background-color was removed (by accident?) for the current variant timeline indicator. Adding this back

- delete branch

**Changelog entry**
- Fix missing background color for the `current` variant in the `Timeline` indicator


qroll (approved):
thanks for spotting and fixing this!

PR #698, opened 2025-06-04
Backport masthead changes to V2

qroll:
**Changes**

- Remove hardcoded background colour
- Set css variables to configure the horizontal spacing and max-width
- Have hardcoded the horizontal spacing to `0.75rem` to match `Layout.Container`. This still does not fully align with `Layout.Content` due to the dynamic horizontal spacing but main intention is to avoid the logo from being too close to the edge in V2
- [delete] branch


weili-govtech:
are we backporting the `stretch` prop?

qroll:
have reworded the commit to be more specific for V2

PR #700, opened 2025-06-09
Fix DropdownList search not working for capitalised substrings

jhyeep:
**Changes**
Filtering for components using `DropdownList` is returning no results when searching for capitalised substrings. The search string is converted to lowercase before comparison with the list items, but the item names are not.

- delete branch

**Changelog entry**
- Fix `DropdownList` search not working for capitalised substrings


qroll (changes requested):
thanks for the fix! could I trouble you to update `tests/input-select/input-select.spec.tsx` to cover this behaviour?

existing unit tests:
- `should support default search for title`
- `should support default search for label`

we could `spongecase` the strings here to test the case sensitivity

qroll (approved):

PR #701, opened 2025-06-10
Add workaround for button label in Chrome/Voiceover

qroll:
**Changes**

- Chrome/Voiceover gets stuck rereading the button contents as the countdown is changing too fast
- This implements an alternative way of exposing the countdown using `aria-describedby`
- [delete] branch

**Changelog entry**

- Fix reading of countdown in Chrome/Voiceover

### DIFF
--- a/src/otp-input/otp-input.tsx
+++ b/src/otp-input/otp-input.tsx
@@ -1,5 +1,6 @@
 import React, { useEffect, useRef, useState } from "react";
 import { FormErrorMessage } from "src/form/form-label";
+import { VisuallyHidden } from "../shared/accessibility";
 import { SimpleIdGenerator, StringHelper } from "../util";
 import {
     CTAButton,
@@ -43,6 +44,8 @@ export const OtpInput = ({
     const [internalId] = useState(() => SimpleIdGenerator.generate());
 
     const hasError = !!errorMessage;
+    const errorId = `${internalId}-error`;
+    const timerId = `${internalId}-timer`;
 
     // =============================================================================
     // EFFECTS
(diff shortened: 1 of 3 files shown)


src/otp-input/otp-input.tsx
```diff
@@ -190,19 +193,33 @@ export const OtpInput = ({
...
+                      countDown > 0
+                          ? `Resend OTP in ${displaySeconds(countDown)}`
+                          : "Resend OTP",
+                  // use a fixed label to avoid repeated screen reader updates when the button is focused
+                  "aria-label":
+                      countDown > 0
+                          ? `Resend OTP in ${displaySeconds(cooldownDuration)}`
+                          : "Resend OTP",
+                  // use a secondary element to reflect the latest countdown
+                  "aria-describedby": timerId,
```
weili-govtech, line 216: works well in safari but chrome skips it, though at the end of the countdown, it still announces `Ready to resend OTP`

qroll: thanks so much for catching this! after some further tests with Voiceover:
1. `aria-describedby` doesn't get read on the button when the describing element has a `role="timer"` set in Safari
2. `aria-description` works in the latest Safari but it's still fairly new, so `aria-describedby` is still recommended
3. Chrome was skipping the button depending on where the automatic read-all function started. If you start on an element before the OtpInput, it works fine; if you start within the OtpInput, it skips the button. I found that moving the visually hidden items to render before the button fixes that somehow 🙃

weili-govtech: wow thanks for doing a deeper dive into this, I also learnt new things today :)

PR #705, opened 2025-06-13
Commportal 360

ryhuz:
**Changes**
- Add `size` prop to Pill
- Add large Pill

- [delete] branch

**Changelog entry**

- `Pill` now has `size` prop to show a large pill


ryhuz:
closing MR as the PA use-case will be using `button` instead

PR #707, opened 2025-06-13
[CCUBE-1784][XY] Add dark mode theme

ninjasheeps:
**Changes**
- Added DSThemeProvider for dark mode theme switching
- Added dark mode theme usage documentation in Storybook
- Added A11y Playground semantic tokens for dark mode

- [keep] branch

**Changelog entry**

- [BREAKING] Dark mode theme implementation

### DIFF
--- a/src/theme/colour-primitive/theme-helper.ts
+++ b/src/theme/colour-primitive/theme-helper.ts
@@ -28,11 +28,62 @@ export const ColourSpec: ThemeCollectionSpec<
     defaultValue: "lifesg",
 };
 
+const getDarkColourSet = (scheme: ColourScheme): PrimitiveColourSet => {
+    // Check if explicit dark colour set exists
+    const explicitDarkSet = ExplicitDarkColourSpec.collections[scheme];
+    if (explicitDarkSet) {
+        return explicitDarkSet;
+    }
+
+    // Fallback to light colours (products can add explicit dark sets as needed)
+    const lightSet = ColourSpec.collections[scheme];
+    return lightSet;
+};
+
+// Explicit dark colour sets for brands that have custom dark palettes
+// Uses light primitive colours if dark mode is handled through semantic tokens
+const ExplicitDarkColourSpec: ThemeCollectionSpec<
+    ColourCollectionsMap,
+    ColourScheme
+> = {
+    collections: {
+        lifesg: LifeSgColourSet,
+        bookingsg: BookingSgColourSet,
+        rbs: RBSColourSet,
+        mylegacy: MyLegacyColourSet,
+        ccube: CCubeColourSet,
+        oneservice: OneServiceColourSet,
+        pa: PAColourSet,
+        a11yplayground: A11yPlaygroundColourSet,
+    },
+    defaultValue: "lifesg",
+};
+
+// Dynamic dark colour spec that uses explicit colours
+const DarkColourSpec: ThemeCollectionSpec<ColourCollectionsMap, ColourScheme> =
+    {
+        collections: {
+            lifesg: getDarkColourSet("lifesg"),
+            bookingsg: getDarkColourSet("bookingsg"),
+            rbs: getDarkColourSet("rbs"),
+            mylegacy: getDarkColourSet("mylegacy"),
+            ccube: getDarkColourSet("ccube"),
+            oneservice: getDarkColourSet("oneservice"),
+            pa: getDarkColourSet("pa"),
+            a11yplayground: getDarkColourSet("a11yplayground"),
+        },
+        defaultValue: "lifesg",
+    };
+
 export const getPrimitiveColour = (key: keyof PrimitiveColourSet) => {
     return (props: StyledComponentProps): string => {
         const theme = props.theme;
+        const isDarkMode = theme?.colourMode === "dark";
+
+        // Select the appropriate color spec based on theme mode
+        const spec = isDarkMode ? DarkColourSpec : ColourSpec;
         const colorSet: PrimitiveColourSet = getCollection(
-            ColourSpec,
+            spec,
             theme?.colourScheme
         );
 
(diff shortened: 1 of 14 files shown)


qroll (changes requested):
could this branch be rebased on top of master? to ensure you have the latest changes

1. run git fetch to pull the latest repo changes
2. switch to this branch
3. run `git rebase origin/master`

src/theme/index.ts
```diff
@@ -165,3 +168,61 @@ export const A11yPlaygroundTheme: ThemeSpec = {
...
+
+// =============================================================================
+// THEME COLLECTIONS
+// =============================================================================
+
+// Theme presets with light and dark variants
+export const LifeSGTheme = {
+    light: LifeSGThemeBase,
+    dark: { ...LifeSGThemeBase, colourMode: "dark" as const },
+} as const;
```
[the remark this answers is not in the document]
ninjasheeps: was using it for autocomplete purposes, but on second look, it's not necessary to have it. will remove, thanks for highlighting!

src/theme/colour-primitive/theme-helper.ts
```diff
@@ -28,18 +28,61 @@ export const ColourSpec: ThemeCollectionSpec<
...
+        const spec = isDarkMode ? DarkColourSpec : ColourSpec;
+        const colourSet: PrimitiveColourSet = getCollection(
+            spec,
             theme?.colourScheme
         );
 
-        if (theme?.overrides?.primitiveColour) {
-            return getValue(colorSet, key, theme.overrides.primitiveColour);
+        // Handle case where colourSet might be undefined
+        if (!colourSet) {
```
qroll, line 63: on second thought, is there a case where `colourSet` is undefined? as we would always fallback to `ColourSpec` and the `lifesg` scheme if the theme is not available

src/theme/ds-theme-provider.tsx
```diff
@@ -0,0 +1,68 @@
+import { ReactNode, useEffect, useState } from "react";
+import { ThemeProvider } from "styled-components";
+import { ColourMode, ThemeSpec } from "./types";
+import { getSystemColourMode } from "./use-theme-mode";
+
+export interface DSThemeProviderProps {
+    theme: ThemeSpec;
+    children: ReactNode;
+}
```
qroll, line 9: could this be moved to `types.ts`? it's a convention to put public types there, and would help avoid accidental edits

src/theme/use-theme-mode.ts
```diff
@@ -0,0 +1,51 @@
+import { useContext } from "react";
+import { ThemeContext } from "styled-components";
+import { ColourMode, ThemeSpec } from "./types";
+
+export interface DSTheme {
+    colourMode: ColourMode;
+    theme: ThemeSpec | null;
+}
```
qroll, line 8: as above, to move to `types.ts`

src/theme/use-theme-mode.ts
```diff
@@ -0,0 +1,51 @@
...
+// Hook to access the current theme mode and theme object
+export const useDSTheme = (): DSTheme => {
+    const theme = useContext(ThemeContext) as ThemeSpec;
+
+    if (!theme) {
+        console.warn(
+            "useDSTheme must be used within a DSThemeProvider or ThemeProvider"
+        );
+        return {
+            colourMode: "light" as ColourMode,
```
qroll, line 19: casting not needed as Typescript can pick up the type from the functions' return value

```suggestion
            colourMode: "light",
```

tests/theme/dark-mode.spec.tsx
```diff
@@ -0,0 +1,191 @@
...
+    color: ${Colour["text"]};
+`;
+
+// Component to test the useDSTheme hook
+const ThemeModeDisplay = () => {
+    const { colourMode } = useDSTheme();
+    return <div data-testid="theme-mode">{colourMode}</div>;
+};
+
+describe("Dark Mode Theme", () => {
```
qroll, line 23: we could assert that the tokens are mapped correctly in styled-components

for example we could render the test component like so:

```
const ThemeModeDisplay = () => {
    const { colourMode } = useDSTheme();
    return (
        <div>
            <div data-testid="theme-mode">{colourMode}</div>
            <div>
                <TestComponent data-testid="theme-component" />
            </div>
        </div>
    );
};
```

then the first test case can have these assertions
```
        expect(getByTestId("theme-mode")).toHaveTextContent("light");
        expect(getByTestId("theme-component")).toHaveStyle({
            color: "#282828",
            background: "#ffffff",
        });
```

tests/theme/dark-mode.spec.tsx
qroll, line 1: I'm thinking this could be merged with `theme-colour.spec.tsx` instead

tests/theme/dark-mode.spec.tsx
qroll, line 23:
```suggestion
describe("Colour mode", () => {
```

tests/theme/dark-mode.spec.tsx
```diff
@@ -0,0 +1,191 @@
...
+    it("should use light mode by default", () => {
+        const { getByTestId } = render(
+            <DSThemeProvider theme={LifeSGTheme.light}>
```
qroll, line 26: based on the unit test description, this should take the theme object without an explicit colour mode?

```suggestion
            <DSThemeProvider theme={LifeSGTheme}>
```

tests/theme/dark-mode.spec.tsx
qroll, line 23: we should also test the system preference behaviour

may need to mock `window.matchMedia` to return the expected values and trigger the change events, a possible method can be found [here](https://stackoverflow.com/questions/39830580/jest-test-fails-typeerror-window-matchmedia-is-not-a-function)

qroll (changes requested):

PR #720, opened 2025-07-02
[CCUBE-1833][XY] Add border to drop shadow components

ninjasheeps:
**Changes**
Add border and update drop shadows of components that are not visible in dark mode

- [delete] branch

**Additional information**

- You may refer to this [figma file](https://www.figma.com/design/5MXHzLCMH5w8HeRYw0gsAZ/%F0%9F%8F%B7-Design-Tokens-Sandbox?node-id=5932-7483&t=VV6eLP8VSSVYTJxF-4) for the updated changes

### DIFF
--- a/src/navbar/navbar.styles.tsx
+++ b/src/navbar/navbar.styles.tsx
@@ -33,15 +33,29 @@ export const Wrapper = styled.div<StyleProps>`
     left: 0;
     right: 0;
     width: 100%;
-    border-bottom: ${Border["width-010"]} ${Border["solid"]} ${Colour["border"]};
-    box-shadow: 0 2px 8px rgba(from ${Colour["bg-inverse-subtle"]} r g b / 16%);
+    border-bottom: ${(props) =>
+        props.theme?.colourMode === "dark"
+            ? `${Border["width-010"](props)} ${Border["solid"](props)} ${Colour[
+                  "border"
+              ](props)}`
+            : "none"};
+    box-shadow: ${(props) =>
+        props.theme?.colourMode === "dark"
+            ? "none"
+            : `0 2px 8px rgba(from ${Colour.Primitive["neutral-30"](
+                  props
+              )} r g b / 16%)`};
 `;
 
 export const Nav = styled.nav<StyleProps>`
-    height: ${(props) =>
-        props.$compress
+    height: ${(props) => {
+        const baseHeight = props.$compress
             ? NAVBAR_HEIGHT.compress
-            : NAVBAR_HEIGHT.notCompress}rem;
+            : NAVBAR_HEIGHT.notCompress;
+        return props.theme?.colourMode === "dark"
+            ? `calc(${baseHeight}rem - 1px)`
+            : `${baseHeight}rem`;
+    }};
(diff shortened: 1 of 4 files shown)


src/navbar/navbar.styles.tsx
```diff
@@ -33,7 +33,8 @@ export const Wrapper = styled.div<StyleProps>`
     left: 0;
     right: 0;
     width: 100%;
-    box-shadow: 0 1px 3px rgba(0, 0, 0, 0.1);
+    border-bottom: ${Border["width-010"]} ${Border["solid"]} ${Colour["border"]};
+    box-shadow: 0 2px 8px rgba(from ${Colour["bg-inverse-subtle"]} r g b / 16%);
```
qroll, line 37: visually, this looks odd in dark mode, it feels more like a white glow. are shadows common in dark modes?

src/navbar/navbar.styles.tsx
```diff
@@ -33,7 +33,8 @@ export const Wrapper = styled.div<StyleProps>`
     left: 0;
     right: 0;
     width: 100%;
-    box-shadow: 0 1px 3px rgba(0, 0, 0, 0.1);
+    border-bottom: ${Border["width-010"]} ${Border["solid"]} ${Colour["border"]};
```
qroll, line 36: will need to account for the extra 1px height introduced by the border

the Nav `height` should be `calc(${xxx}rem - 1px)` to compensate

src/navbar/navbar.styles.tsx
```diff
@@ -33,14 +33,29 @@ export const Wrapper = styled.div<StyleProps>`
...
+            ? `${Border["width-010"](props)} ${Border["solid"](props)} ${Colour[
+                  "border"
+              ](props)}`
+            : "none"};
+    box-shadow: ${(props) =>
+        props.theme?.colourMode === "dark"
+            ? "none"
+            : `0 2px 8px rgba(from ${Colour.Primitive["neutral-30"](
+                  props
+              )} r g b / 16%)`};
```
qroll, line 47: for readability, we can apply the styles at once based on the condition check

```
    ${(props) => {
        return props.theme?.colourMode === "dark"
            ? css`
                  border-bottom: ${Border["width-010"](props)}
                      ${Border["solid"](props)} ${Colour["border"]};
              `
            : css`
                  box-shadow: 0 2px 8px
                      rgba(
                          from ${Colour.Primitive["neutral-30"](props)} r g b /
                              16%
                      );
              `;
    }}
```

qroll (changes requested):
should card, menu and smart app banner also apply the same border/box-shadow logic?

ninjasheeps:
updated card, menu and smart app banner to apply the same border/box-shadow logic.
since box-shadow is not used in light mode anymore, i reverted the colour & opacity to the current design in figma (using neutral-50 and 24% opacity)

qroll (changes requested):
I just noticed that it's back to the primitive rather than `bg-inverse`, I'm assuming they'll be updated anyway when the shadow tokens are in?

PR #721, opened 2025-07-03
Accordion a11y (part 1)

qroll:
**Changes**

- It was reported that the content in a collapsed accordion item could still be accessed by screen readers. For example, you can still tab to the link in `This is the second item` of [this example](https://designsystem.life.gov.sg/react/index.html?path=/docs/content-accordion--docs)
- The fix is to add the `inert` prop to make the element inaccessible when the accordion item is collapsed
- Created a shared util to handle some React 19 compatibility, and applied to other uses of `inert` in the codebase
- [delete] branch

TODO, covering in a separate PR: reworking the accordion structure to match Figma specs

### DIFF
--- a/src/countdown-timer/countdown-timer.tsx
+++ b/src/countdown-timer/countdown-timer.tsx
@@ -3,6 +3,7 @@ import { useEffect, useRef, useState } from "react";
 import { useInView } from "react-intersection-observer";
 import { useMediaQuery } from "react-responsive";
 import { useTheme } from "styled-components";
+import { inertValue } from "../shared/accessibility";
 import { Breakpoint } from "../theme";
 import { TimeHelper } from "../util/time-helper";
 import {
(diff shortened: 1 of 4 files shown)


custom-types/react-augment.d.ts
```diff
@@ -1,12 +1,13 @@
 // https://stackoverflow.com/questions/58469229/react-with-typescript-generics-while-using-react-forwardref/58473012
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
qroll, line 2: the import is needed to expose existing React modules, so have ignored the error

src/accordion/types.ts
qroll, line 1: added missing selectors for completeness

src/shared/accessibility/index.ts
```diff
@@ -8,3 +9,17 @@ export const VisuallyHidden = styled.div`
     position: absolute;
     white-space: nowrap;
 `;
+
+/**
+ * Handle compatibility of `inert` across React versions.
+ * 19+ provides official support as a boolean, while it is handled as string in older versions
+ */
+export const inertValue = (value: boolean | undefined) => {
```
qroll, line 17: referenced from https://github.com/adobe/react-spectrum/pull/7519/files

weili-govtech (approved):

PR #722, opened 2025-07-03
[MISC][XY] Accessibility improvements for button

ninjasheeps:
**Changes**
- Add option to make it focusable when disabled

- [delete] branch


src/button/button.tsx
qroll, line 1: I see what you intended to do with the event listeners now

a thought, we could probably destructure the `onClick` prop and only forward it if the button is not disabled

the click, key press events will still propagate to the parent container but I think that's up to the consumer to handle 🤔

we only ensure the standard React `onClick` (which is the most common usage) won't be fired

```
const { onClick, ...otherProps } = props;
<Main
  onClick={disabled ? undefined : onClick}
>
```

we can have an Accessibility section in the docs to explain the use of `focusableWhenDisabled` along with the caveats

qroll (changes requested):

src/button/button.tsx
```diff
@@ -28,7 +30,11 @@ const DefaultComponent = (props: ButtonProps, ref: ButtonRef) => {
         <Main
             ref={ref}
             data-testid={otherProps["data-testid"] || "button"}
-            disabled={disabled}
+            disabled={disabled && !focusableWhenDisabled}
+            aria-disabled={disabled}
+            aria-busy={loading}
+            tabIndex={disabled && focusableWhenDisabled ? 0 : undefined}
```
qroll, line 36: we can remove `tabIndex` too, since non-disabled buttons are always focusable and disabled buttons are not by default

qroll (changes requested):

qroll (approved):

qroll (approved):
thanks for including the unit tests 😄

PR #726, opened 2025-07-08
[MISC][XY] Accessibility improvements for breadcrumb

ninjasheeps:
**Changes**
Improve accessibility implementation for breadcrumb

- delete branch

**Changelog entry**

*Caret*
- Add aria-hidden to hide decorative element from screen readers

*Item*
- Add aria-current to last element on breadcrumb

*nav*
- Add aria-label to indicate the type of navigation for screen readers

**Additional information**

- You may refer to this [figma file](https://www.figma.com/design/wDAxbRmlq4xOn4UJ8zPuEF/%F0%9F%A7%91%F0%9F%8F%BB%E2%80%8D%F0%9F%A6%AF-Flagship-A11y-Annotations?node-id=288-798)

### DIFF
--- a/src/breadcrumb/breadcrumb.tsx
+++ b/src/breadcrumb/breadcrumb.tsx
@@ -114,7 +114,7 @@ export const Breadcrumb = ({
 
             if (index === links.length - 1 || !link.href) {
                 element = (
-                    <CurrentLabel weight="semibold">
+                    <CurrentLabel weight="semibold" as="span">
                         {link.children}
                     </CurrentLabel>
                 );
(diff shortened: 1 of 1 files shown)


src/breadcrumb/breadcrumb.tsx
qroll, line 1: should the `CurrentLabel` get a tab index 0 so that they are tabbable and can be read in focus navigation?

qroll: oh yep, thanks for bringing this up.

one consideration was this storybook example, the second item is not focusable, so the "hierarchy" is skipped which feels wrong:

Screenshot 2025-07-10 at 5 56 29 PM

on the other hand
- the user can still navigate to it in reading mode
- it's probably unlikely to have non-links in the middle of a breadcrumb
- yes, it would indeed be confusing to mark this as an interactive element

so I'm good to skip the tab index

PR #730, opened 2025-07-14
[MISC][JH] tag changes for accessibility

jhyeep:
**Changes**
- add `aria-hidden` prop for `Tag` icon
- add `role="presentation"` for nested `Label` to prevent it from being announced as separate semantic element (removes extra "group" announcement by screenreader)

- delete branch

### DIFF
--- a/src/tag/tag.tsx
+++ b/src/tag/tag.tsx
@@ -21,11 +21,11 @@ export const Tag = ({
         >
             {iconPosition === "left" &&
                 icon &&
-                React.cloneElement(icon, { "aria-hidden": "true" })}
+                React.cloneElement(icon, { "aria-hidden": true })}
             <Label role="presentation">{children}</Label>
             {iconPosition === "right" &&
                 icon &&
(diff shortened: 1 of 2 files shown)


src/tag/tag.tsx
```diff
@@ -18,9 +19,13 @@ export const Tag = ({
             $interactive={interactive}
             {...otherProps}
         >
-            {iconPosition === "left" && icon}
-            <Label>{children}</Label>
-            {iconPosition === "right" && icon}
+            {iconPosition === "left" &&
+                icon &&
+                React.cloneElement(icon, { "aria-hidden": "true" })}
```
qroll, line 24: we can use boolean directly here, React will handle the casting to the right html attribute

can the other instance be updated as well?

```suggestion
                React.cloneElement(icon, { "aria-hidden": true })}
```

qroll (changes requested):

qroll (approved):
thanks for updating button too!

PR #732, opened 2025-07-14
[MISC][XY] Accessibility improvements to Navbar

ninjasheeps:
**Changes**

_Navbar_
- add `aria-label="Main navigation menu"`
- add `aria-hidden` to links using icons to prevent it from announcing the link as an image

_Brand_
- add `aria-label` for primary and secondary branding
- add `NonClickable` so screen reader doesn't read non-interactive logos as a link

_Drawer_
- add `aria-label="Mobile navigation menu"`
- add arrow keyboard navigation

- keep branch

**Additional information**

- You may refer to this [figma](https://www.figma.com/design/wDAxbRmlq4xOn4UJ8zPuEF/%F0%9F%A7%91%F0%9F%8F%BB%E2%80%8D%F0%9F%A6%AF-Flagship-A11y-Annotations?node-id=1647-949)
- This update only includes changes to elements used by A11y Playground (navbar-items, brand & drawer)

### DIFF
--- a/src/navbar/brand.tsx
+++ b/src/navbar/brand.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { Clickable, NonClickable } from "./brand.styles";
+import { Container } from "./brand.styles";
 import { BrandType, NavbarBrandingProps } from "./types";
 import { ImageWithFallback } from "../shared/image-with-fallback/image-with-fallback";
 
@@ -34,36 +34,33 @@ export const Brand = ({
         }
     };
 
-    // Determine if this should be clickable - brands are clickable by default when onClick is provided
+    // Determine if this should be clickable - brands are clickable by default
     const isClickable = !!onClick;
 
-    const ariaLabel =
-        type === "primary"
+    const logoLabel = isClickable
+        ? type === "primary"
             ? `Go to ${resources.brandName} homepage`
-            : `Go to ${resources.brandName}`;
+            : `Go to ${resources.brandName}`
+        : resources.brandName;
 
-    if (isClickable) {
-        return (
-            <Clickable
-                data-id={dataId}
-                data-testid={testId}
-                $type={type}
-                href="#"
-                onClick={handleClick}
-                aria-label={ariaLabel}
-            >
-                <ImageWithFallback src={resources.logoSrc} alt="" />
-            </Clickable>
-        );
-    }
+    const props = isClickable
+        ? {
+              role: "link",
+              tabIndex: 0,
+              onClick: handleClick,
+          }
+        : {};
 
     return (
-        <NonClickable data-id={dataId} data-testid={testId} $type={type}>
-            <ImageWithFallback
-                src={resources.logoSrc}
-                alt={resources.brandName}
-            />
-        </NonClickable>
+        <Container
+            data-id={dataId}
+            data-testid={testId}
+            $type={type}
+            as={isClickable ? "a" : "div"}
+            {...props}
+        >
+            <ImageWithFallback src={resources.logoSrc} alt={logoLabel} />
+        </Container>
     );
 };
 
--- a/src/navbar/brand.tsx
+++ b/src/navbar/brand.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { Clickable } from "./brand.styles";
+import { Clickable, NonClickable } from "./brand.styles";
 import { BrandType, NavbarBrandingProps } from "./types";
 import { ImageWithFallback } from "../shared/image-with-fallback/image-with-fallback";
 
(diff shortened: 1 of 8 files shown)


qroll:
qn: does the drawer sidebar require arrow key navigation? the default tab focus seem sufficient to me since these are non-floating link items

src/navbar/brand.tsx
qroll, line 1: actually, since Clickable and NonClickable have the same styles, you can use the same styled component and change the underlying element via `as` props

```
return (
  <Container
    data-id={dataId}
    data-testid={testId}
    as={isClickable ? "a" : "div"}
    onClick={isClickable ? handleClick : undefined}
    {...props}
  >
```

src/navbar/brand.tsx
```diff
@@ -29,24 +29,41 @@ export const Brand = ({
...
+            ? `Go to ${resources.brandName} homepage`
+            : `Go to ${resources.brandName}`;
+
+    if (isClickable) {
+        return (
+            <Clickable
+                data-id={dataId}
+                data-testid={testId}
+                $type={type}
+                href="#"
```
qroll, line 51: don't set the href (it becomes a functioning link that might not reflect the right url or visited state), continue to set a `role="link"` to prevent Lighthouse from complaining about SEO

planning to audit the usage of anchor tags and expose href props across the navigation elements in a future enhancement

src/navbar/brand.tsx
```diff
@@ -29,24 +29,41 @@ export const Brand = ({
...
+                onClick={handleClick}
+                aria-label={ariaLabel}
```
qroll, line 53: we can continue to set the label on the `alt` instead

qroll (changes requested):

src/navbar/navbar.tsx
```diff
@@ -292,7 +291,7 @@ const Component = <T,>(
     const renderNavbar = () => {
         return (
             <Layout.Content stretch={isStretch}>
-                <Nav $compress={compress}>
+                <Nav $compress={compress} aria-label="Main navigation menu">
```
qroll, line 294: suggestion: expose a `headerLabel` and `drawerLabel` prop to support customisation of the aria-label

src/navbar/navbar-items.tsx
```diff
@@ -62,6 +62,25 @@ export const NavbarItems = <T,>({
         setShowSubMenu(false);
     };
 
+    const addAriaHiddenToIcon = (children: React.ReactNode) => {
```
qroll, line 65: hmm given that the navbar items are totally customisable, I think the `aria-hidden` handling should be the responsibility of the consumer

also, for icons, as they indicate some functionality within a navbar, they shouldn't be hidden and should be labelled appropriately with `aria-label`

qroll (changes requested):

ninjasheeps:
tab navigation is indeed sufficient for a simple navigation menu

arrow key navigation for the drawer sidebar is not required for WCAG compliance, but it is considered best practice for menus with submenus (ref: [W3C](https://www.w3.org/WAI/GL/wiki/Using_ARIA_menus#Keyboard_Actions_if_focus_is_on_a_menu_item))

i added it in with the context of adding aria menu roles for submenus in the future, perhaps i should look into moving it into `menu` instead of putting it in `drawer` since it is not required?

src/navbar/drawer.tsx
```diff
@@ -106,7 +146,15 @@ const Component = (
 
     return (
         <Wrapper ref={ref} data-testid="drawer">
-            <Container $show={show} $viewHeight={viewHeight}>
+            <Container
+                as="nav"
```
qroll, line 150: we can declare Container as `styled.nav` directly, so there's no need to pass in an additional prop

src/navbar/types.ts
```diff
@@ -134,4 +136,6 @@ export interface NavbarProps<T = void> extends NavbarSharedProps {
     /** Specifies if masthead should be rendered */
     masthead?: boolean | undefined;
     layout?: "default" | "stretch" | undefined;
+    /** Custom aria-label for the main navigation header */
+    headerLabel?: string | undefined;
```
qroll, line 140: include `drawerLabel` here as well as this is the entry point for all props

make sure to forward it to `<Drawer>` in `navbar.tsx`

src/navbar/drawer.tsx
```diff
@@ -56,6 +52,50 @@ const Component = (
...
+        (event: React.KeyboardEvent<HTMLDivElement>) => {
+            if (!show) return;
+
+            if (event.key === "Escape") {
+                event.preventDefault();
+                onClose?.();
+                // Return focus to the hamburger menu button
+                setTimeout(() => {
+                    const hamburgerButton = document.querySelector(
+                        '[data-testid="button__mobile-menu"]'
```
qroll, line 65: hmm we shouldn't use `data-testid` as that's meant for testing. also this might not work in some Storybook pages as it'll always pick the first match

can we save a ref for the MobileMenuButton and pass that down to Drawer? something like

```
const mobileMenuRef = useRef<HTMLButtonElement>(null);

<MobileMenuButton
   ref={mobileMenuRef}
>

<Drawer
   mobileMenuRef={mobileMenuRef}
>
```

and in drawer, we can do this directly:
```
mobileMenuRef.current?.focus()
```

qroll (changes requested):

src/navbar/drawer.tsx
qroll: closing as I saw that you've updated the role to `dialog`, which is more appropriate as this is an overlaying element

qroll (approved):

PR #735, opened 2025-07-15
Fix masthead dark mode

### LINKED ISSUE #724: Masthead in dark mode is unreadable
**Component**
`Masthead`

**Describe the bug**
In dark mode, the background of the information panel turns dark while the text remains black, making it unreadable

**To Reproduce**

1. Go to https://designsystem.life.gov.sg/react/index.html?path=/story/navigation-masthead--default
2. Switch to dark mode
3. Click 'How to identify' to expand the information panel

**Expected behavior**
The masthead should follow the behaviour in the [SGDS storybook](https://webcomponent.designsystem.tech.gov.sg/?path=/story/components-masthead--basic&globals=theme:night)

**Screenshots**

Image

qroll:
**Changes**

Resolves #724 

- [delete] branch

_Load SGDS themes_
- Import the css from `@govtechsg/sgds-web-component` and inject the values at build time
- This uses the PostCSS plugin to insert style tags into the `<head>` at run time
- To prevent duplicate style tags from being inserted (e.g. when both Masthead and Navbar are conditionally rendered on a page), a unique id is set

_Pin SGDS components to major version_
- Note: this is a **breaking change** as consumers have to update their CSP whitelist if applicable
- This ensures that Masthead and css are kept in sync as much as possible and reduce the possibility of unexpected breaking changes
- Ideally we could have imported the React version of the Masthead from the library as well, however I faced difficulties with loading it in CRA/NextJS apps and decided to leave this for now
- Updated the script url to match the latest example provided in the [installation instructions](https://www.webcomponent.designsystem.tech.gov.sg/?path=/docs/getting-started-installation--docs#method-2-using-cdn)

**Changelog entry**

- Fix `Masthead`
- [BREAKING] Pin SGDS `Masthead` to v3. If your site has whitelisted the url in the Content Security Policy previously, update it to match the new value.

```diff
- https://cdn.jsdelivr.net/npm/@govtechsg/sgds-web-component/Masthead/index.js
+ https://cdn.jsdelivr.net/npm/@govtechsg/sgds-web-component@3/components/Masthead/index.umd.js
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,6 +27,7 @@
                 "@dnd-kit/sortable": "^7.0.2",
                 "@floating-ui/dom": "^1.6.10",
                 "@floating-ui/react": "^0.26.23",
+                "@govtechsg/sgds-web-component": "^3.0.6",
                 "@inquirer/prompts": "^7.0.0",
                 "@lifesg/react-icons": "^1.6.0",
                 "@playwright/test": "^1.45.0",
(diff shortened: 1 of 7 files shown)


qroll:
> Ideally we could have imported the React version of the Masthead from the library as well, however I faced difficulties with loading it in CRA/NextJS apps and decided to leave this for now

to elaborate, I first encountered `tslib` error (even though it was installed), and then another error with `HTMLElement is not defined`

I suspect SSR is a problem and I want to avoid consumers having to do this each time
```
dynamic(
	() => import("@lifesg/react-design-system/...").then((mod) => mod.Component),
	{ ssr: false },
);
```

can possibly revisit this in the future

choo95dev (approved):
LGTM

PR #737, opened 2025-07-16
[MISC][ZY] Improve a11y for Form.Input

tanzyy96:
Use simpleidgenerator for Form.Input, also enabling tab access for disabled Input

**Changes**
- Update `Form.Input` to use `form-field-{generatedId}` instead of just `form-field`
- Update `disabled` state to `readonly` + `aria-disabled` to allow keyboard tab access
- Added more stories to cover a11y cases
- [delete] branch

**Changelog entry**
- Fix accessibility warnings for `Form.Input`
- [BREAKING] Update `Form.Input` to no longer apply `disabled` attribute to underlying `input` element

https://github.com/user-attachments/assets/157af52e-6873-48de-b5f3-5c3c76c3f22f

### DIFF
--- a/src/form/form-input.tsx
+++ b/src/form/form-input.tsx
@@ -3,12 +3,13 @@ import { Input } from "../input";
 import { InputRef } from "../input/types";
 import { FormWrapper } from "./form-wrapper";
 import { FormInputProps } from "./types";
+import { SimpleIdGenerator } from "../util";
 
 const Component = (props: FormInputProps, ref: InputRef): JSX.Element => {
     const {
         label,
         errorMessage,
-        id = "form-field",
+        id = `form-field-${SimpleIdGenerator.generate()}`,
         "data-error-testid": errorTestId,
         "data-testid": testId,
         layoutType,
--- a/src/form/form-input.tsx
+++ b/src/form/form-input.tsx
@@ -34,6 +34,7 @@ const Component = (props: FormInputProps, ref: InputRef): JSX.Element => {
 
     return (
         <FormWrapper
+            id={inputId}
             data-testid={testId}
             label={label}
             errorMessage={errorMessage}
--- a/src/form/form-input.tsx
+++ b/src/form/form-input.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState } from "react";
 import { Input } from "../input";
 import { InputRef } from "../input/types";
 import { FormWrapper } from "./form-wrapper";
@@ -9,7 +9,7 @@ const Component = (props: FormInputProps, ref: InputRef): JSX.Element => {
     const {
         label,
         errorMessage,
-        id = `form-field-${SimpleIdGenerator.generate()}`,
+        id,
         "data-error-testid": errorTestId,
         "data-testid": testId,
         layoutType,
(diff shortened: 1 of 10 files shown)


qroll:
the subtitle is read together with the label, whereas the requirement is to associate the subtitle with the input with `aria-describedby` instead

to achieve this, we would need to split label and description into two separate containers

src/input/input.tsx
```diff
@@ -120,13 +120,14 @@ const Component = (
...
-                    disabled={disabled}
+                    aria-disabled={disabled}
                     value={updatedValue}
                     onChange={handleChange}
                     type={type}
-                    readOnly={readOnly}
+                    readOnly={readOnly || disabled}
                     $showClear={showClear}
                     $styleType={styleType}
+                    tabIndex={0}
```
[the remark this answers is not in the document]
tanzyy96: removed as not needed

src/form/form-input.tsx
```diff
@@ -25,9 +26,13 @@ const Component = (props: FormInputProps, ref: InputRef): JSX.Element => {
         ...otherProps
     } = props;
 
+    const [inputId] = useState(
+        () => id ?? `form-field-${SimpleIdGenerator.generate()}`
+    );
```
qroll, line 31:
```suggestion
    const [internalId] = useState(
        () => `form-field-${SimpleIdGenerator.generate()}`
    );
    const inputId = id ?? internalId;
```

src/form/form-input.tsx
```diff
@@ -25,9 +26,13 @@ const Component = (props: FormInputProps, ref: InputRef): JSX.Element => {
...
+
     return (
         <FormWrapper
-            id={id}
+            id={inputId}
```
qroll, line 35: on further thought I think the main `data-testid` should be set on `FormWrapper`

this lets consumers query the main container, then they can further query the children as needed

there's also not much point to default as `id` because it could be dynamically generated now, which is useless for unit tests

src/form/form-input.tsx
```diff
@@ -45,8 +50,8 @@ const Component = (props: FormInputProps, ref: InputRef): JSX.Element => {
             xxlCols={xxlCols}
         >
             <Input
-                id={`${id}-base`}
-                data-testid={testId || id}
+                id={`${inputId}-base`}
+                data-testid={testId || inputId}
```
qroll, line 54: and the input will receive the suffixed id and data-testid

```suggestion
                data-testid={testId ? `${testId}-base` : undefined }
```

src/form/form-wrapper.tsx
```diff
@@ -51,6 +51,14 @@ export const FormWrapper = ({
         return !!errorMessage;
     };
 
+    const hasSubtitleLabel = (): boolean => {
+        return typeof label === "object" && !!label?.subtitle;
+    };
+
+    const getSubtitleId = (): string => {
+        return `${id}-label-subtitle`;
```
qroll, line 59: just subtitle?
```suggestion
        return `${id}-subtitle`;
```

src/form/form-wrapper.tsx
```diff
@@ -141,6 +149,8 @@ export const FormWrapper = ({
             "aria-invalid": isInvalidState(),
             "aria-describedby": isInvalidState()
                 ? getErrorTestMessageId()
+                : hasSubtitleLabel()
+                ? getSubtitleId()
                 : undefined,
```
qroll, line 154: `aria-describedby` supports multiple ids. we can concatenate error (takes precedence?) + subtitle together, separated by space https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-describedby#id_reference_list

qroll (changes requested):

src/form/form-wrapper.tsx
tanzyy96: Using `-label-subtitle` because we are concatenating `-subtitle` inside `FormLabel`, which is passed `{testId}-label`. Do we want to apply it differently?

PR #753, opened 2025-07-23
[MOL-19007][GW] Add menu component

LowGinWee:
**Changes**
Add Menu component (to be used with Avatar/Navbar component)

- delete branch

**Changelog entry**

- Add Menu popover component

**Additional information**

- You may refer to this [MOL-19007](https://sgtechstack.atlassian.net/browse/MOL-19007)
- You may refer to this [MOL-19016](https://sgtechstack.atlassian.net/browse/MOL-19016)

### DIFF
--- a/stories/menu/menu.mdx
+++ b/stories/menu/menu.mdx
@@ -8,16 +8,18 @@ import { PropsTable } from "./props-table";
 
 ## Overview
 
-A component which appears when hovering or clicking on a target element and displays a menu, consisting of menu sections containing links or items.  
-It is commonly used together with components such as **Avatar** and **Navbar** with dropdown menus.
+A component which appears when hovering or clicking on a target element and
+displays a menu. It is commonly used in combination with
+[Avatar](/docs/navigation-avatar--docs) and
+[Navbar](/docs/navigation-navbar--docs).
 
 This module contains:
 
 -   **Menu** - a utility component that handles the triggering of the menu, extending [PopoverTrigger](/docs/overlays-popoverv2--docs)
--   **Menu.Content** - a presentational component of the menu
+-   **Menu.Content** - the main wrapper containing the menu items
 -   **Menu.Section** - groups menu items and optionally shows a section label
--   **Menu.Item** - a menu item with optional label and subLabel
--   **Menu.Link** - a menu item that renders as a hyperlink
+-   **Menu.Item** - a menu item with optional label and sublabel
+-   **Menu.Link** - a menu item that contains a hyperlink
 
 ```tsx
 import { Menu } from "@lifesg/react-design-system/menu";
@@ -27,12 +29,12 @@ import { Menu } from "@lifesg/react-design-system/menu";
 
 ## Menu content
 
-<Canvas of={MenuStories.ContentOnly} />
+<Canvas of={MenuStories.MenuContent} />
 
 ## Resize and flip
 
 As the menu extends [PopoverTrigger](/docs/overlays-popoverv2--docs), resize and
-flip can be configured
+flip behaviour can be configured
 
 <Canvas of={MenuStories.ResizeAndFlip} />
 
--- a/stories/menu/menu.stories.tsx
+++ b/stories/menu/menu.stories.tsx
@@ -44,51 +44,42 @@ const menuContent = (
 export const Default: StoryObj<Component> = {
     render: (_args) => {
         return (
-            <>
-                <Menu menuContent={menuContent}>
-                    <Avatar>Name</Avatar>
-                </Menu>
-            </>
+            <Menu menuContent={menuContent}>
+                <Avatar>Name</Avatar>
+            </Menu>
         );
     },
 };
 
-export const ContentOnly: StoryObj<Component> = {
+export const MenuContent: StoryObj<Component> = {
     render: (_args) => {
         return menuContent;
     },
 };
 
 export const ResizeAndFlip: StoryObj<Component> = {
-    render: (_args) => {
+    args: {
+        menuContent: menuContent,
+        zIndex: 10,
+    },
+    render: (args) => {
         return (
             <>
-                <Menu zIndex={10} menuContent={menuContent}>
+                <Menu {...args}>
                     <Avatar>Name</Avatar>
                 </Menu>
-                <Menu enableResize={true} zIndex={10} menuContent={menuContent}>
+                <Menu enableResize={true} {...args}>
                     <Avatar>Name</Avatar>
                 </Menu>
-                <Menu
-                    enableResize={true}
-                    enableFlip={false}
-                    zIndex={10}
-                    menuContent={menuContent}
-                >
+                <Menu enableResize={true} enableFlip={false} {...args}>
                     <Avatar>Name</Avatar>
                 </Menu>
-                <Menu
-                    enableResize={false}
-                    enableFlip={false}
-                    zIndex={10}
-                    menuContent={menuContent}
-                >
+                <Menu enableResize={false} enableFlip={false} {...args}>
                     <Avatar>Name</Avatar>
                 </Menu>
             </>
         );
     },
-
     decorators: [
         GridDecorator({
             columns: 4,
--- a/stories/menu/props-table.tsx
+++ b/stories/menu/props-table.tsx
@@ -6,18 +6,6 @@ const MENU_DATA: ApiTableSectionProps[] = [
     {
         attributes: [
             {
-                name: "menuContent",
-                description: "The menu content inside this menu",
-                propTypes: ["Menu.Content"],
-                mandatory: true,
-            },
-            {
-                name: "data-testid",
-                description: "The test identifier for the component",
-                propTypes: ["string"],
-                defaultValue: `"menu"`,
-            },
-            {
                 name: "-",
                 description: (
                     <>
@@ -34,11 +22,23 @@ const MENU_DATA: ApiTableSectionProps[] = [
                 ),
             },
             {
+                name: "menuContent",
+                description: "The menu content inside this menu",
+                propTypes: ["Menu.Content"],
+                mandatory: true,
+            },
+            {
+                name: "data-testid",
+                description: "The test identifier for the component",
+                propTypes: ["string"],
+                defaultValue: `"menu"`,
+            },
+            {
                 name: "position",
                 description: (
                     <>
-                        The visual position of the <code>Popover</code> in
-                        relation to its trigger
+                        The visual position of the menu in relation to its
+                        trigger
                     </>
                 ),
                 propTypes: [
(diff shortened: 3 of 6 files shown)


src/menu/menu-link.styles.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+// =============================================================================
+// MENU ITEM STYLES
+// =============================================================================
+export const MenuLinkDiv = styled.div`
+    display: flex;
+    flex-direction: column;
+`;
+
+export const Link = styled(Typography.LinkMD)`
```
LowGinWee, line 14: More or less copied over from the original menu link with some changes

src/menu/menu-section.styles.tsx
```diff
@@ -0,0 +1,28 @@
...
+// =============================================================================
+// MENU SECTION STYLES
+// =============================================================================
+export const Section = styled.div<MenuSectionStyleProps>`
+    ${(props) =>
+        props.$showDivider &&
+        css`
+            border-top: ${Border["width-010"]} ${Border["solid"]}
+                ${Colour["border"]};
+        `}
```
LowGinWee, line 21: decided to allow dividers for the section component only (and only allow the top divider), cause if menuItems/Links have dividers, they should be in another separate section for accessibility (where dividers can then be added)

src/menu/menu.tsx
```diff
@@ -0,0 +1,91 @@
...
+                handleViewportChange();
+            });
+    };
+
+    const handleViewportChange = () => {
+        if (panelRef.current) {
+            const rect = panelRef.current.getBoundingClientRect();
+            setMaxHeight(window.innerHeight - rect.top);
+        }
+    };
```
LowGinWee, line 55: handles scroll behaviour of the menu (this needs to be used with any of the position bottoms, and enableFlip = false).

probably good to refactor this into a util/custom hook, so that it can also be used by dropdowns and popover.
(i see a similar issue with dropdowns - not sure if fixed already, should the dropdown is at the bottom of the screen, it extends past the viewport)

LowGinWee:
to test with navbar^

src/menu/menu.tsx
qroll: for scroll behaviour, what about the size middleware? https://floating-ui.com/docs/size

however, there's a limitation with size in that all floating props needs to be passed down to the scrollable container, else the height computation doesn't work well and breaks the flip logic

I'm using it here with flip in `element-with-dropdown` and I'm still looking to improve the default behaviour, but it seems that exposing the floating props is unavoidable
https://github.com/LifeSG/react-design-system/blob/improve-auto-positioning/src/shared/dropdown-wrapper/element-with-dropdown.tsx#L156

src/menu/menu.styles.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+// =============================================================================
+// MENU STYLES
+// =============================================================================
+export const MenuPanel = styled.div<MenuPanelStyleProps>`
+    border-radius: ${Radius["md"]};
+    border: ${Border["width-010"]} ${Border["solid"]} ${Colour["border"]};
+    background: ${Colour["bg"]};
+    box-shadow: 0 ${Border["width-020"]} ${Border["width-080"]} 0
```
qroll, line 18: box shadow can be hardcoded for now, there will be separate tokens for those in the future

src/menu/menu.styles.tsx
```diff
@@ -0,0 +1,28 @@
...
+        rgb(from ${Colour["border-stronger"]} r g b / 16%);
```
qroll, line 19: can we stick to `Primitive["neutral-50"]` here?

src/menu/menu.tsx
qroll, line 1: ok I see that MenuTrigger is reusing PopoverTrigger, I didn't think of that. do you forsee any possible divergence of behaviour (other than aria annotations) or would the positioning logic be the same?

qroll (commented):
initial comments

src/menu/menu.tsx
LowGinWee: updated/refactored to use "size" into popover content (thought that popover content should also have this same behaviour) (updated storybook to see all the variations)

src/popover-v2/popover-trigger.tsx
```diff
@@ -50,10 +53,19 @@ export const PopoverTrigger = ({
...
-            flip(),
+            enableFlip && flip(),
             shift({
                 limiter: limitShift(),
             }),
+            enableContentScroll &&
+                size({
+                    apply({ availableHeight, elements }) {
+                        Object.assign(elements.floating.style, {
+                            maxHeight: `${availableHeight}px`,
```
qroll, line 64: the side effect of setting overflow y on the floating wrapper is this:

Screenshot 2025-07-25 at 10 09 24 PM

it has to be set on the container with the border instead

src/menu/index.ts
```diff
@@ -0,0 +1,6 @@
+export * from "./menu";
+export * from "./menu-item";
+export * from "./menu-link";
+export * from "./menu-section";
+export * from "./menu-content";
```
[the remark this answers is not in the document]
LowGinWee: updated

src/popover-v2/types.ts
```diff
@@ -38,6 +38,8 @@ export interface PopoverV2TriggerProps {
     delay?:
         | { open?: number | undefined; close?: number | undefined }
         | undefined;
+    enableFlip?: boolean | undefined;
+    enableContentScroll?: boolean | undefined;
```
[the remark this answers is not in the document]
LowGinWee: updated

src/menu/menu-content.styles.tsx
```diff
@@ -0,0 +1,18 @@
+import styled from "styled-components";
+import { Border, Colour, Radius } from "../theme";
+
+// =============================================================================
+// MENU CONTENT STYLES
+// =============================================================================
+export const MenuPanel = styled.div`
```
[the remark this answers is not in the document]
LowGinWee: udpated

src/menu/menu-section.tsx
```diff
@@ -0,0 +1,26 @@
...
+    label,
+    showDivider = true,
+    "data-testid": testId = "menu-section",
+    ...otherProps
+}: MenuSectionProps): JSX.Element => {
+    // =============================================================================
+    // RENDER FUNCTIONS
+    // =============================================================================
+    return (
+        <Section
```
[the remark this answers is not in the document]
LowGinWee: updated

src/menu/menu-item.styles.tsx
```diff
@@ -0,0 +1,16 @@
+import styled from "styled-components";
+import { Colour, Spacing } from "../theme";
+import { Typography } from "../typography";
+
+// =============================================================================
+// MENU ITEM STYLES
+// =============================================================================
+export const MenuItemDiv = styled.div`
```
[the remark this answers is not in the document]
LowGinWee: updated to Li and extend from Li

src/popover-v2/popover-trigger.tsx
LowGinWee: updated to set the maxHeight and overflow to the popoverCard.

cant really think of an elegant solution to pass both of them if a customPopoverContent is used instead -> user has to receive popoverResizeProps to set the maxHeight and overflow themselves

src/menu/menu-content.styles.tsx
```diff
@@ -0,0 +1,41 @@
...
+// =============================================================================
+interface MenuPanelStylesProps {
+    $overflow?: overflow | undefined;
+    $maxHeight?: number | undefined;
+}
+
+// =============================================================================
+// MENU CONTENT STYLES
+// =============================================================================
+export const MenuPanel = styled.div<MenuPanelStylesProps>`
```
qroll, line 16: would be good to set the scrollbar style too. this is used for dropdown

```
&::-webkit-scrollbar {
    width: 14px;
}

&::-webkit-scrollbar-track {
    background: transparent;
}

&::-webkit-scrollbar-thumb {
    background: ${Colour["bg-inverse-subtlest"]};
    border: 5px solid transparent;
    border-radius: ${Radius["full"]};
    background-clip: padding-box;
}
```

src/menu/menu-section.tsx
```diff
@@ -0,0 +1,27 @@
...
+            $showDivider={showDivider}
+            data-testid={testId}
+            role="group"
+            {...otherProps}
+        >
+            {label && <Label weight="semibold">{label}</Label>}
```
qroll, line 21: something that I missed out: we'll also need to associate the section with the label, so that the screenreader will read out "Menu link, Category 1, group" label when a list item in the section is first focused

this can be done by using `aria-labelledby` to point to the label. we can generate a unique id with:
```
const [internalId] = useState(() => SimpleIdGenerator.generate());
```

the usage would be something like this:
```
<Section aria-labelledby={internalId}>
  <Label id={internalId}>
```

src/popover-v2/popover-trigger.tsx
qroll: this works! it's unavoidable that information needs to be passed down to the floating component at this point

src/popover-v2/types.ts
```diff
@@ -9,17 +9,27 @@ export interface PopoverV2Props {
     onMobileClose?: (() => void) | undefined;
 }
 
+export interface PopoverResizeProps {
```
qroll, line 12:
```suggestion
export interface PopoverRenderProps {
```

src/popover-v2/types.ts
```diff
@@ -9,17 +9,27 @@ export interface PopoverV2Props {
...
+
 export type PopoverV2Position = Position | AlignedPosition;
 
 export interface PopoverV2TriggerProps {
     children: React.ReactNode;
-    popoverContent: string | JSX.Element | (() => React.ReactNode);
+    popoverContent:
+        | string
+        | JSX.Element
+        | ((resizeProps?: PopoverResizeProps | undefined) => React.ReactNode);
```
qroll, line 32:
```suggestion
        | ((renderProps: PopoverRenderProps) => React.ReactNode);
```

src/popover-v2/types.ts
```diff
@@ -9,17 +9,27 @@ export interface PopoverV2Props {
...
+    maxHeight?: number | undefined;
+}
+
 export type PopoverV2TriggerType = "click" | "hover";
 
 type Position = "top" | "right" | "bottom" | "left";
 type Alignment = "start" | "end";
 type AlignedPosition = `${Position}-${Alignment}`;
 
+export type overflow = "visible" | "hidden" | "clip" | "scroll" | "auto";
```
qroll, line 23:
```suggestion
export type PopoverOverflowType = "visible" | "hidden" | "clip" | "scroll" | "auto";
```

qroll (changes requested):

src/popover-v2/types.ts
LowGinWee: updated

src/popover-v2/types.ts
LowGinWee: updated

src/menu/menu-section.tsx
LowGinWee: updated

src/menu/menu-content.styles.tsx
LowGinWee: udpate. also updated the popoverCard for default popovercontent

qroll (approved):

PR #755, opened 2025-07-24
Improve auto positioning for dropdown elements

qroll:
**Changes**

- [delete] branch
- Ensure the parent receives the floating props directly to that size computation and behaviour are handled correctly by Floating UI
- These props are passed through the render function and also context for convenience

**Changelog entry**

- Improve auto positioning of dropdown in date pickers and select inputs

### DIFF
--- a/src/shared/dropdown-wrapper/element-with-dropdown.tsx
+++ b/src/shared/dropdown-wrapper/element-with-dropdown.tsx
@@ -16,17 +16,26 @@ import {
     useInteractions,
     useTransitionStyles,
 } from "@floating-ui/react";
-import { RefObject, useContext, useRef } from "react";
+import {
+    CSSProperties,
+    RefObject,
+    createContext,
+    useContext,
+    useRef,
+} from "react";
 import { useResizeDetector } from "react-resize-detector";
 import { ThemeContext } from "styled-components";
 import { useFloatingChild } from "../../overlay/use-floating-context";
 import { Breakpoint } from "../../theme";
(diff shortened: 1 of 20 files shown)


src/shared/dropdown-wrapper/element-with-dropdown.tsx
```diff
@@ -176,32 +245,14 @@ export const ElementWithDropdown = ({
                     <FloatingFocusManager
                         context={context}
                         modal={false}
-                        initialFocus={dropdownRef}
+                        initialFocus={-1}
```
qroll, line 248: the wrapper component was removed

but this disables automatic focus, components will handle their own dropdown focus behaviour anyway

didn't observe any change in behaviour

src/time-range-picker/combobox-picker/combobox-picker.tsx
```diff
@@ -285,8 +295,6 @@ export const ComboboxPicker = ({
     };
 
     const renderDropdown = () => {
-        if (!dropdownOpen) return;
```
qroll, line 288: this caused an error in the unit tests

```
 Error: Uncaught [TypeError: relatedTarget.hasAttribute is not a function]
```

likely the dropdown was not mounted in time when Floating UI tried to access it

src/util/merge-refs.ts
qroll, line 1: a simple util to simplify assigning multiple refs to 1 element, inspired by libraries like [react-merge-refs](https://www.npmjs.com/package/react-merge-refs)

PR #776, opened 2025-08-07
[MOL-19265][JC] add expandCollapseClickCallback to Accordion

gnzhen:
**Changes**
Add new optional props in Accordion - a callback func to be triggered when user click on show all/hide all

- delete branch

**Changelog entry**

- Add expandCollapseClickCallback in `Accordion`

**Additional information**

- It's to solve issue 2 in [MOL-19265](https://sgtechstack.atlassian.net/browse/MOL-19265)

### DIFF
--- a/src/accordion/accordion.tsx
+++ b/src/accordion/accordion.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { AccordionContext } from "./accordion-context";
 import { AccordionItem } from "./accordion-item";
 import {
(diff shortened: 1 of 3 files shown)


src/accordion/types.ts
```diff
@@ -8,6 +8,7 @@ export interface AccordionProps {
     "data-testid"?: string | undefined;
     className?: string | undefined;
     headingLevel?: number | undefined;
+    expandCollapseClickCallback?: () => void,
```
qroll, line 11: rename callback to match react naming conventions, add `| undefined` to support [exactOptionalPropertyTypes](https://www.typescriptlang.org/tsconfig/exactOptionalPropertyTypes.html)

also propose to provide the next state of the accordion

```suggestion
    onExpandCollapseChange?: ((expanded: boolean) => void) | undefined,
```

src/accordion/accordion.tsx
```diff
@@ -19,21 +19,23 @@ const AccordionBase = ({
...
 }: AccordionProps): JSX.Element => {
     const [expandAll, setExpandAll] = useState<boolean>(
         initialDisplay === "expand-all"
     );
 
-    const handleExpandCollapseClick = (event: React.MouseEvent) => {
+    const handleExpandCollapseClick = (event: React.MouseEvent, callback?: () => void) => {
         event.preventDefault();
         setExpandAll((prevExpandValue) => !prevExpandValue);
+        callback?.();
```
qroll, line 31: actually since the callback is destructured, there's no need to pass it as an argument

this can directly be `onExpandCollapseAll?.()`

qroll (changes requested):

qroll (approved):

PR #787, opened 2025-08-14
[MOL-19180][LP] Accessibility enhancements for Drawer

linhpk12:
**Changes**
Description of changes...

- [delete] branch

**Changelog entry**

**Additional information**

### DIFF
--- a/src/drawer/drawer.styles.tsx
+++ b/src/drawer/drawer.styles.tsx
@@ -27,20 +27,23 @@ interface StyleProps {
 const FLOATING_TRANSITION_STYLES = css`
     transition-property: right, visibility;
 
-    &[data-status="open"],
-    &[data-status="close"] {
-        transition-duration: ${Motion["duration-800"]};
-    }
-
-    &[data-status="initial"],
-    &[data-status="close"] {
+    &[data-status="initial"] {
         right: -100%;
         visibility: hidden;
     }
 
     &[data-status="open"] {
+        transition-duration: ${Motion["duration-800"]};
+        transition-timing-function: ${Motion["ease-entrance"]};
         right: 0;
         visibility: visible;
+    }
+
+    &[data-status="close"] {
+        transition-duration: ${Motion["duration-800"]};
+        transition-timing-function: ${Motion["ease-exit"]};
+        right: -100%;
+        visibility: hidden;
     }
 `;
 
--- a/src/drawer/drawer.styles.tsx
+++ b/src/drawer/drawer.styles.tsx
@@ -77,22 +77,21 @@ export const Header = styled.div`
     align-items: center;
     gap: ${Spacing["spacing-16"]};
     padding: ${Spacing["spacing-32"]} ${Spacing["spacing-16"]}
-        ${Spacing["spacing-16"]} ${Spacing["spacing-64"]};
+        ${Spacing["spacing-16"]} calc(2rem + ${Spacing["spacing-32"]});
     background-color: ${Colour.bg};
     border-bottom: ${Border["width-010"]} ${Border.solid} ${Colour.border};
 
     ${MediaQuery.MaxWidth.lg} {
         gap: ${Spacing["spacing-8"]};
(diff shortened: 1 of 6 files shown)


src/drawer/drawer.styles.tsx
```diff
@@ -73,26 +73,28 @@ export const Container = styled.div<StyleProps>`
 
 export const Header = styled.div`
     top: 0;
-
     display: flex;
     align-items: center;
     gap: ${Spacing["spacing-16"]};
     padding: ${Spacing["spacing-32"]} ${Spacing["spacing-16"]}
-        ${Spacing["spacing-16"]};
+        ${Spacing["spacing-16"]} ${Spacing["spacing-64"]};
```
qroll, line 80: spacing is in px while the icon is in rem, using only the spacing token here will lead to misalignment when a different font size is set by the user

you can set `calc(<icon width> + <Spacing token>)` instead for more accurate positioning

src/drawer/drawer.styles.tsx
```diff
@@ -73,26 +73,28 @@ export const Container = styled.div<StyleProps>`
...
         padding: ${Spacing["spacing-32"]} ${Spacing["spacing-20"]}
-            ${Spacing["spacing-16"]};
+            ${Spacing["spacing-16"]} ${Spacing["spacing-64"]};
     }
 `;
 
 export const CloseButton = styled(ClickableIcon)`
     color: ${Colour.icon};
-    padding: 0;
+    padding: 2px;
```
qroll, line 93: what's the use of the custom padding here?

src/drawer/drawer.tsx
```diff
@@ -52,6 +64,36 @@ export const Drawer = ({
         event.stopPropagation();
     };
 
+    const handleTabKeyPress = (e: KeyboardEvent) => {
```
qroll, line 67: suggest to use the Floating UI's `FloatingFocusManager` to perform focus trapping instead, tab detection is not sufficient for Android Talkback

you may refer to the latest changes for the Filter component: https://github.com/LifeSG/react-design-system/pull/788/files

src/drawer/drawer.styles.tsx
```diff
@@ -73,26 +73,28 @@ export const Container = styled.div<StyleProps>`
...
+    position: absolute;
     order: -1; // show button on the left of the header
```
qroll, line 95: since the button is now repositioned, the custom flex order styles can be cleaned up from the Header

qroll (changes requested):

src/drawer/drawer.styles.tsx
```diff
@@ -73,26 +73,27 @@ export const Container = styled.div<StyleProps>`
...
 
     ${MediaQuery.MaxWidth.lg} {
         gap: ${Spacing["spacing-8"]};
         padding: ${Spacing["spacing-32"]} ${Spacing["spacing-20"]}
-            ${Spacing["spacing-16"]};
+            ${Spacing["spacing-16"]} calc(2rem + ${Spacing["spacing-24"]});
     }
 `;
 
 export const CloseButton = styled(ClickableIcon)`
```
qroll, line 91: set the svg height as `Font.Spec["heading-lh-md"]` instead to align the button with the heading text

src/drawer/drawer.tsx
```diff
@@ -61,34 +79,45 @@ export const Drawer = ({
...
         >
-            <Container
-                $show={show}
-                data-testid="drawer"
-                onClick={handleClick}
-                role="dialog"
-                aria-labelledby={id}
-                onTransitionEnd={handleDialogVisibility}
-                {...otherProps}
+            <FloatingFocusManager
```
qroll, line 82: FloatingFocusManager should only be rendered when the drawer is active per the fix here https://github.com/LifeSG/react-design-system/pull/788/files

if conditional rendering breaks the animation, take a look at `useTransitionStatus()` to only unmount when the animation completes https://floating-ui.com/docs/usetransition#usetransitionstatus

qroll (changes requested):

PR #788, opened 2025-08-19
[MISC][NT] Fixed overlay with aria-hidden when filters are open

nobeltws:
Fixed overlay with aria-hidden when filters are open in mobile view. When more than 1 filter component is rendered, floating focus is set to the last rendered filter, causing the other filters to be appended with `aria-hidden`

**Changes**
Fixed overlay with `aria-hidden` when more than 1 filter is rendered in mobile view

- [delete] branch

**Changelog entry**
- Fixed overlay with `aria-hidden` when more than 1 filter is rendered in mobile view

### DIFF
--- a/src/filter/filter.tsx
+++ b/src/filter/filter.tsx
@@ -18,6 +18,7 @@ import {
     FilterHeader,
     FilterHeaderButton,
     FilterTitle,
+    FloatingWrapper,
     MobileContainer,
     MobileOverlayContainer,
 } from "./filter.styles";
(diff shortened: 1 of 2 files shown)


src/filter/filter.tsx
```diff
@@ -145,28 +145,36 @@ const FilterBase = ({
...
+                                            onClick={handleDoneClick}
+                                            type="button"
+                                        >
+                                            {doneButtonLabel}
+                                        </FilterDoneButton>
+                                    </FilterFooter>
+                                </MobileContainer>
+                            </MobileOverlayContainer>
+                        </FloatingFocusManager>
+                    ) : undefined}
```
nobeltws, line 177: had to explicitly use `undefined` because `Overlay` children prop expects `JSX.Element | undefined`

qroll (changes requested):
this is logged to the console

```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?
```

I'm guessing it's because FloatingFocusManager isn't actually a DOM element, and Overlay is trying to set a ref on it

some potential workarounds
1. pass a custom ref prop to Overlay?
2. wrap the FloatingFocusManager in another `div`

PR #793, opened 2025-08-25
[MOL-19181][LP] Accessibility enhancement for Countdown Timer

linhpk12:
**Changes**
Description of changes...

- [keep | delete] branch

**Changelog entry**

**Additional information**

### DIFF
--- a/stories/countdown-timer/countdown-timer.stories.tsx
+++ b/stories/countdown-timer/countdown-timer.stories.tsx
@@ -1,4 +1,5 @@
 import type { Meta, StoryObj } from "@storybook/react";
+import dedent from "dedent";
 import { useState } from "react";
 import { Button } from "src/button";
 import { CountdownTimer } from "src/countdown-timer";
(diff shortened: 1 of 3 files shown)


shaunl89 (dismissed):

src/countdown-timer/countdown-timer.style.tsx
```diff
@@ -105,3 +105,14 @@ export const TimerIcon = styled(ClockIcon)<CountdownStyleProps>`
     color: ${(props) =>
         props.$warn ? Colour["icon-error"] : Colour["icon-primary"]};
 `;
+
+export const HiddenSpan = styled.span`
```
qroll, line 109: can use the existing `VisuallyHidden` component from the shared folder

src/countdown-timer/types.ts
```diff
@@ -31,6 +31,10 @@ interface CountdownTimerBaseProps extends React.HTMLAttributes<HTMLDivElement> {
     onNotify?: (() => void) | undefined;
     /** Called when countdown reaches 0 */
     onFinish?: (() => void) | undefined;
+    /** Specifies the interval (in seconds) for the polite notification */
+    politeInterval?: number | undefined;
+    /** Specifies the threshold (in seconds) for the assertive notification */
+    assertiveTimer?: number | undefined;
```
qroll, line 37: should this match with `notifyTimer` (changes the display from blue to red)? or does UX prefer to keep it separate?

src/countdown-timer/types.ts
```diff
@@ -31,6 +31,10 @@ interface CountdownTimerBaseProps extends React.HTMLAttributes<HTMLDivElement> {
     onNotify?: (() => void) | undefined;
     /** Called when countdown reaches 0 */
     onFinish?: (() => void) | undefined;
+    /** Specifies the interval (in seconds) for the polite notification */
+    politeInterval?: number | undefined;
```
qroll, line 35:
```suggestion
    /** Specifies the interval (in seconds) between polite announcements of the remaining time left */
    reminderInterval?: number | undefined;
```

stories/countdown-timer/countdown-timer.stories.tsx
```diff
@@ -18,7 +18,7 @@ export const Default: StoryObj<Component> = {
         return (
             <CountdownTimer
                 show
-                timer={15}
+                timer={200}
```
qroll, line 21: is this just for testing? or intended for demonstration purposes?

src/countdown-timer/countdown-timer.tsx
```diff
@@ -174,6 +244,10 @@ export const CountdownTimer = ({
                 inert={inertValue(!isVisible)}
                 $visible={isVisible}
                 $warn={warn}
+                tabIndex={0}
+                role="timer"
+                aria-label="Countdown timer"
+                onFocus={handleOnFocus}
```
qroll, line 250: can explain why a custom label needs to be set here and on focus?

src/countdown-timer/countdown-timer.tsx
```diff
@@ -96,6 +101,32 @@ export const CountdownTimer = ({
...
+            const initialMessage = `Starting countdown timer: ${getAccessibleTimeText(
+                remainingSeconds
+            )}`;
+            setInitialAnnouncement(initialMessage);
+        }
+    }, [isPlaying, remainingSeconds, timer]);
+
+    // Periodic announcements
+    useEffect(() => {
+        const timeElapsed = (timer || 0) - remainingSeconds;
```
qroll, line 116: the consumer can set the countdown with either `timer` or `timestamp`, so the new calculation logic will also need to cater for that instead of checking `timer` directly

suggestion: enhance the `useTimer` hook to derive the initial seconds

src/countdown-timer/countdown-timer.tsx
```diff
@@ -96,6 +101,32 @@ export const CountdownTimer = ({
         }
     }, [wrapperRef.current]);
 
+    // Initial announcement effect
+    useEffect(() => {
+        if (isPlaying && remainingSeconds === timer) {
```
qroll, line 106: as above, to also account for `timestamp` prop

qroll (changes requested):

src/countdown-timer/types.ts
linhpk12: i see figma does not specify this. should i use `notifyTimer ` for assertive notification (remove `assertiveTimer` and add default value for `notifyTimer`) ?

src/countdown-timer/countdown-timer.tsx
```diff
@@ -144,23 +186,53 @@ export const CountdownTimer = ({
...
+        }
+    };
     // =============================================================================
     // RENDER FUNCTION
     // =============================================================================
 
+    const renderInitialAnnouncement = () => (
+        <VisuallyHidden aria-live="polite" aria-atomic="true">
+            {initialAnnouncement}
+        </VisuallyHidden>
```
qroll, line 210: I noticed that in reading mode (where the screen reader navigates through all items), the "Start of..." gets announced as well. Wondering if there's a way to workaround this? Could the initialAnnouncement state be cleared after some time?

PR #794, opened 2025-08-26
[COMMPORTAL-805][SW] disable active/hover states for button, checkbox…

ryhuz:
**Changes**
disable active/hover states for button, checkbox, and toggle

- [delete] branch

**Changelog entry**

- [WARNING] active/hover state for button, checkbox, and toggle are only applicable for mouse events, and not touch (ie applicable on desktop but not on mobile with touch)

### DIFF
--- a/src/button/button.style.tsx
+++ b/src/button/button.style.tsx
@@ -47,7 +47,9 @@ export const Main = styled.button<MainStyleProps>`
 
                     &:hover,
                     &:active {
-                        background-color: ${Colour["bg-hover-neutral"]};
+                        @media (pointer: fine) {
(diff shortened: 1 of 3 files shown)


qroll:
any difference versus using media hover?
> @media (hover: hover)

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/hover

ryhuz:
> any difference versus using media hover?
> 
> ```
> @media (hover: hover)
> ```
> 
> https://developer.mozilla.org/en-US/docs/Web/CSS/@media/hover

doesn't work. tried and seems like on mobile it still triggers the hover effect

qroll (approved):

PR #803, opened 2025-08-29
[COMMPORTAL-1004][SW] add callback on opening mobile filter

ryhuz:
**Changes**
- [delete] branch

**Changelog entry**

- add callback on opening mobile filter in `Filter`

### DIFF
--- a/src/filter/types.ts
+++ b/src/filter/types.ts
@@ -41,6 +41,7 @@ export interface FilterProps {
     onClear?: (() => void) | undefined;
     /** @deprecated Forces mobile mode. Not for public use, will be removed in future versions */
     _dangerouslyForceMobile?: boolean | undefined;
+    onModalOpen?: (() => void) | undefined;
 }
 
 export interface FilterItemProps {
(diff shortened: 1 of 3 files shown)


src/filter/types.ts
```diff
@@ -41,6 +41,7 @@ export interface FilterProps {
     onClear?: (() => void) | undefined;
     /** @deprecated Forces mobile mode. Not for public use, will be removed in future versions */
     _dangerouslyForceMobile?: boolean | undefined;
+    onShowMobileFilter?: (() => void) | undefined;
```
qroll, line 44:
```suggestion
    /** Called with the next open state */
    onFilterOpenChange?: ((open) => void) | undefined;
```

stories/filter/props-table.tsx
```diff
@@ -71,6 +71,11 @@ const FILTER_DATA: ApiTableSectionProps[] = [
                 description: "The test identifier for the component",
                 propTypes: ["string"],
             },
+            {
+                name: "onShowMobileFilter",
+                description: "Called when filter is open on mobile",
```
qroll, line 76:
```suggestion
                name: "onModalOpen",
                description: "Called when filter modal is opened (mobile only)",
```

qroll (changes requested):

PR #806, opened 2025-09-02
[MISC][XY] Fix input + select read-only states to align with left margin

ninjasheeps:
**Changes**
Standardise "no-indent" styling for read-only states

- [delete] branch

**Changelog entry**

- [BREAKING] Styling changes to read-only states in input + select components

### DIFF
--- a/src/input/input.style.tsx
+++ b/src/input/input.style.tsx
@@ -23,12 +23,12 @@ export const InputElement = styled(BasicInput)<InputStyleProps>`
     ${(props) =>
         props.$styleType !== "no-border" &&
         css`
-            padding: ${Spacing["spacing-16"]}
-                ${props.$visuallyReadOnly
-                    ? 0
-                    : props.$showClear
-                    ? Spacing["spacing-16"]
-                    : Spacing["spacing-16"]};
+            padding-left: ${props.$visuallyReadOnly
+                ? 0
+                : Spacing["spacing-16"]};
+            padding-right: ${props.$visuallyReadOnly || props.$showClear
+                ? 0
+                : Spacing["spacing-16"]};
         `}
 `;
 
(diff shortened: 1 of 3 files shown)


qroll:
the mobile variant that wraps to next line is styled separately, this is the `Container` in `src/date-range-input/date-range-input.style.tsx`

Screenshot 2025-09-03 at 10 36 03 PM

src/input/input.style.tsx
```diff
@@ -22,8 +23,12 @@ export const InputElement = styled(BasicInput)<InputStyleProps>`
...
         props.$styleType !== "no-border" &&
         css`
-            padding-left: ${Spacing["spacing-16"]};
-            padding-right: ${props.$showClear ? 0 : Spacing["spacing-16"]};
+            padding: ${Spacing["spacing-16"]}
+                ${props.$visuallyReadOnly
+                    ? 0
+                    : props.$showClear
+                    ? Spacing["spacing-16"]
+                    : Spacing["spacing-16"]};
```
qroll, line 31: the logic should be:
padding left: if readonly, 0, else spacing-16
padding right: if readonly or showClear, 0, else spacing-16

this is because the clear button comes with its own padding already, so the additional right padding is not needed

src/shared/input-wrapper/input-wrapper.tsx
```diff
@@ -92,7 +92,7 @@ export const InputWrapper = styled(InputBox)<InputWrapperStyleProps>`
     position: relative;
     height: max-content;
     width: 100%;
-    padding: 0 ${Spacing["spacing-16"]};
+    padding: 0 0 0 ${(props) => (props.$readOnly ? "0" : Spacing["spacing-16"])};
```
qroll, line 95: hmm I don't think we should remove right padding here. for shorter fields, the text would be right against the edge

qroll (approved):

PR #809, opened 2025-09-08
[COMMPORTAL-813] feat: add prefix to OTP input

ryhuz:
**Changes**
add in a prefix option for OTP input

- [delete] branch

**Changelog entry**

- Added prefix option to `OTPInput`

### DIFF
--- a/src/otp-input/otp-input.tsx
+++ b/src/otp-input/otp-input.tsx
@@ -10,7 +10,7 @@ import {
     Wrapper,
 } from "./otp-input.styles";
 import { OtpInputProps } from "./types";
-import { strippedOtpFromAutofill, validateUserInput } from "./utils";
+import { stripOtpFromAutofill, validateUserInput } from "./utils";
 
 export const OtpInput = ({
     id,
@@ -102,7 +102,7 @@ export const OtpInput = ({
 
             /* handle autofill */
             const fieldIsEmpty = otpValues.every((num) => num === "");
-            const autofillOtp = strippedOtpFromAutofill(
+            const autofillOtp = stripOtpFromAutofill(
                 eventValue,
                 numOfInput,
                 prefix
@@ -156,7 +156,7 @@ export const OtpInput = ({
         if (!event.clipboardData) return;
 
         const pastedValue = event.clipboardData.getData("text");
-        const autofillOtp = strippedOtpFromAutofill(
+        const autofillOtp = stripOtpFromAutofill(
             pastedValue,
             numOfInput,
             prefix
@@ -254,11 +254,10 @@ export const OtpInput = ({
                 aria-label={`${numOfInput}-digit OTP input field`}
             >
                 {prefix && (
-                    <Prefix
-                        forwardedAs="span"
-                        aria-label="O T P prefix"
-                        data-testid="otp-prefix"
-                    >{`${prefix.value} ${prefix.separator}`}</Prefix>
+                    <Prefix forwardedAs="span" data-testid="otp-prefix">
+                        <VisuallyHidden>O T P prefix</VisuallyHidden>
+                        {`${prefix.value} ${prefix.separator}`}
+                    </Prefix>
                 )}
                 {otpValues.map((data, index) => {
                     return (
--- a/src/otp-input/utils.ts
+++ b/src/otp-input/utils.ts
@@ -3,7 +3,7 @@ import { OtpInputProps } from "./types";
 export const validateUserInput = (value: string, length = 1) =>
     !value ? false : RegExp(`^[0-9]{${length}}$`).test(value);
 
-export const strippedOtpFromAutofill = (
+export const stripOtpFromAutofill = (
     autofillValue: string,
     validOtpLength: number,
     prefix?: OtpInputProps["prefix"]
--- a/stories/otp-input/otp-input.stories.tsx
+++ b/stories/otp-input/otp-input.stories.tsx
@@ -32,6 +32,19 @@ export const WithErrorMessage: StoryObj<Component> = {
     decorators: [StoryDecorator({ maxWidth: true })],
 };
 
+export const WithPrefix: StoryObj<Component> = {
+    render: (_args) => {
+        return (
+            <OtpInput
+                numOfInput={6}
+                cooldownDuration={60}
+                prefix={{ value: "WGS", separator: "-" }}
+            />
+        );
+    },
+    decorators: [StoryDecorator({ maxWidth: true })],
+};
+
 export const WithCustomActionButton: StoryObj<Component> = {
     render: (_args) => {
         const numOfInput = 4;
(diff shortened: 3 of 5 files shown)


src/otp-input/otp-input.tsx
```diff
@@ -125,11 +127,15 @@ export const OtpInput = ({
         if (!event.clipboardData) return;
 
         const pastedValue = event.clipboardData.getData("text");
-        const pastedValueArr = pastedValue.split("");
-        if (pastedValue && validateUserInput(pastedValue, numOfInput)) {
-            setOtpValues(pastedValueArr);
+        const valueWithoutPrefix = pastedValue.replace(prefix + "-", "");
```
qroll, line 130: this will become `undefined-` when prefix is not specified, to check if prefix is truthy before runnign the replacement

src/otp-input/types.ts
```diff
@@ -13,6 +13,8 @@ export interface OtpInputProps extends React.AriaAttributes {
     actionButtonProps?: ButtonProps | undefined;
     /** The number of characters for the Otp */
     numOfInput: number;
+    /** OTP prefix. Usually 3 alpha-characters */
```
qroll, line 16:
```suggestion
    /** Optional OTP prefix, usually consisting of a few alphabetic characters */
```

src/otp-input/otp-input.tsx
qroll: actually can the `-` be part of the prefix too? or are we assuming `-` is always the separator

src/otp-input/otp-input.tsx
```diff
@@ -226,6 +232,11 @@ export const OtpInput = ({
                 role="group"
                 aria-label={`${numOfInput}-digit OTP input field`}
             >
+                {prefix && (
+                    <Typography.HeadingSM style={{ textWrap: "nowrap" }}>
```
qroll, line 236: 1. SM is 22px which seems kind of huge, default font size is 18px for most themes, checking if this is part of the design?
2. instead of inlining the style, declare a styled component
3. set `as` prop so that this doesn't render as a `h5` element

qroll (changes requested):

src/otp-input/otp-input.tsx
ryhuz: PA mock up uses SM. When looking on storybook with LSG theme, the size when SM looks to correspond with the OTP input box.

noted on points 2 and 3

src/otp-input/otp-input.tsx
ryhuz: ah found an issue.
prefix with OTP = `abc-12356`
when we display, aesthetically it's more pleasing if it's displayed `abc - 123456` (with spaces, and the numbers are boxes)

src/otp-input/otp-input.tsx
```diff
@@ -125,13 +157,8 @@ export const OtpInput = ({
...
         const pastedValue = event.clipboardData.getData("text");
-        const pastedValueArr = pastedValue.split("");
-        if (pastedValue && validateUserInput(pastedValue, numOfInput)) {
-            setOtpValues(pastedValueArr);
-            if (onChangeRef.current) {
-                onChangeRef.current(pastedValueArr);
-            }
-        } else {
+
+        if (!handleInputFullValue(pastedValue)) {
```
qroll, line 161: can handlePaste also include the strip logic? my thought process is even on desktop, user can also copy the whole OTP including the prefix

src/otp-input/otp-input.tsx
```diff
@@ -226,6 +250,9 @@ export const OtpInput = ({
                 role="group"
                 aria-label={`${numOfInput}-digit OTP input field`}
             >
+                {prefix && (
+                    <Prefix>{`${prefix.value} ${prefix.separator}`}</Prefix>
```
qroll, line 254: set `forwardedAs="span"`

src/otp-input/utils.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    if (prefix) {
+        const [autofillPrefixValue, autofillOtpValue] = autofillValue.split(
+            prefix.separator
+        );
+        return autofillPrefixValue === prefix.value &&
+            autofillOtpValue.length === validOtpLength
+            ? autofillOtpValue
+            : "";
+    }
```
qroll, line 24: for this condition, actually the original logic to trim the prefix is a safer approach. this way we can handle empty string for separator

we could probably check if autofill value starts with the prefix first, then slice out the prefix accordingly

qroll (changes requested):
can include a unit test to cover the new prefix behaviour

src/otp-input/utils.ts
ryhuz: i would rather we not allow empty separators. i think OTP standards would have some kind of separator, have never seen an OTP with prefix without separator.

qroll: defensive handling in case someone does pass in an empty string. `split("")` doesn't work at all

src/otp-input/otp-input.tsx
ryhuz: ok just need to clarify
- autofill otp expects field to be empty, ie if i already key in 1 number, the autofill won't work
- paste current behaviour is to completely overwrite if the clipboard value is valid
- do we want to follow either one of the behaviours, or to have them maintain their own behaviour?

src/otp-input/utils.ts
ryhuz: i mean, `split("")` will just not split it, which means the paste/autofill won't work. which technically is correct. so it's really consumer error at this point.

plus again, i've really never seen an OTP that is ABC123. that's just bad design

ryhuz: > "ABC".split("") gives you [A, B, C]

yeah then the autofill fails. that's the intention

src/otp-input/utils.ts
```diff
@@ -0,0 +1,28 @@
+import { OtpInputProps } from "./types";
+
+export const validateUserInput = (value: string, length = 1) =>
+    !value ? false : RegExp(`^[0-9]{${length}}$`).test(value);
+
+export const strippedOtpFromAutofill = (
```
qroll, line 6:
```suggestion
export const stripOtpFromAutofill = (
```

stories/otp-input/props-table.tsx
```diff
@@ -49,6 +49,12 @@ const DATA: ApiTableSectionProps[] = [
                 propTypes: ["number"],
                 mandatory: true,
             },
+            {
+                name: "prefix",
+                description:
+                    "Optional prefix to be displayed. When autofilling, this prefix + separator will be stripped from the OTP to be pasted",
```
qroll, line 55:
```suggestion
                    "Optional prefix to be displayed. When the OTP is autofilled or pasted from clipboard, the prefix value and separator will be stripped",
```

stories/otp-input/otp-input.stories.tsx
```diff
@@ -88,3 +88,16 @@ export const WithCustomActionButtonSelfHandle: StoryObj<Component> = {
     },
     decorators: [StoryDecorator({ maxWidth: true })],
 };
+
+export const WithPrefix: StoryObj<Component> = {
```
qroll, line 92: can move this story to under WithErrorMessage to match the position in Storybook docs

PR #819, opened 2025-09-10
Ccube 1937: Support container-agnostic UneditableSection

hhung2710:
**Changes**
Description of changes...

- [keep | delete] branch

**Changelog entry**

**Additional information**

### DIFF
--- a/stories/uneditable-section/props-table.tsx
+++ b/stories/uneditable-section/props-table.tsx
@@ -100,10 +100,14 @@ const MAIN_DATA: ApiTableSectionProps[] = [
             },
             {
                 name: "fullWidth",
-                description:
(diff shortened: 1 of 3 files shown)


src/uneditable-section/full-width-container.tsx
qroll, line 1: since all this container does is to render a child, there's no need for it to be a standalone component

basically the main uneditable section can do this:
```
// render logic
if (fullWidth) {
  return <FullWidthContainerProps $background={background} {...otherProps}>
     {renderChildren()}
  </FullWidthContainer>;  
}

// in uneditable-section.styles.tsx
export const FullWidthContainer = styled.div`
  ...
`
```

qroll (changes requested):

qroll (approved):

PR #834, opened 2025-09-23
[MISC][NT] Fix sidenav item hover highlight

nobeltws:
**Changes**
Fix sidenav hover highlight persisting even after cursor leaves the sidenav item. highlight will persist if there is a drawer on that item.

- [delete] branch

**Changelog entry**

- Fix `Sidenav` hover highlight persisting even after cursor leaves the sidenav item

### DIFF
--- a/src/sidenav/sidenav-item.tsx
+++ b/src/sidenav/sidenav-item.tsx
@@ -163,6 +163,10 @@ export const SidenavItem = ({
         setCurrentItem({ itemId: id, content: children });
     };
 
+    const handleMouseLeave = () => {
+        !isOpen && setCurrentItem(undefined);
+    };
+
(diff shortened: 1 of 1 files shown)


src/sidenav/sidenav-item.tsx
```diff
@@ -195,6 +199,9 @@ export const SidenavItem = ({
                 {...otherProps}
                 $highlight={isSelected || isCurrent}
                 {...getReferenceProps()}
+                onMouseLeave={() => {
+                    handleMouseLeave(isOpen);
+                }}
```
qroll, line 204: `isOpen` comes from state? could the function directly access the variable instead?

```suggestion
                onMouseLeave={handleMouseLeave}
```

PR #837, opened 2025-09-26
Update FilterItemCheckbox to support checkbox in tablet/mobile + Make minimizing optional

PhamQuocHuy12:
**Changes**

- Add showAsCheckboxInMobile for FilterItemCheckbox to display checkbox list for mobile view
- Add minimisableOptions for toggling View more/less option for the FilterItemCheckbox

### DIFF
--- a/stories/filter/filter-addons.stories.tsx
+++ b/stories/filter/filter-addons.stories.tsx
@@ -21,7 +21,7 @@ export const FilterCheckbox: StoryObj<Component> = {
             saveFilters,
             dismissFilters,
             clearFilters,
-        } = useFilters({ cat1: [], cat2: [] }); // your custom filter state management here
+        } = useFilters({ cat1: [], cat2: [], cat3: [] }); // your custom filter state management here
         return (
             <Filter
                 clearButtonDisabled={clearButtonDisabled}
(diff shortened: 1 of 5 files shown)


stories/filter/filter-addons.stories.tsx
```diff
@@ -90,6 +90,87 @@ export const FilterCheckbox: StoryObj<Component> = {
                                 },
                             ]}
                         />
+                        <Filter.Checkbox
+                            title="With more than 5 items with checkbox display for mobile view"
```
qroll, line 94:
```suggestion
                            title="With checkbox display for mobile view"
```

stories/filter/filter-addons.stories.tsx
```diff
@@ -90,6 +90,87 @@ export const FilterCheckbox: StoryObj<Component> = {
...
+                                    label: "Another Label 7",
+                                },
+                                {
+                                    value: "8",
+                                    label: "Another Label 8",
+                                },
+                            ]}
+                        />
+                        <Filter.Checkbox
+                            title="With more than 5 items but View more button is hidden"
```
qroll, line 134:
```suggestion
                            title="Without minimisable options"
```

stories/filter/filter-addons.stories.tsx
```diff
@@ -90,6 +90,87 @@ export const FilterCheckbox: StoryObj<Component> = {
...
+                            showAsCheckboxInMobile
```
qroll, line 135: can exclude this prop from this example, so we can reference the behaviour in both checkbox list and toggle variants

qroll (changes requested):
Noticed an extra gap between each checkbox on mobile, will need to update the styling accordingly

Image

PR #843, opened 2025-10-03
[MISC][T] Add accessibility to uneditable section component

FUnmthank02:
**Changes**
Add accessibility to uneditable section component

**Additional information**

- Refer to ticket [MOL-19629](https://sgtechstack.atlassian.net/browse/MOL-19629)

### DIFF
--- a/stories/uneditable-section/doc-elements.tsx
+++ b/stories/uneditable-section/doc-elements.tsx
@@ -8,8 +8,10 @@ export const SAMPLE_ITEMS: UneditableSectionItemProps[] = [
     },
     {
         label: "NRIC or FIN",
-        value: "S••••534J",
+        value: "S1234534J",
         displayWidth: "half",
+        maskRange: [1, 4],
+        maskState: "masked",
     },
     {
         label: "Date of birth",
--- a/stories/uneditable-section/uneditable-section.stories.tsx
+++ b/stories/uneditable-section/uneditable-section.stories.tsx
@@ -333,8 +333,10 @@ export const ComposingFromScratch: StoryObj<Component> = {
                         />
                         <UneditableSection.Item
                             label="NRIC or FIN"
-                            value="S••••534J"
+                            value="S1234534J"
                             displayWidth="half"
+                            maskRange={[1, 4]}
+                            maskState="masked"
                         />
                         <UneditableSection.Item
                             label="Residential Address"


qroll:
have identified other requirements based on the Figma requirements

1. header hierarchy
    - set the `forwardAs="h2"` prop on `Title` so that it does not disrupt the default page hierarchy
    - you can read up more on this [here](https://www.w3.org/WAI/tutorials/page-structure/headings/)
2. masked input button
    - it should have the `aria-pressed` state with "Display <label text>" OR read out "Display/Hide <label text>"  depending on the currently masked/unmasked state. these are mutually exclusive. could you check with UX on which is preferred?
    - you can read up more on aria-pressed [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-pressed)
3. custom description of the masked text
    - example given for NRIC was "starting with S and ending with 234A"
    - however due to the dynamic nature of our mask functionality, there isn't a quick solution for this
4. discrepancy for error with "Try again?" compared to masked input. check with UX if the correct behaviour which doesn't read out the original label is acceptable

src/uneditable-section/section-item.tsx
```diff
@@ -115,7 +115,7 @@ export const UneditableSectionItem = ({
                     <>
                         <ErrorIcon />
                         <ErrorLabel>Error</ErrorLabel>
-                        <TryAgainLabel>Try again?</TryAgainLabel>
+                        <TryAgainLabel title={label}>Try again?</TryAgainLabel>
```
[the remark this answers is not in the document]
qroll: to associate an element with a different label, there are different methods

1. aria-label - if only 1 source is needed to describe the element and can be represented as a plain string
2. aria-labelledby - if multiple sources describe the element
3. visually hidden text - text that's styled to be invisible but screen readers can read them out as part of the text

for this specific try again button, (3) is more appropriate

you may refer to the BoxContainer which uses the shared VisuallyHidden component to render extra text: https://github.com/LifeSG/react-design-system/blob/master/src/box-container/box-container.tsx#L150-L155

src/uneditable-section/section-item.tsx
```diff
@@ -161,6 +168,11 @@ export const UneditableSectionItem = ({
                 aria-busy={maskLoadingState === "loading"}
                 aria-live="polite"
                 type="button"
+                title={StringHelper.getMaskedDescription(
```
qroll, line 171: do not use `title` attribute. use `aria-label` instead to describe the masked input.  and `aria-label` should only be set on the masked children. ensure that this does not override the accessible label for the fail and loading state

src/uneditable-section/section-item.styles.tsx
```diff
@@ -47,6 +47,12 @@ export const IconContainer = styled.div`
     justify-content: center;
     color: ${Colour["icon-primary"]};
     margin-left: 0.5rem;
+    background-color: transparent;
+    border: none;
+
+    &:hover {
+        cursor: pointer;
+    }
```
qroll, line 55: can explain what the changes here are for?

qroll (changes requested):

PR #844, opened 2025-10-05
Accessible label for PhoneNumberInput

qroll:
**Changes**

- Update label + description association for PhoneNumberInput (which is dependent on InputGroup list variant)
- Retain the country code's combobox semantics even when readonly, so that it remains exposed to the screen reader in form mode
- Keyboard navigation will be covered separately as it affects a shared component and the requirements are pending confirmation
- [delete] branch

**Changelog entry**

New
- Improve accessibility of `PhoneNumberInput` and `InputGroup` list variant

Style fix
- Fix unreadable text in dark mode for readonly InputGroup and PhoneNumberInput

**Additional info**
- [Figma](https://www.figma.com/design/wDAxbRmlq4xOn4UJ8zPuEF/%F0%9F%A7%91%F0%9F%8F%BB%E2%80%8D%F0%9F%A6%AF-Flagship-A11y-Annotations?node-id=2069-801&p=f&m=dev)

### DIFF
--- a/src/form/form-input-group.tsx
+++ b/src/form/form-input-group.tsx
@@ -37,10 +37,11 @@ const Component = <T, V>(
     return (
         <FormWrapper
             id={inputId}
+            data-testid={testId}
             label={label}
             errorMessage={errorMessage}
-            disabled={otherProps.disabled}
             data-error-testid={errorTestId}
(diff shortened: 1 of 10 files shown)


nobeltws (approved):
looks good!

PR #845, opened 2025-10-08
[MOL-19653][SL] use disabled prop in FloatingFocusManager instead of unmounting to prevent loss of values

shaunl89:
…
**Changes**
Description of changes...

- [delete] branch

- You may refer to this [ticket](https://sgtechstack.atlassian.net/browse/MOL-19653)

### DIFF
--- a/src/filter/filter.tsx
+++ b/src/filter/filter.tsx
@@ -24,6 +24,7 @@ import {
     MobileOverlayContainer,
 } from "./filter.styles";
 import { FilterProps, Mode } from "./types";
+import { inertValue } from "../shared/accessibility";
 
 const FilterBase = ({
     customLabels,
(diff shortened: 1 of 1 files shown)


src/filter/filter.tsx
```diff
@@ -153,11 +153,12 @@ const FilterBase = ({
                     {toggleFilterButtonLabel}
                 </FilterButton>
                 <Overlay show={visible} disableTransition>
-                    {visible ? (
-                        <FloatingWrapper>
+                    {
+                        <FloatingWrapper aria-hidden={!visible}>
```
qroll, line 157: prefer use of `inert` which also prevents keyboard navigation via Tab

```suggestion
                        <FloatingWrapper inert={inertValue(!visible)}>
```

PR #854, opened 2025-10-15
[MISC][YF] accessibility enhancements for icon-button

Ranchu2000:
**Changes**
- added `aria-label` based on `styleType`
-  extended `disabled` prop to `aria-disabled`
- [delete] branch


qroll (changes requested):
Documentation:
- Could I also trouble you to include an `Accessibility` section for this component?It will match the documentation for [Button](https://designsystem.life.gov.sg/react/index.html?path=/docs/selection-and-input-button-base--docs#accessibility)
- Update `stories/icon-button/props-table.tsx` to include the new prop

PR #855, opened 2025-10-16
[A11Y][MOL-19639] Add accessibilities for Unit Number Input

tientranngoc-alt:
**Changes**
Description of changes...

- Add accessibilities to the Unit Number Input so that AT reader can correctly identify and read it. 

**Changelog entry**
- Add labels to number input components
- Add helper `formatPhraseWithPrefix `
- Add a logic to handle live message for AT reader to read

**Additional information**
- Refer to [MOL-19639](https://sgtechstack.atlassian.net/jira/software/c/projects/MOL/boards/1861?selectedIssue=MOL-19639&sprints=26571)

### DIFF
--- a/src/unit-number/unit-number-input.tsx
+++ b/src/unit-number/unit-number-input.tsx
@@ -355,7 +355,7 @@ export const UnitNumberInput = ({
                 autoComplete={autoComplete}
                 styleType="no-border"
             />
-            <VisuallyHidden id={unitLabelId}>Enter unit number</VisuallyHidden>
+            <VisuallyHidden aria-hidden id={unitLabelId}>Enter unit number</VisuallyHidden>
             {/** Live message for AT reader to read with the combination of prefix and current value for both floor input and unit input */}
             <VisuallyHidden id={liveMessageId} aria-live="polite">
                 {liveMessage}


src/shared/accessibility/index.ts
```diff
@@ -25,3 +25,7 @@ export const inertValue = (value: boolean | undefined) => {
 export const concatIds = (...selectors: (string | undefined)[]) => {
     return selectors.filter((selector) => !!selector).join(" ");
 };
+
+export const formatPhraseWithPrefix = (prefix: string, v: string) => {
+    return v ? `${prefix} ${Array.from(v).join(" ")}` : "";
+};
```
qroll: this can sit as a helper function in unit-number-input for now, as the format is very specific

src/unit-number/unit-number-input.tsx
```diff
@@ -293,6 +335,11 @@ export const UnitNumberInput = ({
                 autoComplete={autoComplete}
                 styleType="no-border"
             />
+            <VisuallyHidden id={floorInstructionId}>
```
qroll: should be `floorLabelId` instead since this dictates the function of the input

src/unit-number/unit-number-input.tsx
```diff
@@ -96,6 +110,27 @@ export const UnitNumberInput = ({
         updateValues(value);
     }, [value]);
 
+    // =============================================================================
+    // MEMO VALUEs
+    // =============================================================================
+
+    const liveMessage = useMemo(() => {
```
qroll: move this under the CONST/STATE/REF section instead

src/unit-number/unit-number-input.tsx
```diff
@@ -285,6 +320,13 @@ export const UnitNumberInput = ({
...
                 aria-label="floor-input"
+                aria-labelledby={concatIds(ariaLabelledBy, floorLabelId)}
+                aria-describedby={concatIds(
+                    ariaDescribedBy,
+                    floorInstructionId,
+                    liveMessageId
+                )}
+                aria-invalid={ariaInvalid}
                 placeholder={
                     currentFocus === "floor" && !readOnly
```
qroll: include the `disabled` check here as well

src/shared/accessibility/index.ts
tientranngoc-alt: Can i still put it here? cuz if i put it in `unit-number-input `as an arrow function, i use it for useMemo in liveMessage so it will throw error `Error: Cannot access 'formatPhraseWithPrefix' before initialization` . 
Cuz the liveMessage is declared first, at this time , arrow function `formatPhraseWithPrefix` has not been init

Or another solution i think is using function declaration for this :` function formatPhraseWithPrefix(){}`

Issue #812, opened 2025-09-09
Upgrade to ESLint v9


qroll:
- Upgrade to ESLint 9
- Switch to the new `eslint.config.js` format, maintaining the original rules as much as possible
- Clean up any deprecated rules
- Fix new linting errors introduced by the upgrade
- Ensure that VSCode correctly highlights out errors and warnings in the IDE
- Ensure that the linter still runs correctly during the pre-commit hook
  - should exit on errors
  - should update problems that are automatically fixable

PR #873, opened 2025-11-03
[#812][VL]Upgrade to ESLint v9

xjing97:
**Changes**
- Upgrade to ESLint 9
- [delete] branch

**Additional information**

- You may refer to this [ticket](https://github.com/LifeSG/react-design-system/issues/812)

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -4,7 +4,6 @@ import typescriptEslint from "@typescript-eslint/eslint-plugin";
 import globals from "globals";
 import tsParser from "@typescript-eslint/parser";
 import js from "@eslint/js";
-import * as eslintMdx from "eslint-mdx";
 import eslintConfigPrettier from "eslint-config-prettier/flat";
 import reactHooks from "eslint-plugin-react-hooks";
 import storybook from "eslint-plugin-storybook";
@@ -165,22 +164,6 @@ export default defineConfig([
             globals: {
                 ...globals.jest,
             },
-        },
-    },
-    {
-        files: ["**/*.mdx"],
-        languageOptions: {
-            parser: eslintMdx,
-            globals: {
-                React: "readonly",
-                JSX: "readonly",
-            },
-        },
-        settings: {
-            "mdx/code-blocks": true,
-        },
-        rules: {
-            semi: "off", // disable false "Missing semicolon" errors
         },
     },
     {


.vscode/settings.json
```diff
@@ -1,10 +1,21 @@
 {
+    "editor.codeActionsOnSave": {
+        "source.fixAll.eslint": true
+    },
     "editor.defaultFormatter": "esbenp.prettier-vscode",
     "editor.formatOnSave": true,
     "eslint.options": {
-        "extensions": [".md", ".mdx"]
+        "extensions": [".md", ".mdx"],
```
qroll: I found that the plugin threw an error when it encountered this `extension` option, to remove this

eslint.config.mjs
```diff
@@ -0,0 +1,164 @@
...
+        },
+    },
+    {
+        files: ["**/*.mdx"],
+        plugins: {
+            mdx,
+        },
+        settings: {
+            "mdx/code-blocks": true,
+        },
```
qroll: could you share what this setting does? how is it different from `mdx.flatCodeBlocks`?

qroll:
may also need to set up ignores to exclude the `node_modules` (not sure if this is a default already) and `dist`. not sure if `.storybook` is already ignored by the storybook plugin by default?

https://eslint.org/docs/latest/use/configure/ignore

eslint.config.mjs
xjing97: I found we need to configure the MDX parser manually for proper linting, I will remove mdx.flat and mdx.flatCodeBlocks.

custom-types/react-augment.d.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-undef */
```
qroll: instead of disabling the rule, are we able to update to use the React namespace?

```diff
+ ReactElement
- React.ReactElement
```

src/navbar/drawer.styles.tsx
```diff
@@ -55,6 +55,7 @@ export const Container = styled.nav<StyleProps>`
 
     ${(props) => VISIBILITY_STYLE(props.$show)}
     ${(props) => {
+        // eslint-disable-next-line no-constant-binary-expression
         const viewHeight = `${props.$viewHeight}px` || "1vh";
```
qroll: the logic here should be updated to check that `viewHeight` is truthy before using it

src/predictive-text-input/predictive-text-input.tsx
```diff
@@ -75,6 +75,7 @@ export const PredictiveTextInput = <T, V>({
             setOptions(fetchedOptions ?? []);
             setIsLoading(false);
         } catch (err) {
+            console.log(err);
```
qroll: omit catch binding

src/util/time-helper.ts
```diff
@@ -148,6 +148,7 @@ export namespace TimeHelper {
 
             return timeValues;
         } catch (error) {
+            console.log(error);
```
qroll: omit catch binding

eslint.config.mjs
```diff
@@ -0,0 +1,202 @@
...
+        ignores: [
+            "**/node_modules/**",
+            "**/dist/**",
+            "**/.storybook/**",
+            "**/e2e/**",
+            "playwright.config.ts",
+        ],
+    },
+    js.configs.recommended,
+    reactPlugin.configs.flat.recommended,
```
qroll: to also include the JSX runtime config recommended by `eslint-plugin-react` for React 17+

https://github.com/jsx-eslint/eslint-plugin-react?tab=readme-ov-file#flat-configs

stories/file-download/file-download.stories.tsx
```diff
@@ -237,6 +237,7 @@ const handleDemoDownload = async (file: FileItemDownloadProps) => {
             // Close the stream
             await writable.close();
         } catch (error) {
+            console.log(error);
```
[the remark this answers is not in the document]
xjing97: updated

src/util/use-event.ts
```diff
@@ -4,18 +4,16 @@ import { useCallback, useLayoutEffect, useRef } from "react";
  * Alternative to `useEffectEvent` which is only available in React 18+
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function useEvent<T extends (...args: any[]) => any>(callback: T): T {
+export function useEvent<T extends(...args: any[]) => any>(callback: T): T {
```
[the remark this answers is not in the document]
xjing97: updated

eslint.config.mjs
qroll: bumps!

qroll (approved):

PR #893, opened 2025-11-18
[MOL-19872][JC] hide i icon in form error message if error msg is empty space

gnzhen:
**Changes**
trim empty space in form error message

- delete branch


src/form/form-wrapper.tsx
qroll: this check still needs to be based on the original errorMessage value to correctly expose the error state via `aria-invalid`

qroll (approved):

PR #901, opened 2025-12-02
[BOOKINGSG-8406][NT] Updated time slot bar component

nobeltws:
**Changes**
Update timeslot bar component to display time markers for all time values

- [delete] branch

**Changelog entry**

- Update timeslot bar component to display time markers for all time values

**Additional information**

- You may refer to this [ticket]([BOOKINGSG-8406](https://sgtechstack.atlassian.net/browse/BOOKINGSG-8406))

### DIFF
--- a/src/time-slot-bar/helper.ts
+++ b/src/time-slot-bar/helper.ts
@@ -68,10 +68,12 @@ export namespace TimeSlotBarHelper {
                 }
                 // For xx:31 to xx:59, adjust to next hour xx:00
                 if (parsedTime.minute() > 30 && parsedTime.minute() < 60) {
-                    return parsedTime
-                        .add(1, "hour")
-                        .set("minute", 0)
-                        .format("HH:mm");
+                    const nextHour = parsedTime.add(1, "hour");
+                    // If it's 23:xx and rounds up to 24:00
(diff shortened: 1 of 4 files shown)


qroll:
this is existing behaviour, but I would like to check again: for something like 08:30am the leading hour is stripped, is this still the expectation?

Screenshot 2025-12-03 at 10 54 08 PM

qroll (changes requested):
could I trouble you to include some unit tests for the actual component?

some scenarios I can think of:
- time slot markers for 00:00 to 24:00
- time slot markers for 00:15 to 23:45
- time slot markers for 00:30 to 23:30
- time slot markers for 00:45 to 23:15

in each scenario, to assert that the text `12am` and `11pm` are visible

no need to test other behaviour like slot display and clicks for now, shall cover that next time

nobeltws:
> this is existing behaviour, but I would like to check again: for something like 08:30am the leading hour is stripped, is this still the expectation?
> 
> Screenshot 2025-12-03 at 10 54 08 PM

yeap we will keep the existing behaviour for the leading time marker

qroll (approved):
thanks for increasing the coverage in the unit tests! 🌟

PR #914, opened 2025-12-18
[MOL-19528][LP] Add numeric variation to Timeline component

linhpk12:
**Changes**
Description of changes...

- [keep | delete] branch

**Changelog entry**

**Additional information**


src/timeline/timeline.tsx
qroll: could you create a unit test to cover the numeric variant? it should assert that the numeric count increases as expected

PR #915, opened 2025-12-18
[MISC][XY] Prevent Drawer from closing during text selection

ninjasheeps:
**Changes**
Fixed issue where drawer would auto-close when users highlight text inside the drawer and drag their mouse outside of drawer

Changes:
- Added `handleDocumentMouseDown` listener to track interaction origin
- Updated `handleWrapperClick` to close drawer if both mousedown AND click happened outside of drawer

- delete branch

### DIFF
--- a/src/modal-v2/modal-v2.tsx
+++ b/src/modal-v2/modal-v2.tsx
@@ -8,6 +8,7 @@ import {
 import React, { useEffect, useRef } from "react";
 import { Overlay } from "../overlay/overlay";
 import { useViewport } from "../shared/hooks";
+import { useEvent } from "../util";
 import { ModalContext } from "./modal-context";
 import { Container, ModalContainer, ScrollContainer } from "./modal-v2.styles";
 import { ModalV2Props } from "./types";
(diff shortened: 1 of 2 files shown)


src/overlay/overlay.tsx
```diff
@@ -236,14 +250,40 @@ const OverlayComponent = ({
...
+        // Track if mousedown started inside modal content
+        mouseDownInsideModalRef.current =
+            clickedInsideContainer && !clickedOnContainer;
+    };
+
+    const handleWrapperClick = (event: React.MouseEvent<HTMLDivElement>) => {
+        const wrapper = event.currentTarget;
+        const target = event.target as Node;
+        const clickedInsideModal =
+            wrapper !== target && wrapper.contains(target);
```
qroll: I think the logic here also needs to check against the `containerRef`/`wrapperRef` (as currentTarget is always pointing to the overlay container)

could we implement the `containerRef` with Modal and ModalV2?

src/modal-v2/slots/card.tsx
```diff
@@ -56,6 +60,8 @@ export const Card = ({
             {hasCloseButton && CloseButtonSlot}
         </ModalCard>
     );
-};
+}
 
-Card.displayName = "ModalV2.Card";
+export const Card = React.forwardRef<HTMLDivElement, ModalCardProps>(CardInner);
+
+(Card as any).displayName = "ModalV2.Card";
```
qroll:
```suggestion
(Card as NamedExoticComponent).displayName = "ModalV2.Card";
```

src/overlay/overlay.tsx
```diff
@@ -21,6 +21,8 @@ const OverlayComponent = ({
     enableOverlayClick = false,
     zIndex: customZIndex,
     id,
+    containerRef,
+    childWithRef,
```
qroll: I think `childWithRef` is not needed as a prop? containerRef is meant to represent the overlay's visual child container already

ninjasheeps:
yep, this makes sense to me. thank you so much for working on this!!

PR #917, opened 2025-12-19
Fix A11Y Issue: select fields not detecting labels

hermanloh-gt:
**Changes**
- added "aria-labelledby" in FormWrapper renderChildren fn
- pass aria attributes to ExpandableElement props through select fields
- update types of select components

**Additional Information**
- components affected:
  - input-select
  - input-multi-select
  - input-nested-multi-select
  - input-nested-select
  - select-histogram

### DIFF
--- a/src/form/form-multi-select.tsx
+++ b/src/form/form-multi-select.tsx
@@ -32,6 +32,7 @@ export const FormMultiSelect = <T, V>({
     return (
         <FormWrapper
             id={inputId}
+            data-testid={testId}
             label={label}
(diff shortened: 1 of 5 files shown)


src/form/form-multi-select.tsx
qroll: also using `inputId` here so that it's correctly associated with the label and updating `data-testid`

```suggestion
id={`${inputId}-base`}
data-testid={testId ? `${testId}-base` : undefined}
```

PR #920, opened 2025-12-23
Update error handling in pipeline script

qroll:
**Changes**

- job runs infinitely when error is encountered, updated the script to exit when the curl fails on 401
- adjust timings
- remove unintended sleep when job status is complete
- [delete] branch


chongwyb (approved):
approved the validation of token if invalid

PR #927, opened 2026-01-05
[BOOKINGSG-8604][LS] Fix combobox picker drop down content reducing to zero on smaller viewports

lionel-sim:
**Changes**
- Fix invalid css width values for menu content, select histogram, and combobox picker
- Add missing `matchElementWidth` for combobox picker, which causes dropdown list to not render in smaller viewports

- [delete] branch

### DIFF
--- a/src/menu/menu-content.styles.tsx
+++ b/src/menu/menu-content.styles.tsx
@@ -25,21 +25,24 @@ export const MenuPanel = styled.div<MenuPanelStylesProps>`
     border: ${Border["width-010"]} ${Border["solid"]} ${Colour["border"]};
     background: ${Colour["bg"]};
     box-shadow: ${Shadow["md-subtle"]};
-    min-width: 15rem;
-    max-width: 24rem;
+
+    --x-spacing: 0px;
+    --available-width: calc(100vw - var(--x-spacing) * 2);
 
     ${MediaQuery.MaxWidth.sm} {
-        max-width: calc(100vw - ${Breakpoint["sm-margin"]}px * 2);
-        min-width: unset;
+        --x-spacing: ${Breakpoint["sm-margin"]}px;
     }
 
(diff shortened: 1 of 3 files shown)


src/time-range-picker/combobox-picker/combobox-picker.tsx
```diff
@@ -311,6 +311,7 @@ export const ComboboxPicker = ({
                         startOptions
                     )}
                     listboxId={internalId}
+                    matchElementWidth={true}
```
qroll: nit: prefer shorthand for boolean props

```suggestion
                    matchElementWidth
```

src/menu/menu-content.styles.tsx
qroll: realised that min-width takes precedence so it still overflows on a small screen with a large browser font even with max-width. so we also need to constrain min-width


i think we could do something like this to simplify the media queries. basically to always take the preferred value unless it going to be wider than the screen

```
    --x-spacing: 0px;
    min-width: min(15rem, calc(100vw - var(--x-spacing) * 2));
    max-width: min(24rem, calc(100vw - var(--x-spacing) * 2));

    ${MediaQuery.MaxWidth.sm} {
        --x-spacing: ${Breakpoint["sm-margin"]}px;
    }

    ${MediaQuery.MaxWidth.xs} {
        --x-spacing: ${Breakpoint["xs-margin"]}px;
    }

    ${MediaQuery.MaxWidth.xxs} {
        --x-spacing: ${Breakpoint["xxs-margin"]}px;
    }
```

src/shared/dropdown-list-v2/dropdown-list.styles.tsx
qroll: similar for dropdown (which is currently broken on mobile 🫢). should be able to use the same technique for `min-width`

PR #933, opened 2026-01-15
[os-font-specs][TIEN] add new OneService font specs

tientranngoc-alt:
**Changes**

- Create new font specs for `OneService` theme

### DIFF
--- a/src/theme/font/specs/oneservice-font-set.ts
+++ b/src/theme/font/specs/oneservice-font-set.ts
@@ -1,6 +1,4 @@
 import { FontSet } from "../types";
 import { getFontSet } from "../utils";
 
-export const OneServiceFontSet: FontSet = getFontSet({
-    disableLigatures: true,
-});
+export const OneServiceFontSet: FontSet = getFontSet();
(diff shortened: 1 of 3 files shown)


src/theme/font/specs/oneservice-font-set.ts
```diff
@@ -0,0 +1,6 @@
+import { FontSet } from "../types";
+import { getFontSet } from "../utils";
+
+export const OneServiceFontSet: FontSet = getFontSet({
+    disableLigatures: true,
```
qroll: can I confirm if ligatures should follow the Open Sans default which does not disable ot? or it's meant to be explicitly disabled?

qroll (changes requested):

src/theme/font/specs/oneservice-font-set.ts
tientranngoc-alt: I’ve just checked with UX, and this should follow the default behavior (not disable).

qroll (approved):

Issue #736, opened 2025-07-16
Improve keyboard navigation for `Navbar` submenu


qroll:
**Component**
`Navbar`

**Describe the feature request**
- Migrate submenu to use `Menu` component
- Add keyboard navigation

**Acceptance criteria**
- Support hover, focus interaction for menu items
- Support arrow keys to navigate between submenu items
- Tab navigation order should make sense
- Minimally test in NVDA + Chrome (Windows), VO + Safari (macOS), Talkback + Chrome (Android)

**References**
[Figma annotations](https://www.figma.com/design/wDAxbRmlq4xOn4UJ8zPuEF/%F0%9F%A7%91%F0%9F%8F%BB%E2%80%8D%F0%9F%A6%AF-Flagship-A11y-Annotations?node-id=1800-2226&p=f&m=dev)

qroll:
As a follow up to the `Avatar` and `Badge` enhancements, proposing a new `Menu` component to handle focus, positioning, keyboard navigation and accessibility. This will be similar to `PopoverV2` which has a `PopoverTrigger`.

It will also feature a composition api instead of configuration
```
<Menu>
  <Menu.Section>
    <Menu.Title />
    <Menu.Item />
  </Menu.Section>
</Menu>
```

PR #943, opened 2026-01-29
Improve keyboard navigation for Navbar submenu #736

benjaminLeongSK:
**Changes**
Description of changes...

- Migrating to new menu component for navbar
- Added tab and arrow navigation

### DIFF
--- a/src/navbar/navbar-items.styles.tsx
+++ b/src/navbar/navbar-items.styles.tsx
@@ -129,9 +129,9 @@ export const LinkLabel = styled.div`
 export const LinkIndicator = styled.div<StyleProps>`
     position: absolute;
     bottom: 0;
+    left: 0;
+    right: 0;
     height: 0.25rem;
-    left: 0.5rem;
-    right: 0.5rem;
     background-color: ${Colour["border-selected"]};
 
     &:hover {


src/popover-v2/popover-trigger.tsx
```diff
@@ -156,19 +211,30 @@ export const PopoverTrigger = ({
...
             >
                 {children}
             </TriggerContainer>
             {visible && (
                 <FloatingPortal root={rootNode}>
-                    <FloatingFocusManager context={context}>
+                    <FloatingFocusManager
+                        context={context}
+                        initialFocus={-1}
+                        returnFocus={false}
```
qroll: same concerns for this. should we only set this for non-modal usage?

src/navbar/navbar-items.styles.tsx
```diff
@@ -70,17 +70,26 @@ export const LinkItem = styled.li<ItemStyleProps>`
 `;
 
 export const Link = styled(Typography.LinkMD)<StyleProps>`
+    background: none;
+    border: none;
+    margin: 0;
+    box-shadow: none;
+    font: inherit;
+    color: inherit;
+    text-align: inherit;
```
qroll: what are these additional styles for?

src/navbar/navbar-items.tsx
```diff
@@ -58,59 +84,153 @@ export const NavbarItems = <T,>({
...
-    const renderItems = () => {
-        return items.map((item, index) => {
+    const renderSubMenu = (
+        subMenu: NavItemCommonProps<T>[],
+        parentIndex: number
+    ) => (
+        <Menu.Content>
+            <Menu.Section
+                showDivider={false}
+                onKeyDownCapture={(e: React.KeyboardEvent<HTMLElement>) => {
```
qroll: why `onKeyDownCapture` rather than `onKeydown`?

src/navbar/navbar-action-buttons.tsx
```diff
@@ -41,17 +41,14 @@ export const NavbarActionButtons = ({
...
-         * to not be run
-         */
+        // prevent overlay from intercepting
         event.stopPropagation();
     };
 
     const handleActionButtonClick = (actionButton: NavbarButtonProps) => {
         return (event: React.MouseEvent<HTMLButtonElement>) => {
-            event.stopPropagation(); // in mobile, this prevents the drawer from intercepting event
+            event.stopPropagation();
```
qroll: restore the comment

src/popover-v2/popover-trigger.tsx
```diff
@@ -156,19 +211,30 @@ export const PopoverTrigger = ({
...
                 {...otherProps}
             >
                 {children}
             </TriggerContainer>
             {visible && (
                 <FloatingPortal root={rootNode}>
-                    <FloatingFocusManager context={context}>
+                    <FloatingFocusManager
+                        context={context}
+                        initialFocus={-1}
```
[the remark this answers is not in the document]
benjaminLeongSK: hmm i put this here so that the first tab of the navmenu will focus on the navbar item first before focusing on the submenu items under the navbar item. without this the focus will go straight to the submenu items and skip the navbar item

src/popover-v2/popover-trigger.tsx
```diff
@@ -156,19 +211,30 @@ export const PopoverTrigger = ({
...
+                        returnFocus={false}
+                        modal={isModal}
+                    >
                         <div
                             ref={(node) => {
                                 popoverRef.current = node;
                                 refs.setFloating(node);
                             }}
+                            onBlurCapture={handleBlurCapture}
```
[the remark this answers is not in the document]
benjaminLeongSK: my thought is that onBlurCapture is more precise in a sense that it can detect focus leaving the menu before focus moves elsewhere whereas onBlur runs after focus has already moved. but i have since tested with onBlur and everything seems to be ok. will revert to onBlur instead!

src/navbar/navbar-items.tsx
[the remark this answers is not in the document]
qroll: sorry should be `Menu.Content` instead since it's the parent

src/navbar/menu.styles.tsx
```diff
@@ -1,80 +1,45 @@
...
+import { Border, Colour, Radius, Spacing } from "../theme";
 import { Typography } from "../typography";
 
-// =============================================================================
-// STYLE INTERFACE, transient props are denoted with $
-// See more https://styled-components.com/docs/api#transient-props
-// =============================================================================
+/* =============================================================================
+   MOBILE MENU WRAPPER
+============================================================================= */
```
qroll: since this file is much smaller now, we can omit these headers since the names are quite self explanatory

src/navbar/navbar-items.tsx
```diff
@@ -93,24 +115,75 @@ export const NavbarItems = <T,>({
...
+        items.map((item, index) => {
             switch (item.itemType) {
                 case "component": {
-                    const component = (item && item.children) || null;
-                    return <li key={index}>{component}</li>;
+                    return <li key={index}>{item.children ?? null}</li>;
                 }
+
                 case "link":
                 default: {
```
qroll: I feel like this switch case handling has become very long and complex, could this be moved into a separate render function?

src/navbar/navbar-items.tsx
```diff
@@ -121,70 +194,152 @@ export const NavbarItems = <T,>({
...
+                                </LinkIconContainer>
+                            )}
+                        </Link>
+                    );
+
+                    return (
+                        <LinkItem
+                            key={index}
+                            $hiddenBranding={hideNavBranding}
+                            onFocusCapture={() => setFocusedIndex(index)}
```
qroll: does the indicator need to move with focus? or only when there is an active submenu?

benjaminLeongSK: hmmm currently it shows when the focus moves as well for non submenu navbar items

src/navbar/navbar-items.tsx
```diff
@@ -58,133 +111,200 @@ export const NavbarItems = <T,>({
...
+        }
+        const subMenuId = subMenuIdsRef.current[index] ?? undefined;
+
+        const isRouteSelected = checkSelected(item);
+        const isOpen = isSubMenuTrigger && openSubMenuIndex === index;
+
+        const selected =
+            activeDesktopSubmenuOwnerIndex !== null
+                ? activeDesktopSubmenuOwnerIndex === index
+                : isRouteSelected;
```
qroll: what's the different between `activeDesktopSubmenuOwnerIndex` and `openSubMenuIndex` and could those be combined?

src/navbar/navbar-items.tsx
```diff
@@ -58,133 +111,200 @@ export const NavbarItems = <T,>({
...
+                data-testid={testId}
+                weight={textWeight}
+                $selected={selected}
+                underlineStyle="none"
+                {...otherItemAttrs}
+                aria-current={
+                    !isSubMenuTrigger && isRouteSelected ? "page" : undefined
+                }
+                aria-haspopup={isSubMenuTrigger ? "menu" : undefined}
+                aria-expanded={isSubMenuTrigger ? isOpen : undefined}
```
qroll: it should be true when submenu is expanded on mobile too

I think at this point links vs links with submenu are too complicated to keep together, I think we can further split into:
- link only
- link with submenu

ok with a bit of duplication if it makes things more readable

src/navbar/navbar-items.tsx
```diff
@@ -58,134 +88,224 @@ export const NavbarItems = <T,>({
...
+                aria-current={isRouteSelected ? "page" : undefined}
+                onClick={handleLinkClick(item, index)}
+                {...options}
+            >
+                <LinkLabel>{children}</LinkLabel>
+                {renderIndicator()}
+            </Link>
+        );
+
+        const renderLinkWithSubmenu = () => {
```
qroll: can reorder this function so only variables are declared when used

also don't think there's a need to instantiate mobileTrigger and desktopTrigger, they can just be declared inline

```
if (mobile) {
  const mobileTrigger 
  return <>
     <Link...>
  </>
}

return <DesktopMenu>
   <LinkButton...>
</DesktopMenu>
```

Issue #878, opened 2025-11-06
Migrate TimePicker to Floating UI


qroll:
**Component**
`TimePicker`

**Describe the feature request**
Part of #443 for improved dropdown display and positioning

- Migrate TimePicker to use the shared FloatingUI dropdown utility
- Update the AM/PM button styling to match Figma
- Expose the `alignment`, `dropdownZIndex`, and `dropdownRootNode` props and document them in Storybook
- Create unit tests to assert the selection and blur/focus behaviour

**Acceptance criteria**
- Component follows the DS styling in Figma
- Dropdown auto-positions when the user scrolls on the page
- Dropdown should be reasonably keyboard navigable
- Unit tests should pass

**References**
[TimePicker Figma](https://www.figma.com/design/tHmHQNszVzvIamtZWixjgR/branch/pc8hwikwKdpjhh0VROF3r6/%F0%9F%8C%B1-Flagship-V3-Component-Kit?node-id=3088-4379&m=dev)

PR #952, opened 2026-02-05
Migrate TimePicker to Floating UI #878

benjaminLeongSK:
**Changes**
- Migrate time picker to use shared dropdown
- Added new props
- Added unit test
- Changed design

### DIFF
--- a/src/timepicker/timepicker.tsx
+++ b/src/timepicker/timepicker.tsx
@@ -60,16 +60,30 @@ export const Timepicker = ({
     const handleClose = () => {
         setIsOpen(false);
         onBlur?.();
+    };
+
+    const handleDismiss = () => {
+        setIsOpen(false);
         restoreFocusToSelector();
     };
 
     const handleChange = (v: string) => {
         onChange?.(v);
-        handleClose();
+        setIsOpen(false);
+        restoreFocusToSelector();
     };
 
     const handleFocus = () => {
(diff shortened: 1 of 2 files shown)


src/shared/timepicker-dropdown/timepicker-dropdown.styles.tsx
```diff
@@ -79,12 +74,7 @@ export const TimePeriodSection = styled.div`
     display: flex;
     gap: ${Spacing["spacing-8"]};
 
-    ${MediaQuery.MaxWidth.lg} {
-        flex-direction: column;
-    }
-
```
qroll: why was this removed?

src/timepicker/timepicker.tsx
```diff
@@ -17,116 +20,106 @@ export const Timepicker = ({
...
             />
         </InputWrapper>
     );
+
+    const renderDropdown = ({
+        styles,
+        setFloatingRef,
+        getFloatingProps,
+    }: DropdownRenderProps) => {
+        if (!isOpen) return null;
```
qroll: this is probably not needed since ElementWithDropdown already handles the visibility?

qroll (approved):
some minor things left but can be addressed in the a11y PR:
- remove `fitAvailableHeight` (don't want the panel to resize)
- disabled field should be focusable

Issue #880, opened 2025-11-06
Migrate TimeRangePicker (dial) to Floating UI


qroll:
**Component**
`TimeRangePicker` dial variant

**Describe the feature request**
Part of #443 for improved dropdown display and positioning

- Migrate TimeRangePicker (dial variant) to use the shared FloatingUI dropdown utility
- Expose the `alignment`, `dropdownZIndex`, and `dropdownRootNode` props and document them in Storybook
- Create unit tests to assert the selection and blur/focus behaviour

**Acceptance criteria**
- Component follows the DS styling in Figma
- Dropdown auto-positions when the user scrolls on the page
- Unit tests should pass

**References**
[TimeRangePicker Figma](https://www.figma.com/design/tHmHQNszVzvIamtZWixjgR/branch/pc8hwikwKdpjhh0VROF3r6/%F0%9F%8C%B1-Flagship-V3-Component-Kit?node-id=3088-4553&m=dev)

PR #958, opened 2026-02-12
Migrate TimeRangePicker (dial) to Floating UI #880

benjaminLeongSK:
**Changes**

Migrate time picker dial version to use shared dropdown
Added new props
Added unit test

### DIFF
--- a/tests/time-range-picker/time-range-picker.spec.tsx
+++ b/tests/time-range-picker/time-range-picker.spec.tsx
@@ -647,11 +647,15 @@ describe("TimeRangePicker", () => {
             expect(getTimepickerDropdown()).not.toBeInTheDocument();
         });
 
-        it("should close dropdown and call onBlur when Cancel is clicked", async () => {
+        it("should close dropdown and keep focus when Cancel is clicked", async () => {
             const user = userEvent.setup();
             const mockOnBlur = jest.fn();
 
-            render(<TimeRangePicker onBlur={mockOnBlur} />);
+            render(
+                <>
+                    <TimeRangePicker onBlur={mockOnBlur} />
+                </>
+            );
 
             await user.click(getFrom());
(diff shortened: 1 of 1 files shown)


src/time-range-picker/dial-picker/dial-picker.tsx
```diff
@@ -16,19 +19,40 @@ export const DialPicker = ({
...
+        if (!isOpen) onFocus?.();
+        setIsOpen(true);
+    };
 
-    const nodeRef = useRef<HTMLDivElement>(null);
+    const close = () => {
+        setIsOpen(false);
+        setActive("none");
+        onBlur?.();
+    };
```
qroll: move to appropriate event handlers or helper function section

stories/form/form-time-range-picker/props-table.tsx
qroll: we can also expose `dropdownRootNode` on the combobox variant, and move these three dropdown props to the common section

qroll (changes requested):

tests/time-range-picker/time-range-picker.spec.tsx
```diff
@@ -566,4 +566,148 @@ describe("TimeRangePicker", () => {
...
+            await user.click(getFrom());
+            await waitFor(() => expect(getTimepickerDropdown()).toBeVisible());
+
+            await user.keyboard("{Escape}");
+
+            await waitForElementToBeRemoved(() => getTimepickerDropdown());
+
+            expect(mockOnBlur).toHaveBeenCalledTimes(0);
+
+            expect(screen.getByTestId("timepicker-container")).toHaveFocus();
```
qroll: can we add a step after this to click outside and assert that onBlur was called once?

qroll (approved):

PR #979, opened 2026-03-04
[BOOKINGSG-8990][GZ] useApplyStyles hooks

ghazwan-gt:
**Type of changes**

**Description of changes**

Implementation of [BOOKINGSG-8990 Create useApplyStyle hook](https://sgtechstack.atlassian.net/browse/BOOKINGSG-8990)
- Create a useApplyStyles hooks

**Checklist**

-   [ ] Updated documentation
-   [X ] Added/updated tests

**Screenshots** 

> [!NOTE]
> No SS

### DIFF
--- a/src/theme/utils/use-apply-styles.ts
+++ b/src/theme/utils/use-apply-styles.ts
@@ -1,25 +1,20 @@
 import { RefObject } from "react";
 import { useIsomorphicLayoutEffect } from "../../util";
 
-type StyleValue<TProps> =
-    | string
-    | number
-    | ((props: TProps) => string)
-    | null
-    | undefined;
+type StyleValue = string | number | null | undefined;
 
+// TODO: Update to FDSVariableName once 9839 merged
 type CSSVariableKey = `--${string}`;
 
-export type ApplyStyleMap<TProps> = {
-    [K in keyof React.CSSProperties]?: StyleValue<TProps>;
+export type ApplyStyleMap = {
+    [K in keyof React.CSSProperties]?: StyleValue;
 } & {
-    [key in CSSVariableKey]?: StyleValue<TProps>;
+    [key in CSSVariableKey]?: StyleValue;
 };
 
-export function useApplyStyle<TElement extends HTMLElement, TProps>(
+export function useApplyStyle<TElement extends HTMLElement>(
     ref: RefObject<TElement>,
-    styles: ApplyStyleMap<TProps>,
-    props: TProps
+    styles: ApplyStyleMap
 ) {
     useIsomorphicLayoutEffect(() => {
         const element = ref?.current;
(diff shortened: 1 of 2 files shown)


src/theme/utils/use-apply-styles.ts
qroll: introduce a `src/theme/utils/index.ts`

tests/theme/utils.spec.tsx
```diff
@@ -0,0 +1,135 @@
...
+        const element = getByTestId("target");
+
+        expect(element.style.background).toBe("red");
+
+        act(() => {
+            rerender(
+                <TestComponent styles={{ background: null }} props={{}} />
+            );
+        });
+        screen.debug(element);
```
zhaoyanxzy: should be removed if not needed

ghazwan-gt: Will do

src/theme/utils/use-apply-styles.ts
```diff
@@ -0,0 +1,40 @@
...
+export type ApplyStyleMap<TProps> = {
+    [K in keyof React.CSSProperties]?: StyleValue<TProps>;
+} & {
+    [key in CSSVariableKey]?: StyleValue<TProps>;
+};
+
+export function useApplyStyle<TElement extends HTMLElement, TProps>(
+    ref: RefObject<TElement>,
+    styles: ApplyStyleMap<TProps>,
+    props: TProps
```
qroll: can share more on how you envision props to be used?

for cases where theme/mode have to be passed down, what does the usage look like?

ghazwan-gt: I was initially mirroring the V3 pattern where styled-components accepted functions with props.

But thinking about it now, I don’t think we actually need that in V4. Even if a style depends on component props dynamically, we can just resolve it before calling the hook like:

```
background: status === "error"
  ? Colour.Primitive["error-50"]
  : Colour.Primitive["neutral-30"]
```

We don't need to specify the theme or mode directly anymore, since both are appliad via `data` attributes and the css variables resolves automatically. So having `Colour.Primitive["error-50"]` already works for any theme and mode. So, the JS side does not need to know theme or mode for styling.

With that approach, we could simplify the hook and remove function support entirely, reducing the type to:

`type StyleValue = string | number | null | undefined;`

That would keep the hook simpler and avoid carrying over patterns from styled-components that we no longer need.

src/theme/utils/use-apply-styles.ts
```diff
@@ -0,0 +1,32 @@
+import { RefObject } from "react";
+import { useIsomorphicLayoutEffect } from "../../util";
+
+type StyleValue = string | number | null | undefined;
+
+// TODO: Update to FDSVariableName once 9839 merged
```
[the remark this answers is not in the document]
ghazwan-gt: Yup, that's what I have in mind. The type is flexible yet we still got the suggestions for our tokens

PR #980, opened 2026-03-04
Bl/884

benjaminLeongSK:
Changes:

- readonly field becomes a button
- Handle masking behaviour
- Add aria labels

### DIFF
--- a/src/masked-input/masked-input.tsx
+++ b/src/masked-input/masked-input.tsx
@@ -131,10 +131,6 @@ const Component = ({
             inputRef.current?.focus();
             return;
         }
-
-        if (document.activeElement === inputRef.current) {
-            inputRef.current?.blur();
-        }
     };
 
     // =============================================================================


src/masked-input/masked-input.tsx
```diff
@@ -51,6 +54,15 @@ const Component = (
     const isEmptyReadOnlyState = readOnly && isEmpty(value);
     const [isMasked, setIsMasked] = useState<boolean>(!disableMask);
     const [updatedValue, setUpdatedValue] = useState<string>(value || "");
+    const valueId = `${otherProps.id ?? dataTestId ?? "masked-input"}-value`;
```
qroll: if we're using this for aria association, the id needs to be guaranteed unique. data testid is not suitable here

generate an internal id using `SimpleIdGenerator`

src/masked-input/masked-input.tsx
```diff
@@ -99,19 +122,35 @@ const Component = (
     };
 
     const handleToggleMask = () => {
```
qroll: this is behaving strangely when clicking on the corner, the masking should be toggled instead of focusing the input

https://github.com/user-attachments/assets/2d6ec35c-df79-4305-af9f-e1ea0bba278b

PR #986, opened 2026-03-05
RangeSelect accessibility #877

benjaminLeongSK:
Changes

- make individual range selector tab-able
- added accessibility


package-lock.json
qroll: to revert

PR #991, opened 2026-03-09
[BOOKINGSG-8907] Prompt for component migration

qroll:
**Description of changes**

-   Link to [ticket](https://sgtechstack.atlassian.net/browse/BOOKINGSG-8907)
-   Added prompts for E2E test generation and V4 component migration. They are separated because generating E2E tests seems to take up a lot of the context window, causing the agent to lose track of earlier instructions (in particular, the one that says to pause for commits before continuing)
-   Fix styling in sample NextJS app
- Refer to this [branch](https://github.com/LifeSG/react-design-system/tree/BOOKINGSG-8907-example) on some of the outputs

### DIFF
--- a/.github/prompts/migrate-v4-component.prompt.md
+++ b/.github/prompts/migrate-v4-component.prompt.md
@@ -25,14 +25,19 @@ Before you begin, create the outline for a step-by-step plan you would have to t
 
 ## Phases
 
-### 1. Migrate V3 design tokens to V4 design tokens
+### 1. Rename style file extension
+
+-   Rename `<component>.style.tsx` or `<component>.styles.tsx` to `<component>.style.ts` to remain consistent with the project conventions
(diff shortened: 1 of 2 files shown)


.github/prompts/migrate-v4-component.prompt.md
```diff
@@ -0,0 +1,135 @@
...
+    `
+
+    <div className={clsx(styles.container, disabled && styles.containerDisabled, className)} />
+    ```
+
+-   If the `css` tag is empty, it can be removed
+
+### 4. Update props documentation in Storybook
+
+If the component is not a top-level component, skip this step.
```
ghazwan-gt: Should we clarify what `top-level component` means?

qroll: so far the AI has been pretty consistent; it's able to detect that the items in the `src/shared` folder are not public

PR #998, opened 2026-03-10
TimePicker accessibility #879

benjaminLeongSK:
Changes: 

- added accessibility features

### DIFF
--- a/src/form/form-timepicker.tsx
+++ b/src/form/form-timepicker.tsx
@@ -30,7 +30,7 @@ export const FormTimepicker = ({
 
     return (
         <FormWrapper
-            id={id}
+            id={inputId}
             label={label}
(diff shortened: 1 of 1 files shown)


src/timepicker/timepicker.tsx
```diff
@@ -130,7 +138,13 @@ export const Timepicker = ({
         setFloatingRef,
         getFloatingProps,
     }: DropdownRenderProps) => (
-        <div ref={setFloatingRef} style={styles} {...getFloatingProps()}>
+        <div
```
qroll: the reason why the "increase hour" text is read out is because the modal does not have a proper title using `aria-label`

for this component, use `Time picker` as the label

qroll (changes requested):

Issue #885, opened 2025-11-06
RangeSlider accessibility


qroll:
**Component**
`RangeSlider`

**Describe the feature request**
- Align the RangeSlider with the accessibility behaviour defined in the Figma specs

**Acceptance criteria**
- Roles should be correctly read out by the AT
- Label and error should be correctly associated with the input field
- Disabled, readonly state should be correctly read out by the AT
- Should remain focusable when disabled
- Minimally test in NVDA + Chrome (Windows), VO + Safari (macOS), Talkback + Chrome (Android)

**References**
[Figma](https://www.figma.com/design/wDAxbRmlq4xOn4UJ8zPuEF/%F0%9F%A7%91%F0%9F%8F%BB%E2%80%8D%F0%9F%A6%AF-Flagship-A11y-Annotations?node-id=558-2232&p=f&m=dev)

PR #999, opened 2026-03-10
RangeSlider accessibility #885

benjaminLeongSK:
Changes: 

- Added accessibility features for range slider behaviours

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -178,7 +178,6 @@
             "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
             "dev": true,
             "license": "MIT",
-            "peer": true,
             "dependencies": {
                 "@babel/code-frame": "^7.29.0",
                 "@babel/generator": "^7.29.0",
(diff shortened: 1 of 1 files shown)


src/input-range-slider/input-range-slider.tsx
```diff
@@ -53,22 +63,31 @@ export const InputRangeSlider = ({
...
+            return;
+        }
+
         if (typeof value === "number") {
-            const val = [value];
-            setSelection(val);
-            onChange?.(val);
-        } else {
-            if (index > -1 && selection[index] === value[index]) {
-                // skip unnecessary update when dragging the start thumb across the end thumb
```
qroll: restore this comment

src/input-range-slider/types.ts
```diff
@@ -21,6 +21,7 @@ export interface BaseSliderProps {
     indicatorLabelPrefix?: string | undefined;
     indicatorLabelSuffix?: string | undefined;
     renderSliderLabel?: ((value: number) => React.ReactNode) | undefined;
+    getAriaValueText?: ((value: number, index: number) => string) | undefined;
```
qroll: not sure what this is for? the component already has a `ariaLabels` prop

src/input-range-slider/types.ts
```diff
@@ -29,6 +29,8 @@ export interface InputRangeSliderProps extends BaseSliderProps {
     numOfThumbs?: number | undefined;
     /** The minimum difference between values */
     minRange?: number | undefined;
+    ariaErrorMessage?: string | undefined;
+    ariaInvalid?: boolean | undefined;
```
qroll: follow the native convention which is to use kebab case `aria-errormessage` and `aria-invalid`

src/input-range-slider/input-range-slider.tsx
```diff
@@ -116,15 +179,100 @@ export const InputRangeSlider = ({
...
     }
 
+    function getValueText(currentValue: number) {
+        return `${sliderLabelPrefix || ""}${currentValue}${
+            sliderLabelSuffix || ""
+        }`;
+    }
+
+    function getThumbDescriptionIds() {
+        return concatIds(
```
qroll: should also be expecting the form label subtitle which is passed as `aria-describedby`

Issue #888, opened 2025-11-06
FullscreenImageCarousel accessibility


qroll:
**Component**
`FullscreenImageCarousel`

**Describe the feature request**
- Align the FullscreenImageCarousel with the accessibility behaviour defined in the Figma specs

**Acceptance criteria**
- Roles should be correctly read out by the AT
- Focus should be trapped when the carousel overlay is active
- Can be navigated with the keyboard
- Minimally test in NVDA + Chrome (Windows), VO + Safari (macOS), Talkback + Chrome (Android)

**References**
[Figma](https://www.figma.com/design/wDAxbRmlq4xOn4UJ8zPuEF/%F0%9F%A7%91%F0%9F%8F%BB%E2%80%8D%F0%9F%A6%AF-Flagship-A11y-Annotations?node-id=2951-780&p=f&m=dev)

PR #1005, opened 2026-03-11
FullscreenImageCarousel accessibility #888

benjaminLeongSK:
Changes: 

- switched to use ModalV2
- added accessibilities features

### DIFF
--- a/src/fullscreen-image-carousel/fullscreen-image-carousel.tsx
+++ b/src/fullscreen-image-carousel/fullscreen-image-carousel.tsx
@@ -48,7 +48,7 @@ import {
     FullscreenImageCarouselRef,
     ImageDimension,
 } from "./types";
-import { useStateCallback } from "../shared/hooks/useStateCallback";
+import { useStateCallback } from "../shared/hooks";
 
 export const Component = (
     {
--- a/src/shared/hooks/index.tsx
+++ b/src/shared/hooks/index.tsx
@@ -1,1 +1,2 @@
 export * from "./useViewport";
+export * from "./use-state-callback";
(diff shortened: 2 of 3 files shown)


src/modal-v2/modal-v2.tsx
```diff
@@ -23,7 +23,13 @@ export const ModalV2 = ({
     zIndex,
     onOverlayClick,
     dismissKeyboardOnShow = true,
+    initialFocus,
+    fullscreen = false,
     "data-testid": testId = "modal",
+    role = "dialog",
```
qroll: why is there a need to expose role?

src/modal-v2/modal-v2.tsx
```diff
@@ -81,21 +87,25 @@ export const ModalV2 = ({
...
-                {...otherProps}
                 data-status={status}
+                {...otherProps}
             >
                 <ModalContext.Provider value={{ onClose }}>
                     {isMounted && (
                         <FloatingFocusManager
                             context={context}
-                            initialFocus={refs.floating}
+                            initialFocus={initialFocus ?? refs.floating}
```
qroll: hmm is it possible for the carousel to do it's own focus in a useEffect? trying to avoid exposing too much in ModalV2

src/modal-v2/modal-v2.tsx
```diff
@@ -24,6 +24,10 @@ export const ModalV2 = ({
     onOverlayClick,
     dismissKeyboardOnShow = true,
     "data-testid": testId = "modal",
+    "aria-label": ariaLabel,
+    "aria-labelledby": ariaLabelledBy,
+    "aria-describedby": ariaDescribedBy,
+    disableInitialFocus = true,
```
qroll: should default to false as that is the original behaviour

qroll (changes requested):

Issue #881, opened 2025-11-06
TimeRangePicker accessibility


qroll:
**Component**
`TimeRangePicker` dial and combobox variants

**Describe the feature request**
- Align TimeRangePicker with the accessibility behaviour defined in the Figma specs

**Acceptance criteria**
- Roles should be correctly read out by the AT
- Label and error should be correctly associated with the input field
- Disabled, readonly state should be correctly read out by the AT
- Can be navigated with the keyboard
- Should remain focusable when disabled
- Minimally test in NVDA + Chrome (Windows), VO + Safari (macOS), Talkback + Chrome (Android)

**References**
[Dial variant Figma](https://www.figma.com/design/wDAxbRmlq4xOn4UJ8zPuEF/%F0%9F%A7%91%F0%9F%8F%BB%E2%80%8D%F0%9F%A6%AF-Flagship-A11y-Annotations?node-id=351-5418&p=f&m=dev)
[Combobox variant Figma](https://www.figma.com/design/wDAxbRmlq4xOn4UJ8zPuEF/%F0%9F%A7%91%F0%9F%8F%BB%E2%80%8D%F0%9F%A6%AF-Flagship-A11y-Annotations?node-id=292-469&m=dev)

PR #1010, opened 2026-03-13
TimeRangePicker accessibility #881

benjaminLeongSK:
Changes:

- Added accessibility features for both dial and combobox component

### DIFF
--- a/src/time-range-picker/combobox-picker/combobox-picker.tsx
+++ b/src/time-range-picker/combobox-picker/combobox-picker.tsx
@@ -418,11 +418,9 @@ export const ComboboxPicker = ({
                     }
                     aria-controls={listboxId}
                     aria-autocomplete="list"
-                    aria-invalid={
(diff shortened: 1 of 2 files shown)


src/form/form-time-range-picker.tsx
```diff
@@ -21,9 +23,13 @@ export const FormTimeRangePicker = ({
     xxlCols,
     ...otherProps
 }: FormTimeRangePickerProps): JSX.Element => {
+    const [internalId] = useState(
+        () => `form-field-${SimpleIdGenerator.generate()}`
```
qroll:
```suggestion
        () => `form-timepicker-${SimpleIdGenerator.generate()}`
```

qroll (changes requested):

src/time-range-picker/combobox-picker/combobox-picker.tsx
qroll: can generate a unique id here instead?

PR #1060, opened 2026-04-08
[BOOKINGSG-9109] refactor: migrate alert to linaria and v4 tokens

ziggyzet:
**Checklist**

-   [X] Added visual tests

**Additional information**
-   Adding `useMemo` to Markup to make sure `maxHeight` attribute in the alert


src/markup/markup.tsx
```diff
@@ -44,6 +41,9 @@ const Component = (props: MarkupProps, ref: React.Ref<HTMLDivElement>) => {
         baseTextSize && styles.sizedMarkup,
         className
     );
+
+    console.log("appliedStyles", className);
```
ghazwan-gt: Let's remove this

src/alert/alert.tsx
```diff
@@ -147,51 +156,87 @@ export const Alert = ({
...
+
+    const renderContent = () => {
+        const containerRef = useRef<HTMLDivElement>(null);
+
+        useApplyStyle(containerRef, {
+            [styles.maxCollapsedHeightVar]:
+                !showHiddenContent && isCollapsed && maxCollapsedHeight
+                    ? `${maxCollapsedHeight}px`
+                    : null,
+        });
```
ghazwan-gt: Let's move the hooks call to top level and keep `renderContent` to be a pure read-only function. 

It’ll align with Hooks conventions and reduce future fragility if content ever becomes conditional.

e2e/tests/components/alert/alert.e2e.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+    test("show more button has correct aria-controls pointing to content", async ({
+        page,
+    }) => {
+        const section = page.getByTestId("story-show-more");
+        const showMoreBtn = section.getByRole("button", { name: "Show more" });
+        await expect(showMoreBtn).toBeVisible({ timeout: 5000 });
+        const controlsId = await showMoreBtn.getAttribute("aria-controls");
+        expect(controlsId).toBeTruthy();
+        await expect(
+            page.locator(`button[aria-controls="${controlsId!}"]`).first()
```
ghazwan-gt: Shouldn't we locate the content instead of the button itself?

```suggestion
            page.locator(`#${controlsId}`)
```

src/alert/alert.styles.ts
ghazwan-gt, line 1: Since we now have this file, let's delete the previous alert.style.tsx

e2e/nextjs-app/src/app/components/alert/page.tsx
ghazwan-gt, line 1: Got this on running on `CI=true`. Problem with applying inline style

Image

src/alert/alert.tsx
```diff
@@ -147,51 +156,87 @@ export const Alert = ({
...
+        }[type] ?? styles.wrapperWarning;
+
+    const iconColorClass =
+        {
+            error: styles.alertIconError,
+            success: styles.alertIconSuccess,
+            warning: styles.alertIconWarning,
+            info: styles.alertIconInfo,
+            description: styles.alertIconDescription,
+        }[type] ?? styles.alertIconWarning;
```
ghazwan-gt: Since this is a static one, can't we have this as a config alongside the icon? Maybe like a mapping of
```
success: { wrapperClass: styles.wrapperSuccess, iconClass: styles.alertIconSuccess, icon: <TickCircleFillIcon aria-hidden /> },
```
then we can have
```
const { wrapperClass, iconClass, icon } = TYPE_CONFIG[type];
```
Later on we just need to check for the custom icon. What do you think?

e2e/nextjs-app/src/app/components/alert/page.tsx
ghazwan-gt, line 1: Why not rely on `e2e/nextjs-app/src/app/components/[component]/[story]/page.tsx` for the page? That one already have a layout with the theme loaded in

ghazwan-gt: I saw that the screenshots don't the expected result

src/alert/alert.styles.ts
```diff
@@ -0,0 +1,204 @@
...
+// WRAPPER
+// =============================================================================
+export const wrapper = css`
+    padding: ${Spacing["spacing-8"]} ${Spacing["spacing-16"]};
+    display: flex;
+`;
+
+export const wrapperError = css`
+    background: ${Colour["bg-error"]};
+    border-left: ${Border["width-020"]} ${Border["solid"]};
```
qroll: since `border-left` is a common style it could be kept in the main `wrapper`. then each variant only needs to specify the colour

src/alert/alert.styles.ts
```diff
@@ -0,0 +1,204 @@
+import { css } from "@linaria/core";
+
+import { Border, Colour, Font, Motion, Spacing } from "../theme/tokens";
+
+// CSS variable for max-height injection
+export const maxCollapsedHeightVar =
+    "--fds-internal-alert-textWrapperContainer-maxCollapsedHeight";
```
qroll: specify in a `tokens` object

```
const tokens = {
  textWrapperContainer: { maxCollapsedHeight: "..." }
}
```

src/alert/alert.styles.ts
```diff
@@ -0,0 +1,204 @@
+import { css } from "@linaria/core";
+
+import { Border, Colour, Font, Motion, Spacing } from "../theme/tokens";
+
+// CSS variable for max-height injection
```
qroll: unnecessary comment

src/alert/alert.tsx
```diff
@@ -147,51 +156,87 @@ export const Alert = ({
...
-                    : undefined
-            }
-            $showMore={showHiddenContent}
-            $hasActionLink={!!actionLink}
-            inert={inertValue(isInert())}
-        >
-            <TextWrapper
-                ref={contentRef}
-                baseTextSize={sizeType === "small" ? "body-sm" : "body-md"}
+    const isCollapsed = isContentOutsideCollapsibleZone();
```
qroll: move `isCollapsed` up to the CONSTS section; for me it's easier to find it there rather than having it mixed with the render functions

you can hoist `isContentOutsideCollapsibleZone` by converting it to a non-arrow function

src/alert/alert.tsx
```diff
@@ -147,51 +156,87 @@ export const Alert = ({
...
                     {renderIcon()}
-                </AlertIconWrapper>
+                </div>
             )}
-            <ContentContainer>
-                {/* 
-                    CSS-BASED TAB ORDER IMPLEMENTATION:
-                    - DOM order: button → link → content (natural tab order)
-                    - Visual order: content → button (via CSS order property)
-                */}
```
qroll, line 191: restore this comment

qroll (changes requested):

e2e/tests/components/alert/alert.e2e.spec.ts
ziggyzet: so i tried this before, but didn't work. not sure why. so the approach i pushed is the one that worked

src/alert/alert.tsx
ziggyzet: does look cleaner! will apply this

src/alert/alert.tsx
```diff
@@ -195,23 +193,33 @@ export const Alert = ({
...
-
-    const iconColorClass =
-        {
-            error: styles.alertIconError,
-            success: styles.alertIconSuccess,
-            warning: styles.alertIconWarning,
-            info: styles.alertIconInfo,
-            description: styles.alertIconDescription,
-        }[type] ?? styles.alertIconWarning;
+    const typeClassMap = {
```
ghazwan-gt: Since the mapping is static, let's move this out of the component so that it isn't recreated on every render

e2e/tests/components/alert/alert.e2e.spec.ts
ghazwan-gt: Yeah I don't think that is correct, since it would mean we're referencing the button to locate the button when we should be locating the content

src/alert/alert.styles.ts
```diff
@@ -0,0 +1,201 @@
...
+    flex: 1;
+    order: 1;
+`;
+
+export const textWrapperContainerShowMoreWithLink = css`
+    margin-bottom: ${Spacing["spacing-8"]};
+`;
+
+export const textWrapperContainerCollapsed = css`
+    max-height: var(${tokens.maxCollapsedHeight});
```
qroll, line 149: do reset the var as well

it's unlikely for alerts to be nested but better be safe

e2e/tests/components/alert/alert.e2e.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+        const story = new StoryPage(page);
+        await use(story);
+    },
+});
+
+test.describe("Alert", () => {
+    // =========================================================================
+    // TYPE VARIANTS
+    // =========================================================================
+    test("type-variants: renders all type variants", async ({ story }) => {
```
qroll: rename the tests for readability and consistency

- Type variants
- Type variants with icon
- Size
- With action link
- With max collapsed height
- Keyboard navigation

e2e/tests/components/alert/alert.e2e.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+        await story.init("type-variants");
+        const alert = story.page.getByLabel("Success").first();
+        await expect(alert).toBeVisible();
+        await compareScreenshot(story, "type-variants");
+    });
+
+    test("type-variants: accessibility aria-label for each type", async ({
```
qroll: can be merged into the previous test case since it's the same page

are you able to use `toMatchAriaSnapshot` here instead?

e2e/tests/components/alert/alert.e2e.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+    story: async ({ page }, use) => {
+        const story = new StoryPage(page);
+        await use(story);
+    },
+});
+
+test.describe("Alert", () => {
+    // =========================================================================
+    // TYPE VARIANTS
+    // =========================================================================
```
qroll: remove these headers, don't find them helpful for sectioning if there's going to be one for every test

e2e/tests/components/alert/alert.e2e.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+        await story.init("with-action-link");
+        const actionLink = story.page.getByTestId("action-link");
+        await expect(actionLink).toBeVisible();
+        await compareScreenshot(story, "with-action-link");
+    });
+
+    // =========================================================================
+    // SHOW MORE / SHOW LESS
+    // =========================================================================
+    test("show-more: shows 'Show more' button when content exceeds maxCollapsedHeight", async ({
```
qroll: merge all 4 into 1 test with steps separating the actions
- On mount
- Click 'Show more'
- Click 'Show less'

e2e/tests/components/alert/alert.e2e.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+        expect(controlsId).toBeTruthy();
+        await expect(
+            story.page.locator(`button[aria-controls="${controlsId!}"]`).first()
+        ).toBeVisible();
+    });
+
+    // =========================================================================
+    // KEYBOARD NAVIGATION
+    // =========================================================================
+    test("keyboard-nav: can toggle show more/less via keyboard (Enter)", async ({
```
qroll: merge all 3 into 1 test. only use the `Tab` to focus the button, this way you can implicitly test the tab behaviour

e2e/nextjs-app/src/app/globals.css
```diff
@@ -11,3 +11,15 @@
     flex-direction: column;
     gap: 1rem;
 }
+
+/* E2E Alert Story Containers */
+.alert-story-container {
```
qroll: for component specific css declare them in the corresponding folder instead

but do we need new styles for these? they seem similar to to the existing `story-row-container` and `story-column-container`. we do not need additional padding as that's built into the story layout

e2e/nextjs-app/src/app/components/alert/max-collapsed-height.e2e.tsx
qroll, line 1: `max-collapsed-height.e2e.tsx`

qroll (changes requested):

e2e/tests/components/alert/__screenshots__/chromium/Alert-size-variants-renders-default-and-small-size-variants-size-variants.png
qroll, line 1: will be good to check the hyperlink and show more button sizes too

e2e/tests/components/alert/alert.e2e.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+    // SHOW MORE / SHOW LESS
+    // =========================================================================
+    test("show-more: shows 'Show more' button when content exceeds maxCollapsedHeight", async ({
+        story,
+    }) => {
+        await story.init("show-more");
+        const showMoreBtn = story.page.getByRole("button", {
+            name: "Show more",
+        });
+        await expect(showMoreBtn).toBeVisible({ timeout: 5000 });
```
ghazwan-gt: Why do we need to put up a timeout?

e2e/tests/components/alert/alert.e2e.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+        await showMoreBtn.focus();
+        await story.page.keyboard.press("Space");
+        await expect(
+            story.page.getByRole("button", { name: "Show less" })
+        ).toBeVisible();
+    });
+
+    test("keyboard-nav: show more button appears before action link in tab order", async ({
```
ghazwan-gt: I suggest to update the case to also check for the tab order once the alert is expanded

zhaoyanxzy:
there's still a tests/alert/alert.spec.tsx referencing the outdated `V2_Color`

e2e/tests/components/alert/alert.e2e.spec.ts
ziggyzet: for this one, i'm just going to delete the approach to find the aria-controls. for some reason we can't do it via the locater without the button tag, which defeats the purpose of finding the content.

e2e/nextjs-app/src/app/components/alert/keyboard-nav.e2e.tsx
```diff
@@ -0,0 +1,19 @@
+"use client";
+
+import { Alert } from "@lifesg/react-design-system/alert";
+
+export default function Story() {
+    return (
+        <div className="alert-keyboard-nav-container">
```
qroll: unreferenced classname, can be removed

e2e/tests/components/alert/alert.e2e.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+test.describe("Alert", () => {
+    test("Type variants", async ({ story }) => {
+        await story.init("type-variants");
+        const alert = story.page.getByLabel("Success").first();
+        await expect(alert).toBeVisible();
+        await compareScreenshot(story, "type-variants");
+
+        await expect(story.page.getByTestId("alert-success"))
+            .toMatchAriaSnapshot(`
+            - text: This is a success alert.
```
qroll: looks like the aria label is not getting reflected

I am expecting something like this
```
- alert "Success": This is a success alert.
```

could be because role is required. can try setting the `role="alert"` prop for the components

qroll (changes requested):

e2e/tests/components/alert/alert.e2e.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+            await expect(
+                story.page.getByRole("button", { name: "Show less" })
+            ).toBeVisible();
+        });
+
+        await test.step("Tab moves focus to next element after alert", async () => {
+            await story.page.getByRole("button", { name: "Show less" }).click();
+            await showMoreBtn.focus();
+            await story.page.keyboard.press("Tab");
+            await expect(story.page.getByTestId("after-alert")).toBeFocused();
```
ghazwan-gt: Let's add an action link such that we're making sure the next navigation from show more button is the action link

e2e/tests/components/alert/alert.e2e.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+    });
+
+    test("Keyboard navigation", async ({ story }) => {
+        await story.init("keyboard-nav");
+
+        const showMoreBtn = story.page.getByRole("button", {
+            name: "Show more",
+        });
+
+        await expect(showMoreBtn).toBeVisible({ timeout: 5000 });
```
ghazwan-gt: Lets remove the timeout

e2e/tests/components/alert/alert.e2e.spec.ts
qroll: can update the aria snapshots with the appropriate status tags per above?

e2e/tests/components/alert/alert.e2e.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+            await compareScreenshot(story, "collapsed");
+        });
+    });
+
+    test("Keyboard navigation", async ({ story }) => {
+        await story.init("keyboard-nav");
+
+        const showMoreBtn = story.page.getByRole("button", {
+            name: "Show more",
+        });
```
qroll: put this in a `locators` object on the StoryPage class. update all instances to use this locator

e2e/tests/components/alert/alert.e2e.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+        await test.step("Tab focuses 'Show more' button", async () => {
+            await story.page.getByTestId("before-alert").focus();
+            await story.page.keyboard.press("Tab");
+            await expect(showMoreBtn).toBeFocused();
+        });
+
+        await test.step("Enter key expands content", async () => {
+            await story.page.keyboard.press("Enter");
+            await expect(
+                story.page.getByRole("button", { name: "Show less" })
```
qroll: extract this into a locator and update all instances

e2e/tests/components/alert/alert.e2e.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+            ).toBeVisible();
+        });
+
+        await test.step("Tab focuses 'Learn more' link", async () => {
+            await story.page.keyboard.press("Tab");
+            await expect(story.page.getByTestId("action-link")).toBeFocused();
```
qroll: could this be a locator too?

e2e/tests/components/alert/alert.e2e.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+        await test.step("Enter key collapses content", async () => {
+            await story.page.keyboard.down("Shift");
+            await story.page.keyboard.press("Tab");
+            await story.page.keyboard.up("Shift");
+            await expect(
+                story.page.getByRole("button", { name: "Show less" })
+            ).toBeFocused();
+
+            await story.page.keyboard.press("Enter");
+            await expect(showMoreBtn).toBeVisible();
```
qroll, line 164: can add a `await expect(showMoreBtn).toBeFocused();` after this line to test tab focus behaviour

e2e/tests/components/alert/alert.e2e.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+            await expect(
+                story.page.getByRole("button", { name: "Show less" })
+            ).toBeVisible();
+        });
+
+        await test.step("Tab moves focus to next element after alert", async () => {
+            await story.page.getByRole("button", { name: "Show less" }).click();
+            await showMoreBtn.focus();
+            await story.page.keyboard.press("Tab");
+            await expect(story.page.getByTestId("after-alert")).toBeFocused();
```
qroll: we don't need to click the button AND focus programmatically here? can simplify to just

```suggestion
            await story.page.keyboard.press("Tab");
            await story.page.keyboard.press("Tab");
            await expect(story.page.getByTestId("after-alert")).toBeFocused();
```

qroll (changes requested):

ghazwan-gt (approved):

e2e/tests/components/alert/alert.e2e.spec.ts
qroll: shall address this when cleaning up the e2e tests

PR #1067, opened 2026-04-09
Jest v30 upgrade

Jazsley:
# Jest v30 Upgrade — Code Review Summary

**Branch:** `jest-v30-upgrade`  
**Author:** Jaz  
**Scope:** Upgrade Jest and related packages from v29 → v30 and fix all
breaking changes, including those introduced by the jsdom v20 → v26 jump.

For the full investigation log see `docs/jest-v30-upgrade-plan.md`.

---

## Commits

| Hash        | Description                                                        |
| ----------- | ------------------------------------------------------------------ |
| `b9351e3` | Bump Jest and related dependencies to v30                          |
| `f463938` | Replace removed Jest 30 matcher aliases with canonical equivalents |
| `6222da3` | Fix setSystemTime calls for @sinonjs/fake-timers v13 compat        |
| `af4e680` | Replace removed jest.SpyInstance type with jest.Spied              |
| `99bfd1d` | Fix slider tests broken by jsdom v26 getBoundingClientRect changes |

---

### 1. Dependency bumps — `package.json`

| Package                     | Before     | After     |
| --------------------------- | ---------- | --------- |
| `jest`                      | `^29.7.0`  | `^30.0.0` |
| `jest-environment-jsdom`    | `^29.7.0`  | `^30.0.0` |
| `@types/jest`               | `^29.5.12` | `^30.0.0` |
| `@testing-library/jest-dom` | `^5.16.3`  | `^6.0.0`  |

No changes to `jest.config.js` were required.

---

### 2. Matcher alias removals — 7 files, 19 occurrences

Jest 30 removed deprecated matcher aliases that had been soft-deprecated since
Jest 26. They now throw at runtime:

```
TypeError: expect(...).toBeCalledWith is not a function
```

| Removed                | Replacement                  |
| ---------------------- | ---------------------------- |
| `.toBeCalled()`        | `.toHaveBeenCalled()`        |
| `.toBeCalledWith(...)` | `.toHaveBeenCalledWith(...)` |
| `.toBeCalledTimes(n)`  | `.toHaveBeenCalledTimes(n)`  |

**Files:** `masked-input.spec.tsx`, `tab.spec.tsx`, `filter.spec.tsx`,
`input.spec.tsx`, `unit-number.spec.tsx`, `time-slot-bar-week.spec.tsx`,
`time-slot-week-view.spec.tsx`

---

### 3. `setSystemTime` now requires a number — 13 files

Jest 30 upgrades `@sinonjs/fake-timers` v11 → v13. Passing a `Date` object
to `setSystemTime` now throws:

```
TypeError: now should be milliseconds since UNIX epoch
```

**Fix:** Append `.getTime()` to convert the `Date` to milliseconds. The fake
clock behaviour is identical — only the argument type changed.

```ts
// Before
jest.setSystemTime(new Date("2024-02-01T12:00:00"));

// After
jest.setSystemTime(new Date("2024-02-01T12:00:00").getTime());
```

**Files:** 13 spec files across `calendar`, `date-input`, `date-range-input`,
`date-navigator`, `footer`, `schedule`, `time-slot-*`, `timetable`, `tests/utils/`.

---

### 4. `jest.SpyInstance` type removed — 1 file

`@types/jest` v30 removes `jest.SpyInstance` from the public API. The
replacement is `jest.Spied<typeof fn>`.

```ts
// Before
let spy: jest.SpyInstance;

// After
let spy: jest.Spied<typeof console.warn>;
```

**File:** `tests/utils/time-helper.spec.ts`

---

### 5. Slider tests broken by jsdom v26 — 4 files

`histogram-slider`, `input-slider`, `input-range-slider`, and
`select-histogram` tests failed after the upgrade. **Confirmed as regressions**
by running them on jest v29 — all passed on v29.

**Root cause:** `react-slider` computes thumb position using two separate DOM
APIs during `mouseDown`:

```
upperBound = slider.clientWidth - thumbRect.getBoundingClientRect().width
value      = (pixelOffset / upperBound) * (max - min) + min
```

In jsdom v20, both APIs returned `0` consistently → `upperBound = 0` → the
library bailed out and never called `onChange`. In jsdom v26, these APIs
returned inconsistent values mid-event → `upperBound` became negative →
`value = NaN` → tests failed.

> Note: `getBoundingClientRect().width` and `clientWidth` are **distinct DOM
> properties** — jsdom tracks them separately, so both must be mocked.

**Fix:** Mock both layout APIs in `beforeEach`, restored in `afterEach`:

```ts
// thumbSize=10, sliderWidth=100 → upperBound=90 → pageX=0 snaps to min ✓
jest.spyOn(Element.prototype, "getBoundingClientRect").mockReturnValue({
    left: 0,
    top: 0,
    right: 100,
    bottom: 10,
    width: 10,
    height: 10,
    x: 0,
    y: 0,
    toJSON: jest.fn(),
});
jest.spyOn(HTMLElement.prototype, "clientWidth", "get").mockReturnValue(100);

// in afterEach:
jest.restoreAllMocks(); // prevents prototype mock leaking to other test files
```

`select-histogram` had an additional issue: `fireEvent.mouseDown` was
incorrectly placed inside `waitFor`, causing the test to retry the event
dispatch every 50ms until timeout (the hang). Events are synchronous and must
be placed **outside** `waitFor`. Only assertions should go inside `waitFor`.

---

## Test results

|                | v29 baseline | v30 (pre-fix) | v30 (post-fix) |
| -------------- | ------------ | ------------- | -------------- |
| Suites passing | 86           | 68            | **86**         |

`npx tsc --noEmit` → zero errors ✅

---

## Known deferred item (v4 branch only)

`tests/theme/utils/css-variable.spec.ts` fails on `pre-release/v4` only (the
file does not exist on `master`). jsdom v26 no longer suppresses `window`
access the same way the test assumes. Tracked as a separate v4 follow-up.


qroll:
for point 5, do you have a link to some article, github issue or SO answer referencing the jsdom changes and proposed workarounds?

Jazsley:
1. How it passed before (jsdom v20, Jest 29)

jsdom v20 never implemented real layout. Every layout API returned a hardcoded zero:

getBoundingClientRect().width → 0
clientWidth → 0
So react-slider computed upperBound = 0 - 0 = 0, hit an internal guard (if upperBound <= 0, do nothing), and never called onChange. Tests passed because the slider silently did nothing — not because the logic was correct.

2. What broke it (jsdom v26, Jest 30)

PR #3514 made jsdom more browser-accurate: during a mouse event dispatch, getBoundingClientRect() now derives its values from the event coordinates (pageX, clientX, etc.) rather than always returning zero.

So mid-event:

getBoundingClientRect().width → non-zero (derived from event)
clientWidth → still 0 (different internal property, still a stub)
Now upperBound = 0 - 10 = -10 → value = (0 / -10) * range = NaN → onChange(NaN) → test assertions fail.

The two references that explain why this happened:

PR #3514 — the actual jsdom code change
Issue #2751 — confirms getBoundingClientRect and clientWidth are tracked separately in jsdom; they are distinct stubs
3. How we fixed it (made tests pass again)

We explicitly mocked both APIs to stable, consistent values before each test:

getBoundingClientRect().width → 10 (thumb size)
clientWidth → 100 (slider width)
Now upperBound = 100 - 10 = 90 → math produces a valid number → onChange is called with a real value → assertions pass.

The references that explain how to do this:

Jest Manual Mocks docs — the jest.spyOn(Prototype, method) pattern for mocking unimplemented jsdom APIs
Jest upgrade guide — confirms jsdom v26 may change behaviour, validating that mocking is the right long-term approach
So the 4 references form a complete story: PR #3514 caused it → issue #2751 explains why jsdom won't fix it → Jest upgrade guide acknowledges the behaviour change → Jest manual mocks docs show the correct fix pattern.

[jsdom/jsdom PR #3514](https://github.com/jsdom/jsdom/pull/3514)
[jsdom/jsdom issue #2751](https://github.com/jsdom/jsdom/issues/2751) 
[Jest upgrade guide — Compatibility](https://jestjs.io/docs/upgrading-to-jest30#compatibility) 
[Jest Manual Mocks — Mocking methods not in JSDOM](https://jestjs.io/docs/manual-mocks#mocking-methods-which-are-not-implemented-in-jsdom)

tests/histogram-slider/histogram-slider.spec.tsx
```diff
@@ -13,6 +13,33 @@ describe("HistogramSlider", () => {
...
+
+        // jsdom v26 (jest-environment-jsdom v30) no longer returns consistent
+        // zeros for layout APIs during event dispatch. react-slider computes
+        // thumb position using both getBoundingClientRect() and clientWidth —
+        // these are separate DOM properties and must be mocked independently.
+        //
+        // getBoundingClientRect().width → thumbSize (set to 10)
+        // clientWidth                   → slider inner width (set to 100)
+        // → upperBound = 100 - 10 = 90; pageX=0 snaps value to min ✓
+        jest.spyOn(Element.prototype, "getBoundingClientRect").mockReturnValue({
```
qroll: noticed that the tests still pass even without this `getBoundingClientRect` mock, so does the value here matter? if not we could fallback to the default stub

Jazsley: actually no need. i think clientWidth enough to not let it become NaN

qroll (approved):

PR #1070, opened 2026-04-09
[DCUBESDQLR-2286] Create new token theme for SGW Digital Lobby

hhung2710:
**Changes**
`New theme tokens`

- Added `sgwdigitallobby` semantic colour token set with full light and dark mode definitions (sgwdigitallobby-tokens.ts)

- Mapped `sgwdigitallobby` to the existing `SupportGoWhereColourSet` for primitive colours

- Registered `sgwdigitallobby` in the semantic colour theme helpers (both light and dark specs)


src/theme/index.ts
qroll, line 213: i see that a new resource scheme was added but it was not updated here

hhung2710: i update the `colourScheme` to use the new `sgwdigitallobby` instead of reuse the `supportgowhere`

qroll: yep that's `ColourScheme` but you also updated `ResourceScheme` which is unused 👀

qroll: we can reuse `"supportgowhere"` because it's being referenced in the footer and navbar resources. no need to add another resource scheme

qroll (approved):

PR #1071, opened 2026-04-10
[MISC][GZ] update readme getting started stylesheet links

ghazwan-gt:
## Summary of Changes
- Updated the root README “Getting Started” CSS setup to use v3 asset URLs.

## Goal
- Prevents consumers from using legacy stylesheet paths that can introduce unintended global styles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,20 +24,29 @@ npm i @lifesg/react-design-system
 
 ## Getting Started
 
-The module requires some custom css for certain components to work properly. As the module does not ship with css, you will need to add this into the `head` of your html file.
+The module requires some custom css for fonts and certain components to work
+properly. You will need to load the corresponding stylesheets in your app.
+
+Add this into the `head` of your html file:
 
 ```tsx
 <link
     rel="stylesheet"
     type="text/css"
-    href="https://assets.life.gov.sg/react-design-system/css/main.css"
+    href="https://assets.life.gov.sg/react-design-system/v3/css/main.css"
+/>
+<link
+    rel="stylesheet"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,20 +24,29 @@ npm i @lifesg/react-design-system
...
 <link
     rel="stylesheet"
     type="text/css"
-    href="https://assets.life.gov.sg/react-design-system/css/main.css"
+    href="https://assets.life.gov.sg/react-design-system/v3/css/main.css"
+/>
+<link
+    rel="stylesheet"
+    type="text/css"
+    href="https://assets.life.gov.sg/react-design-system/v3/css/open-sans.css"
```
qroll, line 46: let's add a note here somewhere (an [alert box](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts)) that this is for LifeSG theme since fonts are theme specific. for other themes, redirect them to the installation instructions at https://designsystem.life.gov.sg/react/index.html?path=/docs/getting-started-installation--docs

(side note: in v4 we could probably `@import` fonts directly for each theme but I haven't checked how nextjs and vite handle these)

README.md
qroll: can be updated for V3 instructions too :oops:

qroll (changes requested):

README.md
ghazwan-gt: Yeah, we could explore this later on

qroll (approved):

PR #1074, opened 2026-04-10
v4 Jest v30 upgrade

Jazsley:
# Jest v30 Upgrade — Code Review Summary

**Author:** Jaz  
**Scope:** Upgrade Jest and related packages from v29 → v30 and fix all
breaking changes, including those introduced by the jsdom v20 → v26 jump.

For the full investigation log see `docs/jest-v30-upgrade-plan.md`.

---

## Commits — `jest-v30-upgrade` (master)

| Hash        | Description                                                        |
| ----------- | ------------------------------------------------------------------ |
| `a36b22675` | Bump Jest and related dependencies to v30                          |
| `1a8da8947` | Replace removed Jest 30 matcher aliases with canonical equivalents |
| `bb8da82de` | Fix setSystemTime calls for @sinonjs/fake-timers v13 compat        |
| `3d30f85ee` | Replace removed jest.SpyInstance type with jest.Spied              |
| `c6134206d` | Fix slider tests broken by jsdom v26 getBoundingClientRect changes |

## Commits — `pre-release/v4-CCUBE-2091` (additional)

| Hash        | Description                                                               |
| ----------- | ------------------------------------------------------------------------- |
| `1f7d9a60f` | Bump Jest and related dependencies to v30                                 |
| `b971d3cc9` | Replace removed Jest 30 matcher aliases with canonical equivalents        |
| `df66367c0` | Fix setSystemTime calls to pass milliseconds for @sinonjs/fake-timers v13 |
| `44c86c457` | Replace removed jest.SpyInstance type with jest.Spied                     |
| `95ffbe6bb` | Fix slider tests broken by jsdom v26 changes                              |
| `2d3d47c14` | Fix css-variable test broken by jsdom v26 non-configurable window         |
| `b1654d9b8` | Remove @types/dompurify — dompurify v3 bundles its own types              |
| `d8445eeb7` | Update time-helper tests for jest.Spied and setSystemTime compat          |
| `7c660107f` | Sync package-lock.json with package.json (patch-package missing entries)  |

---

### 1. Dependency bumps — `package.json`

| Package                     | Before     | After     |
| --------------------------- | ---------- | --------- |
| `jest`                      | `^29.7.0`  | `^30.0.0` |
| `jest-environment-jsdom`    | `^29.7.0`  | `^30.0.0` |
| `@types/jest`               | `^29.5.12` | `^30.0.0` |
| `@testing-library/jest-dom` | `^5.16.3`  | `^6.0.0`  |

No changes to `jest.config.js` were required.

---

### 2. Matcher alias removals — 7 files, 19 occurrences

Jest 30 removed deprecated matcher aliases that had been soft-deprecated since
Jest 26. They now throw at runtime:

```
TypeError: expect(...).toBeCalledWith is not a function
```

| Removed                | Replacement                  |
| ---------------------- | ---------------------------- |
| `.toBeCalled()`        | `.toHaveBeenCalled()`        |
| `.toBeCalledWith(...)` | `.toHaveBeenCalledWith(...)` |
| `.toBeCalledTimes(n)`  | `.toHaveBeenCalledTimes(n)`  |

**Files:** `masked-input.spec.tsx`, `tab.spec.tsx`, `filter.spec.tsx`,
`input.spec.tsx`, `unit-number.spec.tsx`, `time-slot-bar-week.spec.tsx`,
`time-slot-week-view.spec.tsx`

---

### 3. `setSystemTime` now requires a number — 13 files

Jest 30 upgrades `@sinonjs/fake-timers` v11 → v13. Passing a `Date` object
to `setSystemTime` now throws:

```
TypeError: now should be milliseconds since UNIX epoch
```

**Fix:** Append `.getTime()` to convert the `Date` to milliseconds. The fake
clock behaviour is identical — only the argument type changed.

```ts
// Before
jest.setSystemTime(new Date("2024-02-01T12:00:00"));

// After
jest.setSystemTime(new Date("2024-02-01T12:00:00").getTime());
```

**Files:** 13 spec files across `calendar`, `date-input`, `date-range-input`,
`date-navigator`, `footer`, `schedule`, `time-slot-*`, `timetable`, `tests/utils/`.

---

### 4. `jest.SpyInstance` type removed — 1 file

`@types/jest` v30 removes `jest.SpyInstance` from the public API. The
replacement is `jest.Spied<typeof fn>`.

```ts
// Before
let spy: jest.SpyInstance;

// After
let spy: jest.Spied<typeof console.warn>;
```

**File:** `tests/utils/time-helper.spec.ts`

---

### 5. Slider tests broken by jsdom v26 — 4 files

`histogram-slider`, `input-slider`, `input-range-slider`, and
`select-histogram` tests failed after the upgrade. **Confirmed as regressions**
by running them on jest v29 — all passed on v29.

**Root cause:** `react-slider` computes thumb position using two separate DOM
APIs during `mouseDown`:

```
upperBound = slider.clientWidth - thumbRect.getBoundingClientRect().width
value      = (pixelOffset / upperBound) * (max - min) + min
```

In jsdom v20, both APIs returned `0` consistently → `upperBound = 0` → the
library bailed out and never called `onChange`. In jsdom v26, these APIs
returned inconsistent values mid-event → `upperBound` became negative →
`value = NaN` → tests failed.

> Note: `getBoundingClientRect().width` and `clientWidth` are **distinct DOM
> properties** — jsdom tracks them separately, so both must be mocked.

**Fix:** Mock both layout APIs in `beforeEach`, restored in `afterEach`:

```ts
// thumbSize=10, sliderWidth=100 → upperBound=90 → pageX=0 snaps to min ✓
jest.spyOn(Element.prototype, "getBoundingClientRect").mockReturnValue({
    left: 0,
    top: 0,
    right: 100,
    bottom: 10,
    width: 10,
    height: 10,
    x: 0,
    y: 0,
    toJSON: jest.fn(),
});
jest.spyOn(HTMLElement.prototype, "clientWidth", "get").mockReturnValue(100);

// in afterEach:
jest.restoreAllMocks(); // prevents prototype mock leaking to other test files
```

`select-histogram` had an additional issue: `fireEvent.mouseDown` was
incorrectly placed inside `waitFor`, causing the test to retry the event
dispatch every 50ms until timeout (the hang). Events are synchronous and must
be placed **outside** `waitFor`. Only assertions should go inside `waitFor`.

**References:**

-   [jsdom/jsdom PR #3514](https://github.com/jsdom/jsdom/pull/3514) — "MouseEvent: fix pageX/Y and offsetX/Y during dispatch" — a browser-correctness fix (merged Apr 2023, landed jsdom v22+) that made jsdom derive `pageX`/`offsetX` from `getBoundingClientRect()` during event dispatch, matching real browser behaviour. Our tests broke not because this change was wrong, but because they relied on jsdom's old non-standard behaviour where every layout API returned `0` consistently — `react-slider` accidentally worked because `0 - 0 = 0` let the library bail out gracefully. The fix made `getBoundingClientRect().width` non-zero mid-event while `clientWidth` stayed `0`, turning `upperBound` negative and `value` into `NaN`.
-   [jsdom/jsdom issue #2751](https://github.com/jsdom/jsdom/issues/2751) — "Guidance about mock implementation for layout related features" — jsdom's stated position that APIs like `getBoundingClientRect` and `clientWidth` are intentional stubs; the project will not implement real layout, so tests must mock them
-   [Jest upgrade guide — Compatibility](https://jestjs.io/docs/upgrading-to-jest30#compatibility) — official acknowledgement: "jest-environment-jsdom now uses JSDOM v26. This update may introduce behavior changes"
-   [Jest Manual Mocks — Mocking methods not in JSDOM](https://jestjs.io/docs/manual-mocks#mocking-methods-which-are-not-implemented-in-jsdom) — official Jest pattern for mocking unimplemented jsdom APIs (uses `window.matchMedia` as example; same `jest.spyOn` approach applies to `getBoundingClientRect` / `clientWidth`)

---

### `jest-v30-upgrade` (master)

|                | v29 baseline | v30 (pre-fix) | v30 (post-fix) |
| -------------- | ------------ | ------------- | -------------- |
| Suites passing | 86           | 68            | **86**         |

`npx tsc --noEmit` → zero errors ✅

### `pre-release/v4-CCUBE-2091`

|                | v30 (post-fix) |
| -------------- | -------------- |
| Suites passing | **86**         |

`npx tsc --noEmit` → zero errors ✅

---

### 6. css-variable test broken by jsdom v26 — 1 file

`tests/theme/utils/css-variable.spec.ts` exists only on `pre-release/v4`.

**Root cause:** The test simulated a missing `window` by calling
`delete globalThis.window`. In jsdom v26, `globalThis.window` is exposed
as a non-deletable getter — `delete` silently does nothing, `window`
remains, and the `!globalThis.window` guard in `parseCSSVariableValue`
never fires. The test asserting an empty-string return and a console warning
both failed.

The guard in `src/theme/utils/css-variable.ts`:

```ts
if (!globalThis.window || !sourceElement) {
    console.warn(...);
    return "";
}
```

**Fix:** Mock the `window` getter using `jest.spyOn` to return `undefined`.
This intercepts the property read at the point the guard checks it, making
`!globalThis.window` evaluate to `true` — testing the exact branch the
original test intended.

```ts
// Before — delete silently no-ops in jsdom v26 (window is a getter):
delete globalThis.window;
const value = parseCSSVariableValue("var(--fds-breakpoint-md-min)", root);

// After — mock the getter so !globalThis.window is true when the guard runs:
const windowSpy = jest
    .spyOn(globalThis, "window", "get")
    .mockReturnValue(undefined as any);
const value = parseCSSVariableValue("var(--fds-breakpoint-md-min)", root);
// ... assertions ...
windowSpy.mockRestore();
```

**Why `jest.spyOn` works here:** jsdom v26 defines `window` as a configurable
getter on `globalThis` (`configurable: true, get: [Function]`). `delete`
fails silently on it, but `jest.spyOn` intercepts the getter — the same
mechanism used to mock `window.matchMedia` and other browser APIs.

**Committed:** `2d3d47c14` on `pre-release/v4-CCUBE-2091`.

---

### 7. `@types/dompurify` removed — `package.json`

`@types/dompurify` was present in the v4 `package.json` but is incompatible
with the jest v30 upgrade. DOMPurify v3 (already a dependency at `^3.2.5`)
bundles its own TypeScript declarations. Having `@types/dompurify` installed
alongside it creates two competing type definitions for the same module,
causing:

```
error TS2688: Cannot find type definition file for 'dompurify'.
```

**Fix:** Remove `@types/dompurify` from `devDependencies`. The bundled types
in `dompurify` itself take over with no change to functionality.

**Committed:** `b1654d9b8` on `pre-release/v4-CCUBE-2091`.

---

### 8. `package-lock.json` out of sync with `package.json` — CI failure

The CI pipeline (`npm ci`) failed on the branch with:

```
npm error `npm ci` can only install packages when your package.json and
package-lock.json or npm-shrinkwrap.json are in sync.

npm error Missing: patch-package@8.0.1 from lock file
npm error Missing: @yarnpkg/lockfile@1.1.0 from lock file
...
```

**Root cause:** `patch-package` and its transitive dependencies were present
in `package.json` but absent from `package-lock.json`. The lock file had
drifted from the manifest — likely because `patch-package` was added to
`package.json` without a subsequent `npm install` being committed.

**Fix:** Regenerate the lock file with `npm install --package-lock-only
--ignore-scripts` (skipping the `postinstall` script that runs
`patch-package`, which itself requires node_modules to be present). The
resulting lock file includes `patch-package@8.0.1` and all missing
transitive entries.

**Committed:** `7c660107f` on `pre-release/v4-CCUBE-2091`.


qroll:
avoid using the `pre-release/*` or `release/*` prefixes in the future

tests/theme/utils/css-variable.spec.ts
```diff
@@ -98,22 +98,21 @@ describe("parseCSSVariableValue", () => {
...
-        delete globalThis.window;
 
+        // jsdom v26 makes globalThis.window non-configurable, so
+        // `delete globalThis.window` silently no-ops and the guard never
+        // triggers. Passing null as sourceElement hits the same guard branch
+        // (!globalThis.window || !sourceElement) and produces the same result.
         const value = parseCSSVariableValue(
             "var(--fds-breakpoint-md-min)",
-            root
+            null
```
qroll, line 124: this changes the intention of the test though

we should check that we handle **both** missing window and sourceElement correctly

is there an alternative method to simulate that window is not available?

qroll (changes requested):

tests/theme/utils/css-variable.spec.ts
Jazsley: const windowSpy = jest
            .spyOn(globalThis, "window", "get")
            .mockReturnValue(undefined as any);
            
            how about this?

qroll: it doesn't look like this works either, since the window object itself is not configurable

but I don't want to spend too much time fighting with jsdom

seeing some solutions around the jsdom issue to patch the library, would something like [this](https://redirect.github.com/firefox-devtools/profiler/commit/f8aebfe4814ac5b00f69fd851c7f3633975b3ede) work for our current set up?

Jazsley: i tested and it works. im not sure why my pipeline is failing though.

Jazsley: The Firefox Profiler approach (patch-package) could work, but it adds patch-package as a dependency that runs on every npm install and needs to be maintained whenever jsdom updates — which feels like heavy overhead for a single test.

The root issue is that jsdom v26 defines [globalThis.window](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) as configurable: false, making it impossible to spy on with [jest.spyOn](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) — even [Object.defineProperty](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) can't relax a non-configurable property once set. This is a [known Jest 30 issue](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) that jsdom has declined to fix ([jsdom/jsdom#3492](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)).

Since the guard is [!globalThis.window || !sourceElement](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), both conditions lead to the exact same code path — same [console.warn](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), same "" return. Passing null as [sourceElement](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) exercises and verifies the entire branch with identical observable behaviour. The test has been renamed to reflect it covers the general unavailability guard rather than one specific condition.

Happy to go the patch-package route if you'd prefer, but wanted to flag the trade-off first.

tests/utils/time-helper.spec.ts
zhaoyanxzy: here also needs to be cleanup. Should do a global search for `jest.SpyInstance`

qroll (approved):

PR #1079, opened 2026-04-13
DataTable accessibility 

benjaminLeongSK:
**Type of changes**

-   [x] New feature (non-breaking change which adds functionality)

**Description of changes**

- Add aria-labels for data-table
- Make clickable headers focusable
- introduce `scrollable` and `keyFieldLabel` prop

**Checklist**

-   [x] Updated documentation
-   [x] Screen reader annotates as per figma

-   Link to [Figma](https://www.figma.com/design/wDAxbRmlq4xOn4UJ8zPuEF/%F0%9F%A7%91%F0%9F%8F%BB%E2%80%8D%F0%9F%A6%AF-Flagship-A11y-Annotations?node-id=2683-15803&t=3RJgrkT2QVPAEJQM-4)

### DIFF
--- a/src/data-table/data-table.styles.tsx
+++ b/src/data-table/data-table.styles.tsx
@@ -199,6 +199,11 @@ export const HeaderCell = styled.th<HeaderCellProps>`
             `;
         }
     }};
+    &:focus-within {
+        outline: 2px solid ${Colour["focus-ring"]};
+        outline-offset: -3px;
+        z-index: 1;
+    }
 `;
 
 export const HeaderCellWrapper = styled.div`
@@ -211,19 +216,6 @@ export const HeaderCellWrapper = styled.div`
     svg {
         color: ${fontColor};
         margin-left: 0.5rem;
-    }
-`;
-
-export const ClickableHeader = styled(BasicButton)`
-    position: absolute;
-    inset: 0.25rem;
-    z-index: 2;
-    cursor: pointer;
-    border-radius: ${Radius["sm"]};
-
-    &:focus-visible {
-        outline: 2px solid ${Colour["focus-ring"]};
-        outline-offset: 0;
     }
 `;
 
--- a/src/data-table/data-table.styles.tsx
+++ b/src/data-table/data-table.styles.tsx
@@ -211,6 +211,16 @@ export const HeaderCellWrapper = styled.div`
     }
 `;
 
+export const ClickableHeader = styled(BasicButton)`
+    ${Font["body-md-semibold"]}
+    display: inline-flex;
+    align-items: center;
+    width: 100%;
+    color: ${fontColor};
+    cursor: pointer;
+    text-align: left;
+`;
+
 export const BodyRow = styled.tr<BodyRowProps>`
     background-color: ${(props) => {
         if (props.$isSelected) {
--- a/src/data-table/data-table.styles.tsx
+++ b/src/data-table/data-table.styles.tsx
@@ -184,6 +184,7 @@ export const HeaderRow = styled.tr`
 `;
 
 export const HeaderCell = styled.th<HeaderCellProps>`
+    position: relative;
     padding: ${(props) =>
         props.$isCheckbox ? "1.25rem 0.5rem 1.25rem 1.5rem" : "1.25rem 1rem"};
     text-align: left;
@@ -201,6 +202,8 @@ export const HeaderCell = styled.th<HeaderCellProps>`
 `;
 
 export const HeaderCellWrapper = styled.div`
+    position: relative;
+    z-index: 1;
     display: flex;
     flex-direction: row;
     align-items: center;
@@ -212,13 +215,16 @@ export const HeaderCellWrapper = styled.div`
 `;
 
 export const ClickableHeader = styled(BasicButton)`
-    ${Font["body-md-semibold"]}
-    display: inline-flex;
-    align-items: center;
-    width: 100%;
-    color: ${fontColor};
+    position: absolute;
+    inset: 0.25rem;
+    z-index: 2;
     cursor: pointer;
-    text-align: left;
+    border-radius: ${Radius["sm"]};
+
+    &:focus-visible {
+        outline: 2px solid ${Colour["focus-ring"]};
+        outline-offset: 0;
+    }
 `;
 
 export const BodyRow = styled.tr<BodyRowProps>`
--- a/src/data-table/data-table.styles.tsx
+++ b/src/data-table/data-table.styles.tsx
@@ -202,13 +202,10 @@ export const HeaderCell = styled.th<HeaderCellProps>`
     &:focus-within {
         outline: 2px solid ${Colour["focus-ring"]};
         outline-offset: -3px;
-        z-index: 1;
     }
 `;
 
 export const HeaderCellWrapper = styled.div`
-    position: relative;
-    z-index: 1;
     display: flex;
     flex-direction: row;
     align-items: center;
--- a/src/data-table/data-table.styles.tsx
+++ b/src/data-table/data-table.styles.tsx
@@ -184,7 +184,6 @@ export const HeaderRow = styled.tr`
 `;
 
 export const HeaderCell = styled.th<HeaderCellProps>`
-    position: relative;
     padding: ${(props) =>
         props.$isCheckbox ? "1.25rem 0.5rem 1.25rem 1.5rem" : "1.25rem 1rem"};
     text-align: left;
--- a/src/data-table/data-table.tsx
+++ b/src/data-table/data-table.tsx
@@ -1,18 +1,10 @@
 import { ArrowDownIcon, ArrowUpIcon } from "@lifesg/react-icons";
-import {
-    ReactNode,
-    isValidElement,
-    useCallback,
-    useEffect,
-    useRef,
-    useState,
-} from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 import { useInView } from "react-intersection-observer";
 import { useResizeDetector } from "react-resize-detector";
 import { LoadingDotsSpinner } from "../animations";
 import { Checkbox } from "../checkbox";
 import { ErrorDisplay } from "../error-display";
-import { VisuallyHidden } from "../shared/accessibility";
 import { Typography } from "../typography";
 import { useEventListener } from "../util/use-event-listener";
 import {
@@ -36,6 +28,7 @@ import {
     TextButton,
 } from "./data-table.styles";
 import { DataTableProps, HeaderProps, RowProps } from "./types";
+import { VisuallyHidden, concatIds } from "../shared/accessibility";
 
 export const DataTable = ({
     id,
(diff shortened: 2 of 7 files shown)


src/data-table/types.ts
```diff
@@ -56,6 +56,9 @@ interface HeaderItemProps {
     // technically ReactNode also includes string, but this is more explicit
     label: string | ReactNode;
     clickable?: boolean | undefined;
+    sortable?: boolean | undefined;
```
qroll: don't need an additional property

we can follow what was done for renderSortedArrow and check `sortIndicators?.[fieldKey]`

src/data-table/types.ts
```diff
@@ -56,6 +56,9 @@ interface HeaderItemProps {
...
+    /** Marks this column as part of the row checkbox screen reader label */
+    keyFieldLabel?: boolean | undefined;
```
qroll:
```suggestion
    /** Used with `enableMultiSelect`. Marks this column as descriptive of the row (e.g. a name or unique ID). Its value is used to label the row's checkbox for screen readers. */
    keyColumn?: boolean | undefined;
```

src/data-table/data-table.tsx
```diff
@@ -233,20 +329,47 @@ export const DataTable = ({
...
-                    {typeof label === "string" ? (
-                        <Typography.BodyBL weight="bold">
-                            {label}
-                        </Typography.BodyBL>
-                    ) : (
-                        label
-                    )}
-                    {renderSortedArrow(fieldKey)}
-                </HeaderCellWrapper>
+                {clickable ? (
```
qroll: can simplify the markup, no need to use a ternary here

```
<>
  <HeaderCellWrapper>...</HeaderCellWrapper>
  {clickable || sortable && <VisuallyHidden><button>...</button></VisuallyHidden>}
```

src/data-table/data-table.tsx
```diff
@@ -359,12 +486,14 @@ export const DataTable = ({
...
                     <Checkbox
                         checked={isRowSelected(rowId)}
+                        aria-label={getRowCheckboxAriaLabel(row, index)}
                         onClick={() => {
                             if (onSelect) {
                                 onSelect(rowId, !isRowSelected(rowId));
                             }
                         }}
                         disabled={isDisabledRow(rowId)}
+                        aria-disabled={isDisabledRow(rowId)}
```
qroll: these two prop don't make sense to be used together

can use `focusableWhenDisabled` instead since this is the shared Checkbox component, to achieve the tab order described in the specs

src/data-table/data-table.tsx
```diff
@@ -277,12 +400,15 @@ export const DataTable = ({
                 data-testid={getDataTestId("header-selection")}
                 $clickable={false}
                 $isCheckbox={true}
+                scope="col"
```
qroll, line 340: just curious, what does this attribute add? does it change how the screen reader process the table?

src/data-table/data-table.tsx
```diff
@@ -150,6 +159,91 @@ export const DataTable = ({
         return headers.length + (enableMultiSelect ? 1 : 0);
     };
 
+    const getHeaderCheckboxAriaLabel = (): string => {
+        const totalColumns = getTotalColumns();
+        return `Select all rows, Column 1 of ${totalColumns}`;
```
qroll: no need to read out the row/column info as that's typically done by the screen reader. can cmiiw if that's different on other platforms

```suggestion
        return `Select all rows`;
```

src/data-table/data-table.tsx
```diff
@@ -150,6 +159,91 @@ export const DataTable = ({
...
+        }
+
+        if (isValidElement(node)) {
+            return extractTextFromNode(node.props.children);
+        }
+
+        return "";
+    };
+
+    const getKeyFieldText = (header: HeaderProps, row: RowProps): string => {
```
qroll: doesn't need to be so complicated; we can assign a unique id to the row and link that to the checkbox with `aria-labelledby`

qroll (changes requested):

src/data-table/data-table.tsx
benjaminLeongSK: by wrapping the button in <VisuallyHidden> the tab focus border wont be visible. is that acceptable?

src/data-table/data-table.tsx
benjaminLeongSK: Hmm actually didnt change how the screen reader sees the table, just added this in as per the figma descriptions

src/data-table/data-table.tsx
qroll: oh yea ideally we should have a focus ring on the current item

since the button would be a sibling we could use the [css selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/Next-sibling_combinator) here?

```
button:focus-visible + & {
}
```

need to move the button up in my example for this to work I think

src/data-table/data-table.tsx
qroll: ok good to know

src/data-table/data-table.tsx
benjaminLeongSK: > cant seem to get it to work with the VisuallyHidden component. will revert to ClickableHeader button component for now

Managed to get it to work by using focus-within

src/data-table/data-table.tsx
benjaminLeongSK: yeah it is working ok on macbook VO but android and windows wasnt sounding it out

qroll: I'm wondering how useful it is to read the column for a select all checkbox

src/data-table/data-table.tsx
```diff
@@ -150,6 +151,54 @@ export const DataTable = ({
...
+
+    const getSortButtonAriaLabel = (label: string, fieldKey: string) => {
+        const nextSortDirection =
+            getSortDirection(fieldKey) === "asc" ? "descending" : "ascending";
+
+        return `Sort ${label} by ${nextSortDirection} order`;
+    };
+
+    const getRowPositionId = (rowId: string) => {
+        return `${id || "table"}-row-${rowId}-position`;
```
qroll: like with previous PRs, don't use a fixed id because there's a chance of collision

generate a unique id for this table

src/data-table/data-table.tsx
```diff
@@ -150,6 +151,54 @@ export const DataTable = ({
...
+    const getKeyColumnCellId = (rowId: string, fieldKey: string) => {
+        return `${id || "table"}-row-${rowId}-${fieldKey}-key-column`;
+    };
+
+    const getRowCheckboxAriaLabelledBy = (rowId: string) => {
+        const keyColumnIds = headers
+            .map((header) => {
+                if (typeof header === "string" || !header.keyColumn) {
+                    return undefined;
+                }
```
qroll: a minor optimisation, we can extract the key columns to some constant variable so it doesn't need to recomputed for every row

src/data-table/data-table.tsx
```diff
@@ -228,15 +277,37 @@ export const DataTable = ({
               }
             : header;
 
+        const isSortable = !!sortIndicators?.[fieldKey];
```
qroll: reuse `getSortDirection`

src/data-table/data-table.tsx
```diff
@@ -228,15 +277,37 @@ export const DataTable = ({
...
+                            type="button"
+                            aria-label={
+                                typeof label === "string"
+                                    ? isSortable
+                                        ? getSortButtonAriaLabel(
+                                              label,
+                                              fieldKey
+                                          )
+                                        : label
+                                    : undefined
```
qroll: this whole logic could be moved to `getSortButtonAriaLabel`

also thinking that instead of checking for label string only, you could use `aria-labelledby` pointing to the header cell wrapper

src/data-table/data-table.tsx
```diff
@@ -277,12 +348,15 @@ export const DataTable = ({
                 data-testid={getDataTestId("header-selection")}
                 $clickable={false}
                 $isCheckbox={true}
+                scope="col"
             >
                 <CheckBoxWrapper>
+                    <VisuallyHidden>Row selection</VisuallyHidden>
```
qroll, line 343: what's the use of this element?

src/data-table/data-table.tsx
```diff
@@ -335,6 +409,11 @@ export const DataTable = ({
             <BodyCell
                 data-testid={getDataTestId(`row-${cellId}`)}
                 key={cellId}
+                id={
+                    isKeyColumn
+                        ? getKeyColumnCellId(rowId, fieldKey)
+                        : undefined
+                }
```
qroll: I think it's ok to compute the id for every cell. no need to specially handle

```suggestion
                id={getCellId(rowId, fieldKey)}
```

src/data-table/data-table.styles.tsx
```diff
@@ -198,9 +199,16 @@ export const HeaderCell = styled.th<HeaderCellProps>`
...
+    &:focus-within {
+        outline: 2px solid ${Colour["focus-ring"]};
+        outline-offset: -3px;
+        z-index: 1;
+    }
 `;
 
 export const HeaderCellWrapper = styled.div`
+    position: relative;
+    z-index: 1;
```
qroll: do we still need these additional position styles on  HeaderCell and HeaderCellWrapper?

src/data-table/data-table.styles.tsx
```diff
@@ -198,9 +199,16 @@ export const HeaderCell = styled.th<HeaderCellProps>`
             `;
         }
     }};
+    &:focus-within {
+        outline: 2px solid ${Colour["focus-ring"]};
+        outline-offset: -3px;
+        z-index: 1;
```
qroll: why is z index needed?

qroll (changes requested):

src/data-table/data-table.tsx
benjaminLeongSK: Screenshot 2026-04-15 at 10 35 10 AM

it was to fix this accessibility complain when the `enableSelectAll` is false there will be an empty header in place of the select all checkbox

src/data-table/data-table.tsx
benjaminLeongSK: not sure if i understood what you meant correctly. but wont `aria-labelledby` override `aria-label`? or you meant to use `aria-labelledby` instead of `aria-label` something like this?:

```
<VisuallyHidden>
                        <button
                            type="button"
                            aria-labelledby={
                                isSortable
                                    ? getSortButtonAriaLabel(label, fieldKey)
                                    : wrapperId
                            }
                            onClick={() => onHeaderClick?.(fieldKey)}
                        />
                    </VisuallyHidden>
                )}

                <HeaderCellWrapper id={wrapperId}>
                    {typeof label === "string" ? (
```

src/data-table/data-table.tsx
```diff
@@ -59,6 +61,13 @@ export const DataTable = ({
     const headerRef = useRef<HTMLTableSectionElement>(null);
     const actionBarRef = useRef<HTMLDivElement>(null);
     const wrapperRef = useRef<HTMLDivElement>(null);
+    const internalId = useRef(id || SimpleIdGenerator.generate());
```
qroll: use state instead so that you don't have to access via `.current` each time

src/data-table/data-table.tsx
qroll:
```
<HeaderCellWrapper id="headerCellWrapperId">
  {label} // <---------- original code
</HeaderCellWrapper>
<VisuallyHidden>
  <button>Sort <span aria-labelledby="headerCellWrapperId" /> by x order</button>
</VisuallyHidden>
```

src/data-table/data-table.tsx
```diff
@@ -150,6 +159,52 @@ export const DataTable = ({
...
+
+    const getHeaderWrapperId = (fieldKey: string) => {
+        return `${internalId.current}-header-${fieldKey}`;
+    };
+
+    const getRowPositionId = (rowId: string) => {
+        return `${internalId.current}-row-${rowId}-position`;
+    };
+
+    const getKeyColumnCellId = (rowId: string, fieldKey: string) => {
```
qroll:
```suggestion
    const getCellId = (rowId: string, fieldKey: string) => {
```

src/data-table/data-table.styles.tsx
qroll: what about position relative btw?

src/data-table/data-table.tsx
qroll: can drop it and defer to default screen reader behaviour. it seems alright in Talkback

src/data-table/data-table.tsx
```diff
@@ -350,21 +428,28 @@ export const DataTable = ({
...
+
         return (
             <BodyCell
                 data-testid={getDataTestId(`row-${rowId}-selection`)}
                 $isCheckbox={true}
             >
                 <CheckBoxWrapper>
+                    <VisuallyHidden id={getRowPositionId(rowId)}>
+                        {`Row ${index + 1} of ${rows?.length ?? 0}`}
+                    </VisuallyHidden>
```
qroll: this too. should be handled by the screen reader

benjaminLeongSK: hmm the screen reader wont read out the rows by default. unless we add it as part of the aria-describedBy?

qroll: Talkback and Voiceover are reading the row for me. and the confusing thing is it treats the header row as a row, so I'm getting `Row 1 of 5, xxxx, Row 2 of 6`

I think we should defer to the screen reader behaviour

benjaminLeongSK: hmm thats weird. sorry but can you help me to check Voiceover reader on your end after my latest commit. cause mine is not reading it out at all.

qroll: yup it's fine on Voiceover navigation and Talkback's Rows and Columns mode

src/data-table/data-table.styles.tsx
qroll: removed

qroll (changes requested):

qroll (approved):

PR #1083, opened 2026-04-13
[BOOKINGSG-9215][GZ] fix broken css import

ghazwan-gt:
## **Type of changes**

-   [X] Bug fix (non-breaking change which fixes an issue) for [Invalid import path in build](https://sgtechstack.atlassian.net/browse/BOOKINGSG-9215)

## Summary
Fixes broken CSS import paths in DS package artifacts, which caused Next.js consumer builds to fail

## Root Cause
`rollup-plugin-lib-style` uses global magic-path replacement, which is not per-importer-file aware. With nested preserved modules and transitive style dependencies, this can produce invalid relative CSS imports.

## Changes
Adds a local post-build Rollup plugin, it works by
- scans emitted JS and CSS files per output directory
- detects broken CSS import paths
- rewrites to correct relative paths based on actual emitted CSS location
- skips ambiguous basename matches

## **Checklist**

-   [ ] Updated documentation
-   [ ] Added/updated tests

### DIFF
--- a/rollup/fix-css-import-paths-plugin.js
+++ b/rollup/fix-css-import-paths-plugin.js
@@ -3,6 +3,7 @@ import path from "node:path";
 
 const IMPORT_CSS_REGEX = /(import\s*["'])([^"']+\.css)(["'])/g;
 const REQUIRE_CSS_REGEX = /(require\(\s*["'])([^"']+\.css)(["']\s*\))/g;
+const LIB_STYLE_MAGIC_PREFIX = "@@_MAGIC_PATH_@@";
 
 // Recursively collects files with a given extension while skipping selected
 // output directories (for example, skip dist/cjs while processing dist).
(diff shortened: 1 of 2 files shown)


ghazwan-gt:
Raised an issue in the repo
https://github.com/DanielAmenou/rollup-plugin-lib-style/issues/13

rollup/fix-css-import-paths-plugin.js
```diff
@@ -0,0 +1,193 @@
...
+        outputDir,
+        jsFile,
+        importPath,
+    });
+    if (fs.existsSync(resolvedPath)) return null;
+
+    const cssMatches = cssByBasename.get(path.basename(importPath));
+    if (cssMatches?.length !== 1) return null;
+
+    return toRelativeImportPath(jsFile, cssMatches[0]);
```
qroll, line 113: just for my understanding: there should only ever be 1 match for the css file, correct? since the hash generated by Linaria is unique iirc

wondering why `collectFiles` handles it as an array in the first place? is it to cater for jsFiles?

rollup/fix-css-import-paths-plugin.js
```diff
@@ -0,0 +1,193 @@
...
+const getRewrittenCssImportPath = ({
+    outputDir,
+    jsFile,
+    importPath,
+    cssByBasename,
+}) => {
+    if (!isFixableCssImportPath(importPath)) return null;
+
+    if (importPath.startsWith(LIB_STYLE_MAGIC_PREFIX)) {
+        const normalizedPath = normalizeImportPath(importPath);
```
qroll, line 89: found the flow of this a little hard to trace with `normalizeImportPath` being used in two places

if I'm understanding this right

1. if starts with MAGIC prefix
    1. normalizeImportPath removes the prefix and adds the leading `/` (why?)
    2. resolve this path
    3. check that this path exists and return it
2. else
    1.  normalizeImportPath returns the same import
    2. resolve this path
    3. check that this path exists and return it
    4. if it doesn't, resolve the full path relative to the js file and return it

rollup/fix-css-import-paths-plugin.js
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    return files;
+};
+
+// Ensures rewritten imports are valid module specifiers.
+const toRelativeImportPath = (fromFile, toFile) => {
+    const relativePath = path
+        .relative(path.dirname(fromFile), toFile)
+        .split(path.sep)
```
qroll, line 37: is this to handle forward slash in Windows before passing to rollup?

qroll (approved):

rollup/fix-css-import-paths-plugin.js
ghazwan-gt: For Linaria hash result, yes, it is indeed unique. But this is a defensive approach to avoid doing anything when we found more than 1 match. If the basename lookup returns more than one CSS candidate, we skip rewriting rather than guessing and potentially pointing to the wrong file.

So yeah, normally you'd only see one match. The array is there for any edge cases. I guess it would be better to log that when it happens so we can detect and investigate those cases

rollup/fix-css-import-paths-plugin.js
ghazwan-gt: > check that this path exists and return it

It return null in this case. Mea path is already valid, so no rewrite is needed.

> removes the prefix and adds the leading 

We want it to be anchored to the root, not relative to the js file

rollup/fix-css-import-paths-plugin.js
ghazwan-gt: Yeah, should we remove it?

qroll: I see that this is a common pattern even in our node_modules, so I'm good

PR #1085, opened 2026-04-13
[BOOKINGSG-9188] Evaluate css in unit tests

qroll:
**Description of changes**

-   Link to [ticket](https://sgtechstack.atlassian.net/browse/BOOKINGSG-9188)
-   Extract Linaria styles with custom babel plugin and inject in `style` tag. This lets the jest-dom `expect.toBeVisible()` assertion pick up the evaluated style on elements

**Checklist**

-   [ ] ~~Added/updated tests~~

**Screenshots**

N/A


babel-plugins/babel-plugin-inject-linaria-css.js
```diff
@@ -0,0 +1,80 @@
+const { parseSync } = require("@babel/core");
+const { syncResolve } = require("@wyw-in-js/shared");
+const { transformSync } = require("@wyw-in-js/transform/lib/transform");
+const { TransformCacheCollection } = require("@wyw-in-js/transform/lib/cache");
+const {
+    loadWywOptions,
+} = require("@wyw-in-js/transform/lib/transform/helpers/loadWywOptions");
```
ghazwan-gt: Would it be alright to depend on these non-public API?

qroll: actually `loadWywOptions` and `TransformCacheCollection` are available through the main `@wyw-in-js/transform` import. I'll clean this up. good catch!

for the remaining one, `transformSync`:
1. it is not exposed, only the async `transform`, so we do have to import it directly from the internal path. I remember that using the async function caused an error seemingly because jest calls babel synchronously
2. I'm trying to mirror the babel plugin and it uses the sync version instead

^ but let me double check on the first point

qroll: 1. hmm seems like it doesn't work asynchronously. the visibility assertion fails when converting the visitor to async. also [ref](https://stackoverflow.com/questions/42009968/is-it-possible-to-run-async-code-in-a-babel-plugin-visitor)

ghazwan-gt (approved):

PR #1089, opened 2026-04-14
[BOOKINGSG-9207][GZ] migrate day cell to linaria

ghazwan-gt:
**Checklist**

-   [X] Added visual tests

**Additional information**

- Added a visual test for Week Calendar selection as it hasn't been covered by the previous work.


src/shared/internal-calendar/day-cell/day-cell.tsx
```diff
@@ -30,30 +22,40 @@ export const DayCell = ({
...
-    const today = dayjs().isSame(date, "day");
     const isFocused = focusDate ? focusDate.isSame(date, "day") : false;
     const defaultLabel = `${date.format("D MMMM YYYY dddd")}, ${
         disabled ? "Unavailable" : "Available"
     }`; // e.g. 1 January 2025 Tuesday, Unavailable
+    const isGridcellRole = role === "gridcell";
+    const isSelected =
+        labelType === "selected" || labelType === "selected-hover";
+    const labelText = label || defaultLabel;
+    const labelTypeAttr = disabled ? undefined : labelType;
```
zhaoyanxzy: i'm trying to see the scenario where it is `disabled` and `hidden` so as to match the style for `&[data-day-cell-disabled="true"][data-day-cell-label-type="hidden"]`. How can we trigger this?

zhaoyanxzy: i believe i was mistaken, we should just 
1. pass in `hidden` type by updating to `labelTypeAttr = disabled && labelType !== hidden ? undefined : labelType;`
2. don't think we need `&[data-day-cell-disabled="true"][data-day-cell-label-type="hidden"]` since there's already `data-day-cell-label-type="hidden"` styling

src/shared/internal-calendar/day-cell/day-cell.styles.ts
```diff
@@ -0,0 +1,218 @@
...
+    left: calc(50% - 1.25rem);
+    clip-path: inset(-3px -3px -3px 1.25rem);
+`;
+
+export const labelWrapper = css`
+    position: absolute;
+    top: 0;
+    bottom: 0;
+    z-index: 2;
+    cursor: not-allowed;
```
qroll: could `cursor: default` be the default instead? and check `true` and `false` specifically for the `pointer` and `not-allowed` style

same for `pointer-events` event in label. it seems more naturally to have these as defaults in hindsight

src/shared/internal-calendar/day-cell/day-cell.tsx
```diff
@@ -69,47 +71,66 @@ export const DayCell = ({
...
+            ></div>
+            <div
+                className={styles.rightCircle}
+                data-day-cell-type={circleRight}
+            />
+            <span
+                className={styles.labelWrapper}
+                data-day-cell-interactive={interactive}
+            >
+                <button
```
qroll, line 102: hmm why do we need to switch to a button element here from the original div?

src/shared/internal-calendar/day-cell/day-cell.tsx
```diff
@@ -30,30 +22,40 @@ export const DayCell = ({
...
-        if (isFocused && ref.current) {
-            ref.current?.focus();
+        if (isFocused) {
+            buttonRef.current?.focus();
         }
     }, [isFocused]);
 
+    useEffect(() => {
+        setIsToday(dayjs().isSame(date, "day"));
+    }, [date]);
```
qroll: no need for a `useEffect`, `useMemo` is sufficient

src/shared/internal-calendar/day-cell/day-cell.tsx
```diff
@@ -69,47 +71,66 @@ export const DayCell = ({
...
+    };
+
+    const labelContent = (
+        <>
+            {date.date()}
+            {currentDateIndicator && isToday && (
+                <div className={styles.indicator} />
+            )}
+        </>
+    );
```
qroll: could this be inlined with the main jsx?

src/shared/internal-calendar/day-cell/types.tsx
```diff
@@ -30,6 +30,7 @@ export interface CellStyleProps {
 }
 
 export interface DayCellProps extends CellStyleProps {
+    className?: string;
```
qroll: we don't need to style DayCell but since you already added this it's fine

e2e/tests/components/date-range-input/__screenshots__/chromium/DateRangeInput-Week-Default-Click-and-commit-selection-state.png
qroll, line 1: this is already covered by `DateRangeInput-Week-Selected-State-state.png` and `DateRangeInput-Week-selected-and-hover-same-state`

those capture the selected state (with and without hover)

qroll (changes requested):

src/shared/internal-calendar/day-cell/day-cell.styles.ts
qroll: ignore this, realised that `interactive` is handled weirdly in the first place. will stick with the existing logic

src/shared/internal-calendar/day-cell/day-cell.tsx
ghazwan-gt: Thanks for mentioning this. After further checking, you are correct, this is something that I missed
image

Disabed + hidden is not working as expected, since the hidden dates are shown as disabled.

src/shared/internal-calendar/day-cell/day-cell.tsx
ghazwan-gt: I was thinking that using `button` would be semantically correct here, hence the change. The default value of the `role` itself is button. Do you have any specific reason on why we want to keep using `div`?

qroll: find this kind of awkward

```
role={role === "button" ? undefined : role}
```

but I suppose we can simplify to
```
role={role}
```

we should clean up the enter/space key handling of `handleKeyDown` in the cell variants but that shouldn't be done now

src/shared/internal-calendar/day-cell/day-cell.styles.ts
qroll: much better, thank you!

src/shared/internal-calendar/day-cell/day-cell.tsx
qroll: would direct `data-day-cell-label-type={labelType}` work now since visibility has "precedence"?

e2e/tests/components/date-range-input/date-range-input.e2e.spec.ts
```diff
@@ -231,6 +231,38 @@ test.describe("DateRangeInput", () => {
                 /^\d{4}-\d{2}-\d{2}$/
             );
         });
+
+        test.describe("With maxDate constraint", () => {
```
qroll, line 235: good add, could we include 1 for minDate as well?

src/shared/internal-calendar/day-cell/day-cell.tsx
```diff
@@ -69,47 +72,60 @@ export const DayCell = ({
...
-            <RightCircle $type={circleRight} />
-            <LabelWrapper $interactive={interactive}>
-                <Label
-                    ref={ref}
+        <div className={styles.cell} aria-hidden={ariaHidden}>
+            <div className={styles.leftHalf} data-day-cell-type={bgLeft}></div>
+            <div
+                className={styles.leftCircle}
+                data-day-cell-type={circleLeft}
+            ></div>
```
qroll: keep the self closing tag if there are no children

e2e/tests/components/date-range-input/__screenshots__/chromium/DateRangeInput-Range-With-maxDate-constraint-Open--state.png
qroll, line 1: can be deleted

qroll (changes requested):

e2e/nextjs-app/src/app/components/date-range-input/range-with-min-date.e2e.tsx
```diff
@@ -0,0 +1,33 @@
...
+export default function Story() {
+    const [startDate, setStartDate] = useState<string>("");
+    const [endDate, setEndDate] = useState<string>("");
+
+    return (
+        <div className={styles.story}>
+            <DateRangeInput
+                data-testid="date-range-input"
+                variant="range"
+                maxDate="2026-04-10"
```
qroll: think you meant to set minDate here

ghazwan-gt: Lol caught red handed copy-pasted

qroll (approved):

PR #1101, opened 2026-04-16
[BOOKINGSG-9158][GZ] migrate shared dropdown wrapper to linaria

ghazwan-gt:
**Checklist**

-   [ ] Added visual tests

**Breaking changes**

List breaking style or api changes to be added to the migration docs

**Additional information**
- `wrapper` style is moved to `InputRangeSelect`


src/shared/dropdown-wrapper/element-with-dropdown.tsx
```diff
@@ -137,8 +141,10 @@ export const ElementWithDropdown = ({
     // =============================================================================
     // CONST, STATE, REF
     // =============================================================================
-    const theme = useContext(ThemeContext);
-    const mobileBreakpoint = V3_Breakpoint["sm-max"]({ theme });
+    const mobileBreakpointTokenValue = useDesignToken(Breakpoint["sm-max"]);
+    const mobileBreakpoint = parsePxOrRemValue(
+        mobileBreakpointTokenValue || DEFAULT_MOBILE_MAX_WIDTH_BREAKPOINT
+    );
```
qroll: we can use `useSafeMaxWidthMediaQuery` here. I see that it's only used to compute `isMobileScreen`

src/shared/dropdown-wrapper/element-with-dropdown.styles.ts
qroll, line 1: oh this is not even being referenced. iirc we had to change how heights are applied to the dropdown due to some unintended Floating UI behaviour so this became unused 🤔 we can clean this up

src/input-range-select/input-range-select.style.tsx
```diff
@@ -7,6 +8,18 @@ import { V3_Colour, V3_Spacing } from "../v3_theme";
...
+export const wrapper = css`
+    position: relative;
+    width: 100%;
+    min-height: 3rem;
+    height: 3rem; // Need this to persist the height when expanding or collapsing list
+
+    &[data-variant="small"] {
+        min-height: 2.5rem;
+        height: 2.5rem;
+    }
```
qroll: range select doesn't have a size variant

`StyledInputWrapper` also has a height set so I feel this wrapper isn't necessary anymore

qroll (changes requested):

qroll (approved):

PR #1104, opened 2026-04-16
[BOOKINGSG-9121][Nghi] migrate toggle and toggle-icon style to linaria

nghitb:
**Checklist**

-   [x] Added visual tests
-   [x] Listed breaking changes


src/toggle/toggle-icon.tsx
ghazwan-gt, line 1: Since the file is not being used anywhere else aside from `toggle`, should we include this there instead? @qroll

src/toggle/toggle.tsx
ghazwan-gt, line 1: Let's add a unit test for the component. 

@qroll should we also have a e2e story for this component as well?

src/toggle/toggle.styles.ts
```diff
@@ -0,0 +1,388 @@
...
+    flex-direction: column;
+    height: fit-content;
+    background: ${Colour.bg};
+
+    &:has(input:focus-visible) {
+        outline: 2px solid ${Colour["focus-ring"]};
+        outline-offset: 0;
+    }
+
+    &.toggleContainerDefault {
```
ghazwan-gt: Do you perhaps mean to refer to the other linaria class?

```suggestion
    &.${toggleContainerDefault} {
```

src/shared/toggle-icon/toggle-icon.styles.ts
```diff
@@ -0,0 +1,31 @@
+import { css } from "@linaria/core";
+
+import { Colour } from "../../theme";
+
+// =============================================================================
+// STYLING
+// =============================================================================
+export const WrapperBase = css`
```
qroll: all the css variables should be lowercased

src/shared/toggle-icon/toggle-icon.tsx
```diff
@@ -43,13 +44,17 @@ export const ToggleIcon = ({
...
     return (
-        <Wrapper
-            className={className}
-            $active={active}
-            $disabled={disabled}
+        <div
+            className={clsx(
+                active && styles.WrapperActive,
+                active && disabled && styles.WrapperActiveDisabled,
+                disabled && styles.WrapperDisabled,
```
qroll: minor thing but can we move `disabled` to the top? even though the class name order doesn't technically matter, this will match the css source order so we can tell what has precedence

src/shared/toggle-icon/toggle-icon.tsx
```diff
@@ -43,13 +44,17 @@ export const ToggleIcon = ({
...
+                styles.WrapperBase,
```
qroll: this should go first too

src/toggle/toggle.tsx
qroll: yep we should be adding the unit tests too

src/toggle/toggle-icon.tsx
qroll: yeah nothing has been reusing this for such a long time so I'm good to move this

src/toggle/toggle.styles.ts
```diff
@@ -0,0 +1,388 @@
...
+    display: flex;
+    flex-direction: column;
+    overflow-wrap: anywhere;
+    width: 100%;
+    overflow: hidden;
+`;
+
+export const toggleTextContainerDefault = css`
+    color: ${Colour.text};
+`;
```
qroll: this could be merged into textContainer since it's default

src/toggle/toggle.styles.ts
```diff
@@ -0,0 +1,388 @@
...
+    border: none;
+`;
+
+export const toggleInputDisabled = css`
+    cursor: not-allowed;
+`;
+
+export const toggleInputEnabled = css`
+    cursor: pointer;
+`;
```
qroll: redundant

src/toggle/toggle.styles.ts
```diff
@@ -0,0 +1,388 @@
...
+    color: ${Colour["text-primary"]};
+    ${Font["body-baseline-semibold"]}
+    display: flex;
+    align-items: center;
+    justify-content: flex-end;
+    border: none;
+    background-color: ${Colour.bg};
+    cursor: pointer;
+    padding: 0 1rem 0.6875rem 1rem;
+    padding-top: 0rem;
```
qroll: redundant. the `padding` property already specifies 0

src/toggle/toggle.styles.ts
```diff
@@ -0,0 +1,388 @@
+import { css } from "@linaria/core";
+
+import { WrapperBase } from "../shared/toggle-icon/toggle-icon.styles";
+import { Colour, Font, MediaQuery, Radius } from "../theme";
+
+// =============================================================================
```
qroll: these section headers are no longer relevant and should have been cleaned up by the prompt

src/toggle/toggle.tsx
```diff
@@ -132,6 +116,74 @@ export const Toggle = ({
         e.stopPropagation();
     };
 
+    const containerStateClass = (() => {
```
qroll: convert this to a named function `getContainerStateClass()` and put it under a HELPERS section. same for the other class name functions

src/toggle/toggle.tsx
```diff
@@ -183,28 +258,34 @@ export const Toggle = ({
...
+                <div
                     data-id="toggle-composite-children"
-                    $isFinalItem={!collapsible}
-                    $disabled={disabled}
+                    className={clsx(
+                        styles.children,
+                        !collapsible && styles.childrenIsFinalItem,
+                        disabled
+                            ? styles.childrenDisabled
+                            : selected && styles.childrenSelected
```
qroll: I see that selected state was not applied in V3, ~~will confirm with UX what the intended behaviour is~~

src/toggle/toggle.tsx
```diff
@@ -241,28 +327,35 @@ export const Toggle = ({
...
-                $indicator={indicator}
-                $styleType={styleType}
+                className={styles.headerContainer}
             >
-                <IndicatorLabelContainer $addPadding={removable}>
-                    <InputContainer
+                <div
+                    className={clsx(
+                        removable && styles.indicatorLabelContainerAddPadding,
+                        styles.indicatorLabelContainer
```
qroll: reverse order, with the default first

look through and update other instances accordingly in this file

src/toggle/toggle.tsx
```diff
@@ -271,61 +364,84 @@ export const Toggle = ({
...
-            </HeaderContainer>
+            </div>
         );
     };
 
     const renderErrorList = (errors: string[]) => {
         return (
             <>
-                <ErrorText weight="semibold" $disabled={disabled}>
+                <BodyMD
```
qroll: import from the base `../typography` path entrypoint and use the `Typography.*` namespace

src/toggle/toggle.tsx
```diff
@@ -271,61 +364,84 @@ export const Toggle = ({
...
+                    className={clsx(
+                        disabled && styles.errorTextDisabled,
+                        styles.errorText
+                    )}
+                >
                     Error
-                </ErrorText>
-                <ErrorList $disabled={disabled}>
+                </BodyMD>
+                <UnorderedList
```
qroll: same for this

src/toggle/toggle.tsx
```diff
@@ -335,21 +451,24 @@ export const Toggle = ({
...
+                        styles.errorContainer
+                    )}
                     onClick={handleExpandCollapseClick}
                     id={`${generatedId}-error-alert`}
                 >
-                    <AlertContainer
+                    <Alert
                         type={disabled ? "description" : "error"}
-                        className={className}
+                        className={clsx(styles.alertContainer, className)}
```
qroll: hmm. I don't think appending the same class name as the component's parent is intended here. we should remove this

qroll (changes requested):

src/toggle/toggle.tsx
qroll: for backwards compatibility let's remove the unused selected style

src/toggle/toggle.styles.ts
```diff
@@ -0,0 +1,388 @@
...
+
+export const toggleContainerNoBorderError = css`
+        border-color: ${Colour["border-error"]};
+
+        &:has(.${headerContainer}:hover) {
+            @media (pointer: fine) {
+                background: ${Colour["bg-hover-subtle"]};
+            }
+        }
+    }
```
zhaoyanxzy: indentation looks off, seems like there's an extra closing bracket

src/shared/toggle-icon/toggle-icon.tsx
zhaoyanxzy: i find this quite hard to read, can we have something like this:
```
const stateClass = disabled
     ? active
         ? styles.WrapperActiveDisabled
         : styles.WrapperDisabled
        .....

File: src/toggle/toggle.tsx [hunk 16]
```diff
@@ -158,12 +234,11 @@ export const Toggle = ({
...
 
         return (
-            <StyledToggleIcon
+            <ToggleIcon
                 type={toggleIconType}
                 active={selected}
                 disabled={disabled}
-                $selected={selected}
-                $disabled={disabled}
+                className={styledToggleIconStateClass}
```
zhaoyanxzy: why do we still need to pass in the className when we are deriving the styles based in `disabled` and `active` in `ToggleIcon`? The styles `toggleStyledToggleIconDisabledSelected`, `toggleStyledToggleIconDisabled`, `toggleStyledToggleIconSelected` and `toggleStyledToggleIconDefault` are already covered in ToggleIcon

src/toggle/toggle.tsx
nghitb: ok im gonna remove the alertContainer

src/toggle/toggle.styles.ts
nghitb: yeah my bad

src/shared/toggle-icon/toggle-icon.tsx
nghitb: sure, let me move to a helper function

src/toggle/toggle.tsx
nghitb: i was wondering this as well, the previous code still put in the selected and disabled, i just kept the remaining logic, but if u say so, im gonna delete those

src/toggle/toggle.tsx
nghitb: sure

src/toggle/toggle.tsx
nghitb: ok let me align all the classnames

src/toggle/toggle.tsx
nghitb: yeah just reply back to me what needs to be changed

src/toggle/toggle.tsx
nghitb: sure

src/toggle/toggle.styles.ts
nghitb: ok will delete those

src/toggle/toggle.styles.ts
nghitb: sure will delete padding-top

src/toggle/toggle.styles.ts
nghitb: will delete

src/toggle/toggle.styles.ts
nghitb: will merge

src/shared/toggle-icon/toggle-icon.tsx
nghitb: disabeld go first, got it

src/shared/toggle-icon/toggle-icon.tsx
nghitb: agreed

src/shared/toggle-icon/toggle-icon.styles.ts
nghitb: yes sir

src/toggle/toggle.styles.ts
nghitb: forgot to delete, will delete those

src/toggle/toggle.tsx
nghitb: will do

src/toggle/toggle.styles.ts
```diff
@@ -0,0 +1,344 @@
...
+    padding: 0.6875rem 1rem 0.6875rem 0.5rem;
+    border: none;
+    background: none;
+    cursor: pointer;
+`;
+
+export const removeButtonDisabled = css`
+    color: ${Colour["text-disabled"]};
+    cursor: not-allowed;
+`;
```
ghazwan-gt: This has the same content with `expandButtonDisabled` & `errorContainerDisabled`. How about we make this as a general disabled class name?

src/toggle/toggle.styles.ts
```diff
@@ -0,0 +1,344 @@
...
+    color: ${Colour.text};
+`;
+
+export const toggleTextContainerSelected = css`
+    color: ${Colour["text-selected"]};
+`;
+
+export const toggleTextContainerDisabled = css`
+    color: ${Colour["text-disabled"]};
+`;
```
ghazwan-gt: This is similar with `childrenDisabled`, `errorTextDisabled` & `errorListDisabled`. Lets have a general disabled text class name

src/toggle/toggle.styles.ts
```diff
@@ -0,0 +1,344 @@
...
+    color: ${Colour["text-selected"]};
+`;
+
+export const childrenDisabled = css`
+    color: ${Colour["text-disabled"]};
+`;
+
+export const errorText = css`
+    color: ${Colour["text-error"]};
+`;
```
ghazwan-gt: This is similar with `errorList`

src/toggle/toggle.styles.ts
```diff
@@ -0,0 +1,344 @@
...
+    flex-direction: column;
+    overflow-wrap: anywhere;
+    width: 100%;
+    overflow: hidden;
+    color: ${Colour.text};
+`;
+
+export const toggleTextContainerSelected = css`
+    color: ${Colour["text-selected"]};
+`;
```
ghazwan-gt, line 35: This is similar with `childrenSelected`

src/toggle/toggle.styles.ts
```diff
@@ -0,0 +1,344 @@
...
+    justify-content: center;
+`;
+
+export const useContentWidthContainer = css`
+    min-width: unset;
+`;
+
+export const toggleContainerNoBorderErrorDisabled = css`
+    border-color: ${Colour["border-error"]};
+`;
```
ghazwan-gt: This is similar with `toggleContainerDefaultErrorDisabled`

src/toggle/toggle.styles.ts
```diff
@@ -0,0 +1,340 @@
...
+
+export const toggleContainerNoBorderError = css`
+    border-color: ${Colour["border-error"]};
+
+    &:has(.${headerContainer}:hover) {
+        @media (pointer: fine) {
+            background: ${Colour["bg-hover-subtle"]};
+        }
+    }
```
ghazwan-gt: I see a pattern where we have two type of hover styling, for default and selected. Let's extract those to a class name and reference it

src/toggle/toggle.tsx
```diff
@@ -138,6 +143,63 @@ export const Toggle = ({
...
+        }
+
+        if (disabled) {
+            return selected
+                ? styles.toggleContainerDefaultDisabledSelected
+                : styles.toggleContainerDefaultDisabled;
+        }
+
+        if (selected) {
+            return styles.toggleContainerDefaultSelected;
```
ghazwan-gt: I find having `toggleContainerDefault` and `toggleContainerNoBorderDefault` confusing and oxymoron. Because we will have something like `toggleContainerDefaultErrorDisabled`, `toggleContainerDefaultSelected`, `toggleContainerNoBorderDisabledSelected`. 

I was thinking we can omit the `Default` part which then `toggleContainer` and `toggleContainerNoBorder` become the default name.

What do you think?

src/toggle/toggle.styles.ts
nghitb: there is no css called childrenSelected

src/toggle/toggle.tsx
nghitb: agree

src/toggle/toggle.styles.ts
nghitb: im confused, can u give an example

src/toggle/toggle.styles.ts
ghazwan-gt: Oh, it seems to be deleted on the latest changes. Resolving this

src/toggle/toggle.styles.ts
ghazwan-gt: Okay, so we can have something like this
```
const defaultHoverBlock = `
    &:has(.${headerContainer}:hover) {
        @media (pointer: fine) {
            background: ${Colour["bg-hover-subtle"]};
        }
    }
`;

const selectedHoverBlock = `
    &:has(.${headerContainer}:hover) {
        @media (pointer: fine) {
            background: ${Colour["bg-selected-hover"]};

            & .${textContainer} {
                color: ${Colour["text-selected-hover"]};
            }

            & .${toggleIconStyles.wrapperBase} {
                color: ${Colour["icon-selected-hover"]};
            }
        }
    }
`;
```

and then reference them like
```
export const toggleContainerNoBorderError = css`
    border-color: ${Colour["border-error"]};
    ${defaultHoverBlock}
`;

export const toggleContainerDefaultSelected = css`
    border-color: ${Colour["border-selected"]};
    background: ${Colour["bg-selected"]};
    ${selectedHoverBlock}
`;
```
But this makes them just like a regular string

ghazwan-gt: Or if we want to avoid having literal string, we can introduce a hover class name

src/toggle/toggle.styles.ts
nghitb: agree

src/toggle/toggle.styles.ts
nghitb: agree

src/toggle/toggle.styles.ts
nghitb: if we change to a hover classname, it will make the getContainerStateClass more complicated

ghazwan-gt: We don't have to touch `getContainerStateClass`. Can instead introduce a new helper for the hover
```
    const getContainerHoverClass = (() => {
        if (disabled) {
            return undefined;
        }

        if (!error && selected) {
            return styles.toggleContainerHoverSelected;
        }

        return styles.toggleContainerHoverDefault;
    })();
   ```

What do you think?

File: e2e/nextjs-app/src/app/components/toggle/with-composite.e2e.tsx [hunk 24]
```diff
@@ -0,0 +1,67 @@
...
+                </Toggle>
+            </div>
+
+            {/* With errors (collapsed state shows error) */}
+            <div className="story-row-container">
+                <Toggle
+                    type="checkbox"
+                    indicator
+                    checked
+                    error
```
ghazwan-gt: `error` and `checked` are mutually exclusive

e2e/tests/components/toggle/toggle.e2e.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    test.describe(() => {
+        test.beforeEach(async ({ story }) => {
+            await story.init("with-sublabel", { mode: "dark" });
+        });
+
+        test("With sub label (dark mode)", async ({ story }) => {
+            await compareScreenshot(story, "mount");
+        });
+    });
```
ghazwan-gt, line 158: I don't we need dark mode here, already covered by base variants

e2e/tests/components/toggle/toggle.e2e.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    test.describe(() => {
+        test.beforeEach(async ({ story }) => {
+            await story.init("with-composite", { mode: "dark" });
+        });
+
+        test("With composite section (dark mode)", async ({ story }) => {
+            await compareScreenshot(story, "mount");
+        });
+    });
```
ghazwan-gt, line 196: I don't we need dark mode here, already covered by base variants

e2e/tests/components/toggle/toggle.e2e.spec.ts
```diff
@@ -0,0 +1,134 @@
+import { test as base, Page } from "@playwright/test";
```
ghazwan-gt: Add a case for keyboard navigation and also check for `focusableWhenDisabled`

src/toggle/toggle.styles.ts
nghitb: sure, will do

src/toggle/toggle-icon.tsx
```diff
@@ -24,6 +25,22 @@ export const ToggleIcon = ({
...
+        if (disabled) {
+            return styles.wrapperDisabled;
+        }
+
+        if (active) {
+            return styles.wrapperActive;
+        }
+
+        return undefined;
+    })();
```
qroll: instead of calling this immediately here, do the call in className instead. same for the functions toggle file

```
        <div
            className={clsx(
                styles.wrapperBase,
                getWrapperStateClass(),
                className
            )}
```

src/toggle/toggle.tsx
```diff
@@ -1,28 +1,17 @@
...
 import { ChevronUpIcon } from "@lifesg/react-icons/chevron-up";
+import clsx from "clsx";
 import { useEffect, useMemo, useRef, useState } from "react";
 
+import { Alert } from "../alert";
 import { Markup } from "../markup";
-import type { ToggleIconType } from "../shared/toggle-icon/toggle-icon";
+import { UnorderedList } from "../text-list/unordered-list";
+import { useApplyStyle } from "../theme";
+import * as Typography from "../typography/typography";
```
qroll:
```suggestion
import * as Typography from "../typography";
```

src/toggle/toggle.tsx
```diff
@@ -1,28 +1,17 @@
 import { ChevronDownIcon } from "@lifesg/react-icons/chevron-down";
 import { ChevronUpIcon } from "@lifesg/react-icons/chevron-up";
+import clsx from "clsx";
 import { useEffect, useMemo, useRef, useState } from "react";
 
+import { Alert } from "../alert";
 import { Markup } from "../markup";
-import type { ToggleIconType } from "../shared/toggle-icon/toggle-icon";
+import { UnorderedList } from "../text-list/unordered-list";
```
qroll:
```suggestion
import { TextList } from "../text-list";
```

access as TextList.UnorderedList

src/toggle/toggle.tsx
qroll: ehhh alertContainer contributes to the styling so it shouldn't be removed. it should be `className` that is removed

e2e/nextjs-app/src/app/components/toggle/with-composite.e2e.tsx
qroll: theoretically an item could be selected but with error. just that we don't show this in storybook. I think it'll be good to cover this in the e2e tests

e2e/tests/components/toggle/toggle.e2e.spec.ts
qroll: composite section is a variant with significant layout differences so it'll be good to cover this in dark mode too

e2e/nextjs-app/src/app/components/toggle/base-variants.e2e.tsx
qroll, line 1: instead of `base-variants` and `states`, I would suggest to organise like this instead:

each file is a type:
- checkbox: bordered and no-border, with and without indicator, all checked/disabled/error states
- radio: as above
- yes: as above
- no: as above

so there would be 4 variants x 6 states in each file

qroll (changes requested):

e2e/nextjs-app/src/app/components/toggle/with-composite.e2e.tsx
ghazwan-gt: I see, checked on [Figma](https://www.figma.com/design/tHmHQNszVzvIamtZWixjgR/%F0%9F%8C%B1-Flagship-V3-Component-Kit?node-id=3050-2793&m=dev) and we do have selected and error for composite. Resolving this one

e2e/tests/components/toggle/toggle.e2e.spec.ts
ghazwan-gt: Okay, resolving this one

src/toggle/toggle.tsx
```diff
@@ -1,28 +1,17 @@
...
 import { ChevronUpIcon } from "@lifesg/react-icons/chevron-up";
+import clsx from "clsx";
 import { useEffect, useMemo, useRef, useState } from "react";
 
+import { Alert } from "../alert";
 import { Markup } from "../markup";
-import type { ToggleIconType } from "../shared/toggle-icon/toggle-icon";
+import { TextList } from "../text-list";
+import { useApplyStyle } from "../theme";
+import * as Typography from "../typography";
```
ghazwan-gt:
```suggestion
import { Typography } from "../typography";
```

src/toggle/toggle.styles.ts
```diff
@@ -0,0 +1,315 @@
...
+export const useContentWidthContainer = css`
+    min-width: unset;
+`;
+
+export const colorBorderError = css`
+    border-color: ${Colour["border-error"]};
+`;
+
+export const toggleContainerNoBorderError = css`
+    border-color: ${Colour["border-error"]};
```
ghazwan-gt: It says `NoBorder` but we have a border color here

src/toggle/toggle.tsx
```diff
@@ -138,6 +143,63 @@ export const Toggle = ({
...
+                return styles.toggleContainerNoBorderSelected;
+            }
+
+            return styles.toggleContainerNoBorder;
+        }
+
+        if (error) {
+            return disabled
+                ? styles.colorBorderError
+                : styles.toggleContainerError;
```
ghazwan-gt, line 176:
```suggestion
            return clsx(
                styles.colorBorderError,
                !disabled && styles.toggleContainerError
            );
```

And remove border color from `toggleContainerError`

src/toggle/toggle-icon.tsx
qroll: bumps, do refactor for the toggle file as well

e2e/tests/components/toggle/toggle.e2e.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    story: async ({ page }, use) => {
+        const story = new StoryPage(page);
+        await use(story);
+    },
+});
+
+test.describe("Toggle", () => {
+    // -------------------------------------------------------------------------
+    // Checkbox
+    // -------------------------------------------------------------------------
```
qroll: the section headers are not needed. are they being added by the e2e prompt?

e2e/nextjs-app/src/app/components/toggle/checkbox.e2e.tsx
qroll, line 1: commit the snapshots after they are generated

qroll (changes requested):

e2e/tests/components/toggle/toggle.e2e.spec.ts
nghitb: yes they are

qroll: will take a look. resolving this comment

qroll (approved):

PR #1107, opened 2026-04-17
[BOOKINGSG-9157] refactor: migrate dropdown-list to linaria and v4 tokens

ziggyzet:
**Checklist**

-   [X] Added visual tests

**Additional information**
- For some reason, the tokens doesn't work if it's formatted in the following way: `calc(var(<TOKEN>))`
- Not sure if applying style for Virtuoso components using the `useApplyStyles` hook is a must
- Updated the usage in `language-switcher`
- There was an error before the migration regarding the `Virtuoso` component while running the unit test. It doesn't fail the tests and haven't made changes to fix it.


ghazwan-gt:
image
There's a different result on `RangeSelect`

src/shared/dropdown-list/dropdown-list.styles.ts
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    max-width: var(--fds-internal-dropdown-list-container-available-width);
+    min-width: min(
+        23rem,
+        var(--fds-internal-dropdown-list-container-available-width)
+    );
+    max-height: min(
+        27rem,
+        var(--fds-internal-dropdown-list-container-available-height, 9999px)
```
ghazwan-gt: Can check whether this is related to the one in `ElementWithDropdown:191`

src/shared/dropdown-list/dropdown-list.tsx
```diff
@@ -404,26 +395,46 @@ const DropdownListInner = <T, V>(
         setFocusedIndex,
     ]);
 
+    // =========================================================================
+    // APPLY STYLES
+    // =========================================================================
+
+    useApplyStyle(nodeRef, {
+        ...floatingStyles,
+        ...containerWidthStyle,
```
ghazwan-gt, line 410: `useApplyStyle` will applied the value as string  and `containerWidthStyle` seems to be a pure number, this will make the resulting value missing unit.

src/shared/dropdown-list/dropdown-list.tsx
```diff
@@ -404,26 +395,46 @@ const DropdownListInner = <T, V>(
         setFocusedIndex,
     ]);
 
+    // =========================================================================
+    // APPLY STYLES
+    // =========================================================================
+
+    useApplyStyle(nodeRef, {
+        ...floatingStyles,
```
ghazwan-gt: Inside `floatingStyles`, we have `zIndex` and `useApplyStyle` currently not able to map camelCase `zIndex` to `z-index`. We likely don't have it in the result

src/shared/dropdown-list/dropdown-search.tsx
```diff
@@ -22,26 +20,39 @@ const Component = (
...
+                className={clsx(
+                    styles.searchBox,
+                    variant === "small" && styles.searchBoxVariantSmall
+                )}
+            >
+                <MagnifierIcon aria-hidden className={styles.searchIcon} />
+                <BasicInput
                     ref={ref}
                     value={value}
-                    variant={variant}
```
ghazwan-gt, line 31: Shouldn't we keep this to handle small variant font?

src/shared/dropdown-list/dropdown-label.styles.ts
```diff
@@ -0,0 +1,137 @@
...
+
+// -----------------------------------------------------------------------------
+// SECONDARY TEXT
+// -----------------------------------------------------------------------------
+export const secondaryText = css`
+    ${tokens.secondaryText.maxLines}: 2;
+    color: ${Colour["text-subtlest"]};
+    width: 100%;
+    overflow-wrap: break-word;
+`;
```
ghazwan-gt, line 57: `maxLines` token, `width`, and `overflow-wrap` can be extracted to a general text class name

src/shared/dropdown-list/dropdown-label.styles.ts
```diff
@@ -0,0 +1,137 @@
...
+    width: 100%;
+    height: 1lh;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    overflow: hidden;
+    direction: rtl;
+    text-align: right;
+`;
+
+export const truncateSecondLineSingle = css`
```
ghazwan-gt: Similar to `truncateFirstLineSingle`, can be one single class name instead

src/shared/dropdown-list/dropdown-list.styles.ts
```diff
@@ -0,0 +1,210 @@
...
+    ${tokens.xSpacing}: initial;
+
+    border: ${Border["width-010"]} ${Border["solid"]} ${Colour["border"]};
+    border-radius: ${Radius["sm"]};
+    background: ${Colour["bg"]};
+
+    ${tokens.xSpacing}: 0px;
+    ${tokens.availableWidth}: calc(
+        100vw - var(${tokens.xSpacing}) * 2
+    );
```
ghazwan-gt, line 36: Why do this when we already set it as initial just a few lines before?

src/shared/dropdown-list/dropdown-list.styles.ts
```diff
@@ -0,0 +1,210 @@
...
+            "xs-margin"
+        ]};
+    }
+
+    ${MediaQuery.MaxWidth.sm} {
+        --fds-internal-dropdown-list-container-x-spacing: ${Breakpoint[
+            "sm-margin"
+        ]};
+        max-height: 15rem;
+    }
```
ghazwan-gt: Max Width media queries should be in descending order

src/shared/dropdown-list/dropdown-list.styles.ts
```diff
@@ -0,0 +1,210 @@
...
+    Font,
+    MediaQuery,
+    Radius,
+    Spacing,
+} from "../../theme";
+
+export const tokens = {
+    xSpacing: "--fds-internal-dropdown-list-container-x-spacing",
+    availableWidth: "--fds-internal-dropdown-list-container-available-width",
+    availableHeight: "--fds-internal-dropdown-list-container-available-height",
```
ghazwan-gt:
```suggestion
    xSpacing: "--fds-internal-dropdownList-container-xSpacing",
    availableWidth: "--fds-internal-dropdownList-container-availableWidth",
    availableHeight: "--fds-internal-dropdownList-container-availableHeight",
```

src/shared/dropdown-list/dropdown-list.styles.ts
```diff
@@ -0,0 +1,210 @@
...
+    flex-shrink: 0;
+    height: 1lh;
+    width: 1rem;
+    color: ${Colour["icon-selected"]};
+`;
+
+export const unselectedIndicator = css`
+    flex-shrink: 0;
+    height: 1lh;
+    width: 1rem;
```
ghazwan-gt: Can extract these three attributes as a general indicator class name, then have another variants for adding the color

src/shared/dropdown-list/dropdown-list.styles.ts
```diff
@@ -0,0 +1,210 @@
...
+    ${Font["body-md-semibold"]}
+    padding: ${Spacing["spacing-8"]};
+`;
+
+export const tryAgainButton = css`
+    cursor: pointer;
+    overflow: hidden;
+    color: ${Colour["text-primary"]};
+    font-size: inherit;
+    ${Font["body-baseline-semibold"]}
```
ghazwan-gt: Similar to `selectAllButton`, can have another class name for padded one

src/shared/dropdown-list/dropdown-list.tsx
```diff
@@ -122,6 +119,12 @@ const DropdownListInner = <T, V>(
         !!selectedItems &&
         selectedItems?.length === maxSelectable;
 
+    const containerWidthStyle: React.CSSProperties = width
+        ? { width }
+        : matchElementWidth && elementWidth
+        ? { width: elementWidth }
+        : {};
```
ghazwan-gt: Can we avoid having a nested ternary, it has bad readability

src/shared/dropdown-list/dropdown-list.tsx
```diff
@@ -464,9 +476,12 @@ const DropdownListInner = <T, V>(
                     }}
                     role="option"
                     tabIndex={active ? 0 : -1}
-                    $active={active}
-                    $selected={selected}
-                    $disabled={!selected && hasSelectedMax}
+                    className={clsx(
+                        styles.listItem,
+                        disabled && styles.listItemDisabled,
+                        active && selected && styles.listItemActiveSelected,
```
ghazwan-gt: Can't we have `selected` as the only condition?

src/shared/dropdown-list/dropdown-list.tsx
```diff
@@ -464,9 +476,12 @@ const DropdownListInner = <T, V>(
                     }}
                     role="option"
                     tabIndex={active ? 0 : -1}
-                    $active={active}
-                    $selected={selected}
-                    $disabled={!selected && hasSelectedMax}
+                    className={clsx(
+                        styles.listItem,
+                        disabled && styles.listItemDisabled,
```
ghazwan-gt: Should also check when it is not active.

On this limit selection example of MultiSelect, disabled option still have that active background on hover
Image

src/shared/dropdown-list/dropdown-list.tsx
```diff
@@ -122,6 +119,12 @@ const DropdownListInner = <T, V>(
         !!selectedItems &&
         selectedItems?.length === maxSelectable;
 
+    const containerWidthStyle: React.CSSProperties = width
+        ? { width }
+        : matchElementWidth && elementWidth
+        ? { width: elementWidth }
```
zhaoyanxzy: `elementWidth` is just a number, need to have units appended

src/shared/dropdown-list/dropdown-label.styles.ts
```diff
@@ -0,0 +1,137 @@
...
+
+// =============================================================================
+// STYLING
+// =============================================================================
+
+// -----------------------------------------------------------------------------
+// PRIMARY TEXT
+// -----------------------------------------------------------------------------
+export const primaryText = css`
+    ${tokens.primaryText.maxLines}: 2;
```
qroll: move this to primaryTextTruncateEnd to keep the variables together

do the same for secondary text

src/shared/dropdown-list/dropdown-label.tsx
```diff
@@ -57,16 +60,11 @@ export const DropdownLabel = ({
...
-            // best-effort attempt to check if multi-line text will overflow,
-            // rendering an actual element in the DOM might be more accurate
-            // but might not be performant for large lists
             const textWidth = StringHelper.getTextWidth(
                 displayText,
                 `${fontSize} '${fontFamily}'`
             );
 
-            // there's less space than expected due to word breaks, so an
-            // arbitary offset is applied
```
qroll, line 69: restore this comment

I'm seeing a few instances of unrelated comments being removed, please look through and restore them

src/shared/dropdown-list/dropdown-list.styles.ts
```diff
@@ -0,0 +1,210 @@
...
+    border-radius: ${Radius["sm"]};
+    background: ${Colour["bg"]};
+
+    ${tokens.xSpacing}: 0px;
+    ${tokens.availableWidth}: calc(
+        100vw - var(${tokens.xSpacing}) * 2
+    );
+
+    ${MediaQuery.MaxWidth.xxs} {
+        --fds-internal-dropdown-list-container-x-spacing: ${Breakpoint[
```
qroll: can reference the tokens here instead of hardcoding the css variable name

src/shared/dropdown-list/dropdown-list.styles.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    ${MediaQuery.MaxWidth.sm} {
+        --fds-internal-dropdown-list-container-x-spacing: ${Breakpoint[
+            "sm-margin"
+        ]};
+        max-height: 15rem;
+    }
+
+    max-width: var(--fds-internal-dropdown-list-container-available-width);
+    min-width: min(
```
qroll: in the original logic, `customWidth` will not apply `min-width`. you'll need to handle this scenario

src/shared/dropdown-list/dropdown-list.styles.ts
```diff
@@ -0,0 +1,210 @@
...
+    width: 1em;
+`;
+
+export const spinner = css`
+    margin-right: ${Spacing["spacing-8"]};
+    color: ${Colour["icon"]};
+`;
+
+export const noResultDescContainer = css`
+    --horizontal-padding: ${Spacing["spacing-16"]};
```
qroll: what is the use of this variable?

src/shared/dropdown-list/dropdown-list.tsx
qroll: we could add a helper that computes the class name based on the original logic since these are mutually exclusive

src/shared/dropdown-list/dropdown-label.styles.ts
```diff
@@ -0,0 +1,137 @@
...
+export const label = css`
+    text-align: left;
+    width: 100%;
+    overflow: hidden;
+    overflow-wrap: break-word;
+`;
+
+export const labelVariantDefault = css`
+    ${Font["body-baseline-regular"]}
+`;
```
qroll: seeing some inconsistencies with how default is handled, we can standardise this

the default can be merged into the base class to reduce the number of classes we have to manage

src/shared/dropdown-list/nested-dropdown-list.styles.ts
```diff
@@ -0,0 +1,95 @@
...
+    background: ${Colour["bg-hover"]};
+`;
+
+export const indent = css`
+    ${tokens.level}: 0;
+    height: 1px;
+    width: calc((1lh + ${Spacing["spacing-8"]}) * var(${tokens.level}));
+`;
+
+const expandButtonTransition = `transform ${Motion["duration-350"]} ${Motion["ease-standard"]}`;
```
qroll: could this be inlined? doesn't seem like anythign else is reference this

qroll (changes requested):

qroll:
the truncation is not taking effect. the `labelInline` `display: inline` style is overriding the required `display: -webkit-box` from line clamp

tests/shared/dropdown-list/dropdown-list.spec.tsx
```diff
@@ -0,0 +1,309 @@
...
+        global.ResizeObserver = jest.fn().mockImplementation(() => ({
+            observe: jest.fn(),
+            unobserve: jest.fn(),
+            disconnect: jest.fn(),
+        }));
+    });
+
+    // =========================================================================
+    // Single selection
+    // =========================================================================
```
qroll: can remove these headers since they are already grouped by `describe` blocks

tests/shared/dropdown-list/dropdown-list.spec.tsx
```diff
@@ -0,0 +1,309 @@
...
+    // Keyboard navigation
+    // =========================================================================
+    describe("keyboard navigation", () => {
+        it("should move focus down on ArrowDown", async () => {
+            const user = userEvent.setup();
+            renderList();
+
+            const options = screen.getAllByRole("option");
+            options[0].focus();
```
qroll: the dropdown already handles focus by default, instead of programmatically focusing the option, wait for the first option to receive focus

refer to `nested-dropdown-list.spec.tsx` `should expand a collapsed category with right arrow key`

e2e/tests/components/dropdown-list/__screenshots__/chromium/DropdownList-Loading---dark-mode-load-states-dark.png
qroll, line 1: this doesn't look right, we should apply a text color for the result container

e2e/tests/components/dropdown-list/dropdown-list.e2e.spec.ts
qroll, line 1: include a test case for with search bar, default and small sizes

qroll (changes requested):

src/shared/dropdown-list/dropdown-label.styles.ts
ziggyzet: since we're moving the maxLines, I won't move the rest here if that's okay.

src/shared/dropdown-list/dropdown-list.styles.ts
ziggyzet: not sure what the action for this would be for now, because in that code, i didn't see anything that could be applied here since the token is different

src/shared/dropdown-list/dropdown-list.styles.ts
```diff
@@ -0,0 +1,201 @@
...
+    Breakpoint,
+    Colour,
+    Font,
+    MediaQuery,
+    Radius,
+    Spacing,
+} from "../../theme";
+
+export const tokens = {
+    xSpacing: "--fds-internal-dropdownList-container-x-spacing",
```
qroll: follow the naming convention for variables. update the other tokens here

```suggestion
    xSpacing: "--fds-internal-dropdownList-container-xSpacing",
```

src/shared/dropdown-list/dropdown-list.styles.ts
```diff
@@ -0,0 +1,201 @@
...
+     */
+    containerWidth: "--fds-internal-dropdownList-container-width",
+    containerMinWidth: "--fds-internal-dropdownList-container-min-width",
+} as const;
+
+// -----------------------------------------------------------------------------
+// MAIN STYLES
+// -----------------------------------------------------------------------------
+export const container = css`
+    ${tokens.availableHeight}: initial;
```
qroll: put this together with the other tokens

src/shared/dropdown-list/dropdown-list.tsx
qroll: think we are trying to preserve the mutually exclusive logic here, but it will be more readable if we use the `||` condition to short circuit here

src/shared/dropdown-list/dropdown-label.styles.ts
qroll: to update `containerVariantDefault` as well

e2e/tests/components/dropdown-list/dropdown-list.e2e.spec.ts
qroll: resolving since the internal component can't be tested in e2e

qroll (changes requested):

src/shared/dropdown-list/dropdown-label.styles.ts
qroll: some duplication is okay if the styles are very small and only used twice

src/shared/dropdown-list/dropdown-list.tsx
qroll: for this, let's pass the floating styles directly to the component through the inline style tag

CSP shouldn't be an issue since the component can't mount directly at the moment. we'll figure this out later if needed

src/shared/dropdown-list/dropdown-list.styles.ts
qroll: DropdownList and ElementWithDropdown need to share the same variables

src/shared/dropdown-list/dropdown-list.tsx
```diff
@@ -464,9 +491,14 @@ const DropdownListInner = <T, V>(
...
-                    $active={active}
-                    $selected={selected}
-                    $disabled={!selected && hasSelectedMax}
+                    className={clsx(
+                        styles.listItem,
+                        (disabled && styles.listItemDisabled) ||
+                            (active &&
+                                selected &&
+                                styles.listItemActiveSelected) ||
+                            (active && !selected && styles.listItemActive)
```
qroll: can now be simplified to

```suggestion
                            (active && styles.listItemActive)
```

src/shared/dropdown-list/dropdown-list.styles.ts
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    &::-webkit-scrollbar-thumb {
+        background: ${Colour["bg-inverse-subtlest"]};
+        border: 5px solid transparent;
+        border-radius: ${Radius["full"]};
+        background-clip: padding-box;
+    }
+
+    ${Font["body-baseline-regular"]}
```
qroll: nit: move this up with the border and background styles. it's easy to miss out when placed below the complex or secondary styling

src/shared/dropdown-list/dropdown-list.styles.ts
qroll: outdated since the tokens should sit with ElementWithDropdown

src/shared/dropdown-list/dropdown-list.styles.ts
```diff
@@ -0,0 +1,199 @@
...
+} from "../../theme";
+
+export const tokens = {
+    availableWidth: "--fds-internal-dropdownList-container-availableWidth",
+    availableHeight: "--fds-internal-dropdownList-container-availableHeight",
+    /**
+     * Width logic (mirrors old styled-components props):
+     * - customWidth (string): set containerWidth + set containerMinWidth to 0px
+     * - width (number in px): set containerWidth to `${width}px` only
+     * - default: set neither (width=auto, min-width uses fallback)
```
qroll: this comment is no longer applicable since both variables are now being set by the parent

src/shared/dropdown-list/dropdown-label.tsx
```diff
@@ -84,10 +88,13 @@ export const DropdownLabel = ({
         [hasExceededContainer, sublabel]
     );
 
-    // css cannot truncate inline elements so force the display to render
-    // them separately
+    // there's less space than expected due to word breaks, so an
+    // arbitary offset is applied
```
qroll: wrong comment

src/shared/dropdown-list/dropdown-label.tsx
```diff
@@ -119,42 +126,78 @@ export const DropdownLabel = ({
...
-                $disabled={disabled}
-                $truncateType={truncationType}
+                className={clsx(
+                    styles.primaryText,
+                    bold && styles.primaryTextBold,
+                    disabled && styles.primaryTextDisabled,
+                    !disabled && selected && styles.primaryTextSelected,
+                    itemDisplayType !== "next-line" &&
+                        truncationType !== "end" &&
+                        styles.textLabelInline,
```
qroll: 1. rename `textLabelInline` to `primaryTextLabelInline`
2. move above `primaryTextTruncateEnd`
3. move `secondaryTextLabelInline` above `secondaryTextTruncateEnd`
4. this should be able to be simplified to
```suggestion
                    itemDisplayType === "inline" &&
                        styles.textLabelInline,
```
5. same for secondary text

qroll (changes requested):

ghazwan-gt (approved):

src/shared/dropdown-wrapper/element-with-dropdown.tsx
```diff
@@ -188,14 +189,14 @@ export const ElementWithDropdown = ({
                     ) {
                         // reset this so that dropdown can expand as needed
                         elements.floating.style.setProperty(
-                            "--available-height",
+                            `${dropdownWrapperStyles.tokens.availableHeight}`,
```
qroll: nit: no need to put in template literal again. update for both instances

qroll (approved):

PR #1114, opened 2026-04-22
[BOOKINGSG-9209][GZ] migrate shared timepicker dropdown to linaria

ghazwan-gt:
**Checklist**

    -   [ ] `className` is chained correctly with `clsx`
-   [ ] Added visual tests


src/shared/timepicker-dropdown/timepicker-dropdown.styles.ts
```diff
@@ -0,0 +1,154 @@
...
+export const controlSection = css`
+    display: flex;
+    flex-direction: row;
+    justify-content: flex-end;
+    margin-top: ${Spacing["spacing-16"]};
+    gap: ${Spacing["spacing-8"]};
+
+    ${MediaQuery.MaxWidth.xxs} {
+        flex-direction: column;
+        margin-top: ${Spacing["spacing-32"]};
```
qroll, line 52: looking at the [a11y specs](https://www.figma.com/design/wDAxbRmlq4xOn4UJ8zPuEF/%F0%9F%A7%91%F0%9F%8F%BB%E2%80%8D%F0%9F%A6%AF-Flagship-A11y-Annotations?node-id=351-5418&p=f&m=dev), the Done button should receive focus first, then Cancel

so the treatment will need to be reversed - the HTML DOM order has to be switched around and we can use flex to manage the visual order

tests/shared/timepicker-dropdown/timepicker-dropdown.spec.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+    const renderDropdown = (
+        props?: Partial<ComponentProps<typeof TimepickerDropdown>>
+    ) => {
+        return render(
+            <TimepickerDropdown
+                id={ID}
+                show
+                format="12hr"
+                value="03:00AM"
```
qroll: we can use `value=""` as the default and explicitly test the behaviour for prefilled values. we should verify that the inputs were correctly populated with the hours and minutes for 12hr and 24hr format

tests/shared/timepicker-dropdown/timepicker-dropdown.spec.tsx
```diff
@@ -0,0 +1,187 @@
+import { render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import type { ComponentProps } from "react";
+import { TimepickerDropdown } from "src/shared/timepicker-dropdown/timepicker-dropdown";
+
+describe("TimepickerDropdown", () => {
```
qroll, line 6: could we add a test for the special character handling?

qroll (changes requested):

tests/shared/timepicker-dropdown/timepicker-dropdown.spec.tsx
```diff
@@ -115,6 +115,53 @@ describe("TimepickerDropdown", () => {
...
+
+        const numberKeyEvent = new KeyboardEvent("keydown", {
+            key: "1",
+            code: "Digit1",
+            bubbles: true,
+            cancelable: true,
+        });
+
+        const isSpecialKeyAtPrevented =
+            !hourInput.dispatchEvent(specialKeyAtEvent);
```
qroll: is this method of dispatching the keyboard event because we are listening on `addEventListener` rather than through `onKeyDown`? userEvent doesn't work with this?

ghazwan-gt: Yes, `dispatchEvent` is used to get a clear view whether the prevention is working. If you'd like, we can do with `userEvent` to check what the user see in the input value.

qroll: `userEvent` would be preferred. the user-facing end state that we can assert on is that the input value was not updated

though I'm not sure if `userEvent` works so can check that it doesn't have a false negative

ghazwan-gt: Alright, will try with `userEvent`

qroll (approved):

PR #1115, opened 2026-04-22
[BOOKINGSG-9243] migrate menu to linaria

ghazwan-gt:
**Checklist**

-   [X] Added visual tests


e2e/nextjs-app/src/app/components/menu/default.e2e.tsx
```diff
@@ -0,0 +1,55 @@
...
+                when the menu has limited width
+            </Menu.Link>
+        </Menu.Section>
+    </Menu.Content>
+);
+
+export default function Story() {
+    return (
+        <div className="story-column-container">
+            <div className={styles.defaultRow}>
```
zhaoyanxzy: `styles` only have `.placement-item`?

e2e/tests/components/menu/menu.e2e.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+    };
+
+    constructor(page: Page) {
+        super(page);
+
+        this.locators = {
+            triggerDefault: page.getByRole("button", { name: "Open menu" }),
+            contentDefault: page
+                .getByRole("list", { name: "Actions" })
+                .locator("xpath=.."),
```
zhaoyanxzy: Qn: why are we traversing by xpath and not use `getByTestId`? Furthermore there is already the data-testid `menu-content-default`

src/menu/menu-content.styles.ts
```diff
@@ -0,0 +1,66 @@
...
+    border-radius: ${Radius["md"]};
+    border: ${Border["width-010"]} ${Border["solid"]} ${Colour["border"]};
+    background: ${Colour["bg"]};
+    box-shadow: ${Shadow["md-subtle"]};
+
+    ${tokens.panel.maxHeight}: initial;
+    ${tokens.panel.overflow}: initial;
+
+    --x-spacing: 0px;
+    --available-width: calc(100vw - var(--x-spacing) * 2);
```
qroll: let's move these to the tokens declaration too and standardise the variable naming

src/menu/menu-section.styles.ts
```diff
@@ -0,0 +1,19 @@
+import { css } from "@linaria/core";
+
+import { Border, Colour, Spacing } from "../theme/tokens";
+
+export const section = css`
+    padding: ${Spacing["spacing-8"]} 0;
+    margin: 0;
+    list-style: none;
+
+    &.menuSectionDivider {
```
qroll: should be in a standalone class

also can rename to `sectionWithDivider`, as `divider` makes it sound like it's an actual element

e2e/tests/components/menu/menu.e2e.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+            linkLong: page.getByRole("link", {
+                name: "This is a long menu link title that should clamp across lines when the menu has limited width",
+            }),
+            contentOverflow: page
+                .getByRole("list", { name: "Overflow list" })
+                .locator("xpath=.."),
+        };
+    }
+
+    public triggerForPlacement(position: PlacementExpectation) {
```
qroll: can still be part of the locators, simple functions are alright (button names could be moved out)

```
story.locators.triggerForPlacement
story.locators.contentForPlacement
```

else would prefer to name as `getX`

e2e/tests/components/menu/__screenshots__/chromium/Menu-Default-Mount--state.png
qroll, line 1: we should also capture other variants like menu section without label and menu section without divider

qroll (changes requested):

e2e/nextjs-app/src/app/components/menu/default.e2e.tsx
ghazwan-gt: My bad, forgot to cleanup the stylings. Will update it

e2e/tests/components/menu/menu.e2e.spec.ts
ghazwan-gt: I was trying to avoid using `getByTestId` but now that I see it again, it is better to use it in this case

qroll:
can I trouble you to include a dark mode test for the variants?

the menu item text won't look right because it doesn't have the `Colour["text"]` token applied. will need to fix that

qroll (approved):

PR #1116, opened 2026-04-22
Narrow modal child typings

qroll:
**Type of changes**

-   [x] Breaking change (fix or feature that would cause existing apis or functionality to change)

**Description of changes**

- Modal expects a valid parent element (component or inlined JSX) so that it can forward the ref to the element to detect the container boundaries
- However the typing for ReactNode allows strings and multiple children, which don't work with `cloneElement` and causes an error during rendering
- One way to limit this issue is to make the typings stricter. Though it is not possible to detect Fragments afaik; a fragment does not cause an error, but overlay click detection will no longer work

**Checklist**

-   [x] Updated documentation
-   [ ] ~~Added/updated tests~~


thyAshley (commented):
lgtm

thyAshley (approved):

PR #1119, opened 2026-04-23
[BOOKINGSG-9210] refactor: migrate shared/internal-calendar to linaria and v4 tokens

ziggyzet:
**Checklist**

~-   [ ] Added visual tests~

**Additional information**
- No addition to visual tests. Test was based on components using `internal-calendar`


ghazwan-gt:
Can help with rerunning the e2e tests for DateInput and DateRangeInput

src/shared/internal-calendar/calendar-dropdown.styles.ts
```diff
@@ -0,0 +1,31 @@
...
+export const tokens = {
+    calendarWrapper: {
+        width: "--fds-internal-calendarDropdown-calendarWrapper-width",
+    },
+};
+
+export const calendarWrapper = css`
+    --vertical-inset: ${Spacing["spacing-24"]};
+    --horizontal-inset: ${Spacing["spacing-20"]};
+    --header-bottom-spacing: ${Spacing["spacing-4"]};
```
ghazwan-gt: Let's have these to be inside tokens and then update the one in `Calendar` and `CalendarManager`

src/shared/internal-calendar/calendar-dropdown.styles.ts
```diff
@@ -0,0 +1,31 @@
...
+import { Border, Colour, MediaQuery, Radius, Spacing } from "../../theme";
+
+// =============================================================================
+// STYLING
+// =============================================================================
+export const tokens = {
+    calendarWrapper: {
+        width: "--fds-internal-calendarDropdown-calendarWrapper-width",
+    },
+};
```
ghazwan-gt, line 13: Since we only have one styling here, let's have the tokens as immediate property

```suggestion
export const tokens = {
    width: "--fds-internal-calendarDropdown-calendarWrapper-width",
  
};
```

src/shared/internal-calendar/calendar-manager.styles.ts
```diff
@@ -104,45 +78,33 @@ export const Header = styled.div`
...
-
-        if (props.$expanded) {
-            return css`
-                ${IconChevronDown} {
-                    transform: rotate(180deg);
-                }
-            `;
-        }
-    }};
+export const dropdownButtonHidden = css`
```
ghazwan-gt: Can be combined with `optionsOverlayHidden` to be a general `hidden` class name

src/shared/internal-calendar/calendar-manager.tsx
```diff
@@ -508,26 +520,28 @@ const Component = (
         const disabled = !isDayView ? false : doneButtonDisabled;
 
         return (
-            <ActionButtonSection>
-                <ActionButton
+            <div className={styles.actionButtonSection}>
+                <Button.Small
```
ghazwan-gt: Let's utilize `sizeType`

src/shared/internal-calendar/internal-calendar-month.styles.ts
```diff
@@ -0,0 +1,72 @@
...
+export const monthCell = css`
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    border-radius: ${Radius["md"]};
+    margin: 0 0.5rem;
+    transition: ${Motion["duration-150"]} ${Motion["ease-default"]};
+
+    ${Font["body-md-regular"]}
+    border-radius: ${Radius["md"]};
```
ghazwan-gt: Can remove this one, already exist on line 24

src/shared/internal-calendar/internal-calendar-month.tsx
```diff
@@ -187,25 +188,35 @@ export const InternalCalendarMonth = ({
...
+                            disabledDisplay && styles.monthCellDisabledDisplay,
+                            interactive && styles.monthCellInteractive,
+                            (variant === "current-month" &&
+                                styles.monthCellCurrentMonth) ||
+                                (variant === "selected-month" &&
+                                    styles.monthCellSelectedMonth) ||
+                                (interactive &&
+                                    variant !== "selected-month" &&
+                                    styles.monthCellDefaultHover) ||
+                                (interactive && styles.monthCellSelectedHover)
```
ghazwan-gt: I don't think this is the right place to have OR operator, seems like this will have an unexpected result

src/shared/internal-calendar/internal-calendar-year.styles.ts
ghazwan-gt, line 1: I see a lot similarities between this one and the month one, should we have a general styling accommodating both instead?

ziggyzet: i was thinking this but I wasn't sure that we'd have issues where perhaps one of the components have specific styling and it'll get too big for each one.

let me know if combining both would be more suitable

qroll: I'm ok with a little bit of duplication here since it's just a few styles

if we are looking at general styling, what I can think of is a standalone button based on the documented DS parts:

Image

but don't think there's much use for it outside of the calendar so it might not be worth the refactoring effort

src/shared/internal-calendar/internal-calendar-month.tsx
ziggyzet: fixing this one!

src/shared/internal-calendar/calendar-dropdown.tsx
```diff
@@ -1,26 +1,44 @@
...
+        styles: positioningStyles,
+    } = useDropdownRender();
+
+    const wrapperRef = useRef<HTMLDivElement>(null);
+
+    useApplyStyle(wrapperRef, {
+        [styles.tokens.calendarWrapper.width]: `${elementWidth}px`,
+    });
+
+    const setRef = (el: HTMLDivElement | null) => {
```
qroll: are we able to use the `mergeRefs` util here?

src/shared/internal-calendar/internal-calendar-month.tsx
qroll: the current month/year are missing `xCellDefaultHover` since we're short circuiting on the `current-month` and `current-year`

Image

qroll (changes requested):

src/shared/internal-calendar/internal-calendar-year.styles.ts
ghazwan-gt: Fair points, lets keep it then

ghazwan-gt:
Just noticed that the screenshot for `DateRangeInput-Range-With-maxDate-constraint-Open--state.png` is wrong. Please help with generate a new one for it

src/shared/internal-calendar/internal-calendar-year.tsx
```diff
@@ -210,18 +211,30 @@ export const InternalCalendarYear = ({
...
+                                styles.yearCellOtherDecade,
+                            variant === "current-year" &&
+                                styles.yearCellCurrentYear,
+                            variant === "selected-year" &&
+                                styles.yearCellSelectedYear,
+                            interactive &&
+                                variant !== "selected-year" &&
+                                styles.yearCellDefaultHover,
+                            interactive && styles.yearCellSelectedHover
+                        )}
```
qroll, line 230: the styling here is still not quite right. `yearCellInteractive` and `yearCellSelectedHover` have the same conditions so `yearCellSelectedHover` is applying for non-selected years as well. it should only be limited to the selected state

src/shared/internal-calendar/internal-calendar-month.tsx
ghazwan-gt: Hovering non-selected month should not have border

image

src/shared/internal-calendar/calendar-manager.styles.ts
```diff
@@ -116,6 +112,10 @@ export const headerArrowButton = css`
     }
 `;
 
+export const dropdownHidden = css`
```
qroll: hmm naming-wise I think this should still be `hidden`? dropdownHidden makes it seem like it's for a dropdown component

and since this is a common style, can put this under a `COMMON` section to separate it from the other components

ziggyzet: oops, i thought it was a class naming issue before.. i'll revert it back!

qroll (approved):

qroll:
might be another css load order issue, but the header spacing doesn't match production (left). can you check on this?

Screenshot 2026-05-05 at 6 39 52 PM

ghazwan-gt (approved):

PR #1121, opened 2026-04-23
[BOOKINGSG-9233][IT] Migrate box-component to linaria

iant-gds:
**Checklist**

-   [x] Added visual tests


src/box-container/box-container.tsx
```diff
@@ -42,13 +30,20 @@ export const BoxContainer = ({
     );
     const resizeDetector = useResizeDetector();
     const childRef = resizeDetector.ref;
-    const theme = useContext(ThemeContext);
-    const mobileBreakpoint = V3_Breakpoint["sm-max"]({ theme });
+    const mobileBreakpoint = Breakpoint["sm-max"];
     const isMobile = useMediaQuery({ maxWidth: mobileBreakpoint });
```
ghazwan-gt:
```suggestion
    const mobileBreakpoint = useDesignToken(Breakpoint["sm-max"]);
    const isMobile = useSafeMaxWidthMediaQuery(mobileBreakpoint);
```

src/box-container/box-container.styles.ts
```diff
@@ -0,0 +1,143 @@
...
+    height: ${Font.Spec["heading-size-sm"]};
+    width: ${Font.Spec["heading-size-sm"]};
+`;
+
+// Animated component wrapper
+export const AnimatedExpandable = animated.div;
+
+// Icon components
+export const AlertIconComponent = ExclamationCircleFillIcon;
+export const HandleIconComponent = ChevronDownIcon;
```
ghazwan-gt: These can be referenced directly in the component

src/box-container/box-container.styles.ts
```diff
@@ -0,0 +1,143 @@
...
+        margin-top: ${Spacing["spacing-16"]};
+    }
+`;
+
+export const handle = css`
+    background: transparent;
+    border: none;
+    cursor: pointer;
+    padding: ${Spacing["spacing-16"]};
+    margin: -${Spacing["spacing-16"]};
```
ghazwan-gt:
```suggestion
    margin: ${Spacing["spacing-16"]};
```

e2e/tests/components/box-container/box-container.e2e.spec.ts
```diff
@@ -0,0 +1,440 @@
...
+                    0
+                );
+            });
+
+            // Verify container is visible and accessible
+            await test.step("Container should be visible and accessible", async () => {
+                await expect(
+                    story.locators.container.nonCollapsible
+                ).toBeVisible();
+            });
```
qroll: visibility assertions are not required since we have a visual test that captures the current visible elements already

e2e/tests/components/box-container/box-container.e2e.spec.ts
```diff
@@ -0,0 +1,440 @@
...
+test.describe("BoxContainer", () => {
+    test.describe(() => {
+        test.beforeEach(async ({ story }) => {
+            await story.init("non-collapsible");
+        });
+
+        test("Non-collapsible container", async ({ story }) => {
+            // Verify no handle buttons are present for non-collapsible containers
+            await test.step("Handle buttons should not be present", async () => {
+                await story.page.waitForLoadState("networkidle");
```
qroll: no need to specially wait for load state. the init function already waits for the page to be loaded and we are not making any api calls here

e2e/tests/components/box-container/box-container.e2e.spec.ts
```diff
@@ -0,0 +1,440 @@
...
+
+            // Verify container is visible and accessible
+            await test.step("Container should be visible and accessible", async () => {
+                await expect(
+                    story.locators.container.nonCollapsible
+                ).toBeVisible();
+            });
+
+            // Visual regression test
+            await test.step("Visual appearance", async () => {
```
qroll: the step can be removed

e2e/tests/components/box-container/box-container.e2e.spec.ts
qroll, line 1: general comments:
- prefer `toMatchAriaSnapshot()` to assert the aria semantics instead of checking element attributes via `toHaveAttribute()`
- no need to assert visibility if you are already using `compareScreenshot` for a visual test unless you have a good reason to do so
- remove redundant comments from the file

src/box-container/box-container.styles.ts
```diff
@@ -0,0 +1,143 @@
...
+                "spacing-16"
+            ]} 0;
+        margin-left: auto;
+    }
+`;
+
+export const handleIconContainer = css`
+    /* reset variable to prevent leaking to child components */
+    ${tokens.handleIconContainer.rotation}: initial;
+    transform: rotate(var(${tokens.handleIconContainer.rotation}, 0deg));
```
qroll: we don't need a variable here since the rotation value is static. it's either `0` or `180`. use another class to apply the transform property again

qroll (changes requested):

e2e/tests/components/box-container/__screenshots__/chromium/BoxContainer-Container-with-call-to-action-component--call-to-action-container.png
qroll, line 1: the screenshot currently has a hover state on the button. we should capture this screenshot on mount before any interactions in the e2e test

src/box-container/box-container.styles.ts
qroll: the original style did have negative margin. was meant to offset the padding added to increase the touch target for the chevron

iant-gds: Will not be updating this then

ghazwan-gt (approved):

src/box-container/box-container.styles.ts
qroll: the style is still not actually applied due to syntax error. will need to resolve this

Screenshot 2026-04-27 at 6 13 51 PM

e2e/tests/components/box-container/box-container.e2e.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+
+        test("Display states (error and warning icons)", async ({ story }) => {
+            await compareScreenshot(story, "display-states");
+
+            await expect(story.locators.container.error).toMatchAriaSnapshot(`
+                    - region "Container with error state error error":
+                      - status:
+                        - text: Container with error state error
+                        - img "error"
+                      - button "Container with error state error error"
```
qroll: for all handle buttons, assert that they are expanded/collapsed in the snapshot using `expand` and `expanded=false`

e2e/tests/components/box-container/box-container.e2e.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+        test("Expand/collapse interaction", async ({ story }) => {
+            await test.step("Visual appearance of expand/collapse states", async () => {
+                const handles = story.page.getByTestId("handle");
+                await compareScreenshot(story, "expand-collapse-mixed-states");
+
+                // Ensure first is expanded, second is collapsed
+                await handles.nth(0).click();
+                await handles.nth(1).click();
+
+                // First container starts collapsed
```
qroll: we should assert initial mount behaviour first. the click actions can be done in a separate step

e2e/tests/components/box-container/box-container.e2e.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+                      - button "Title"
+                `);
+
+                // Second container starts expanded
+                await expect(story.locators.container.expanded)
+                    .toMatchAriaSnapshot(`
+                    - region "Title":
+                      - status: Title
+                      - button "Title"
+                      - paragraph: Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
```
qroll: nit: the text here is very long, if it's not actually needed, we can simplify the content to just `Lorem ipsum dolor sit amet`

qroll (changes requested):

qroll (approved):

PR #1125, opened 2026-04-24
[BOOKINGSG-9266][GZ] update media query specificity

ghazwan-gt:
**Type of changes**

**Description of changes**
The previous media query has a `0, 2, 0` or `0, 2, 1` specificity. This is caused by the two present classnames alongside the `body` element.
image
Having this kind of specificity is out of the norm for a component library, since in most cases, consumer will have a `0, 1, 0` specificity by adding a classname when using a css module or linaria. In the case consumer use styled component, it usually wins since styled components styling are injected during runtime at the end of the `<head>` by default.

The end result we have is to use `:where` selector which will always have [0 specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:where). This is an example of having a consumer override a property through class name
image

**Checklist**

-   [ ] Updated documentation
-   [ ] Added/updated tests


e2e/nextjs-app/src/app/components/theme/media-query-specificity.e2e.tsx
qroll, line 1: could we include snapshots for this test? (can exclude the text and target the locators directly to reduce flakiness)

e2e/nextjs-app/src/app/components/theme/media-query-specificity.e2e.tsx
```diff
@@ -0,0 +1,94 @@
...
+        <div className={styles.block}>
+            <h3>{title}</h3>
+            {children}
+        </div>
+    );
+}
+
+export default function Story() {
+    useEffect(() => {
+        document.body.classList.add("fds-breakpoint-sm-min");
```
qroll: wondering why we need to force the media query? would it be sufficient to have the snapshots captured at different viewports

e2e/nextjs-app/src/app/components/theme/media-query-specificity.e2e.tsx
ghazwan-gt: Sure, will add it

e2e/nextjs-app/src/app/components/theme/media-query-specificity.e2e.tsx
ghazwan-gt: Yeah we could also do that. I was thinking that having the test on one viewport is sufficient as we want to test the consumer styling overrides ours

qroll: yeah on second thought 1 viewport is sufficient since the layout component doesn't do anything special with specificity

we could use `story.init({ viewport: "sm" })` directly. then there wouldn't be a need to manually override the breakpoint class (which is an implementation detail and we could miss this file out in the future)

ghazwan-gt: I'm not pushing any changes for this. We don't need to have this with different viewports as checking it once is sufficient

qroll: what I mean is to remove the custom `useEffect` and take a screenshot in `sm` viewport mode. no need for other viewport sizes

e2e/tests/components/theme/__screenshots__/chromium/Theme-Enable-consumer-to-override-media-query-specificity--mount.png
qroll, line 1: can be removed

e2e/nextjs-app/src/app/components/theme/media-query-specificity.e2e.tsx
ghazwan-gt: Ooh okay

ghazwan-gt: I forgot to refresh the page, didn't get your recent comment 😅

qroll (approved):

PR #1136, opened 2026-04-29
[BOOKINGSG-9278] Migrate range-slider

qroll:
**Checklist**

-   [x] Updated Storybook documentation
-   [x] Added visual tests
-   [x] Listed breaking changes

**Breaking changes**

- `color` props for sliders simplified to only accept string. The string accepts valid CSS values (e.g. colour hex) or variables for theming

**Additional information**

- we need to migrate away from `react-slider` eventually for various reasons
  - not maintained, React 19 is not supported
  - it uses inline styles which causes violations on strict CSP
- for now to make the component work, the styles from the library need to be applied with `useApplyStyle` which I enhanced to support the camelCase'd css properties

side note this is how it's implemented in [React](https://github.com/facebook/react/blob/ad5dfc82b7107728da1430dd142f75b97b684dac/packages/react-dom-bindings/src/client/CSSPropertyOperations.js#L76)


src/theme/utils/use-apply-styles.ts
```diff
@@ -1,33 +1,81 @@
...
+    "marginRight",
+    "marginBottom",
+    "padding",
+    "paddingTop",
+    "paddingLeft",
+    "paddingRight",
+    "paddingBottom",
+];
+
+function normaliseCssValue(key: string, value: unknown): string {
```
ghazwan-gt: Lets add this case and undefined styles to the `use-apply-styles.spec`

ghazwan-gt (approved):

ghazwan-gt (approved):

src/input-range-slider/slider-components.tsx
```diff
@@ -0,0 +1,75 @@
...
+    { color, className, style, ...otherProps }: TrackProps,
+    ref: React.Ref<HTMLDivElement>
+) => {
+    const trackRef = useRef<HTMLDivElement>(null);
+
+    useApplyStyle(trackRef, {
+        [styles.tokens.track.backgroundColor]: color,
+    });
+
+    useApplyStyle(trackRef, style);
```
zhaoyanxzy, line 64: will having two `useApplyStyle` cause one style (either `color` or `style`) to be overwritten? Or should be merge them like
```
useApplyStyle(trackRef, {
        ... style,
        [styles.tokens.track.backgroundColor]: color,
    });
```

src/theme/utils/use-apply-styles.ts
```diff
@@ -1,33 +1,101 @@
...
+
+    if (/^(o|moz|ms|webkit)-/.test(kebabCased)) {
+        return `-${kebabCased}`;
+    }
+
+    return kebabCased;
+}
+
+function normaliseCssValue(key: string, value: unknown): string {
+    if (value == null) return "";
```
zhaoyanxzy: [nit]: already checks for `==null` before invoking

src/input-range-slider/slider-components.tsx
qroll: was thinking this is still safe, the token has a unique name so the style prop won't overlap

and also reduces unnecessary processing if the style prop did not change since we're not creating a new object reference by spreading it

src/theme/utils/use-apply-styles.ts
qroll: hmm let me try to tighten the typings

the function signature should be

```
function normaliseCssValue(key: string, value: string | number): string
```

and the style entries can be typed like this

```
Object.entries(styles).forEach(([key, value]: [string, StyleValue]) => {
```

PR #1145, opened 2026-04-30
[BOOKINGSG-9160][GZ] migrate input multi select to linaria

ghazwan-gt:
**Checklist**

-   [ ] Migrated the component styles
    -   [ ] `className` is chained correctly with `clsx`
-   [ ] Changes follow the project guidelines in [CONVENTIONS_V4.md](https://github.com/LifeSG/react-design-system/blob/master/CONVENTIONS_V4.md)
-   [X] Added visual tests

**Additional information**
- No direct changes on the component.
- Provided a more complete e2e cases.


src/shared/dropdown-list/dropdown-label.styles.ts
```diff
@@ -10,7 +10,7 @@ export const tokens = {
     secondaryText: {
         maxLines: "--fds-internal-dropdownLabel-secondaryText-maxLines",
     },
-};
+} as const;
```
ziggyzet: is there any particular reason for this addition?

e2e/tests/components/input-multi-select/multi-select.e2e.spec.ts
```diff
@@ -0,0 +1,312 @@
...
+        });
+    });
+
+    test.describe("Virtualization", () => {
+        test.describe(() => {
+            test.beforeEach(async ({ story }) => {
+                await story.init("virtualization");
+            });
+
+            test("", async ({ story }) => {
```
ziggyzet: maybe we can add a title to what we're trying to do in this test

e2e/tests/components/input-multi-select/multi-select.e2e.spec.ts
```diff
@@ -0,0 +1,312 @@
...
+        });
+    });
+
+    test.describe("Custom CTA", () => {
+        test.describe(() => {
+            test.beforeEach(async ({ story }) => {
+                await story.init("custom-cta");
+            });
+
+            test("", async ({ story }) => {
```
ziggyzet: can add a title here too

src/shared/dropdown-list/dropdown-label.styles.ts
ghazwan-gt: This one is to enable TS to know the exact string value when referencing the tokens inside `lineClampDynamicCss`. TS complained about it during compilation

e2e/tests/components/input-multi-select/input-multi-select.e2e.spec.ts
```diff
@@ -70,131 +112,275 @@ test.describe("InputMultiSelect", () => {
...
+                  - option "Option D"
+            `);
+            await compareScreenshot(story, "open", {
+                fullscreen: true,
             });
         });
     });
 
-    test.describe("Keyboard interaction", () => {
+    test.describe("", () => {
```
ziggyzet: some of the test suites are still empty. maybe for better readibility and consistency, we can just omit adding a title here? would the worry be that the structure would be different if we omit it?

ghazwan-gt: Ah my bad, I just noticed that there are duplicated describe blocks. All those blocks are removed!

src/shared/dropdown-list/dropdown-label.styles.ts
qroll: what was the TS error? is it specific to `lineClampCss`? there are a few places where `const` is not specified and those seem alright

e2e/tests/components/input-multi-select/input-multi-select.e2e.spec.ts
```diff
@@ -0,0 +1,485 @@
...
+        });
+    });
+
+    test.describe(() => {
+        test.beforeEach(async ({ story }) => {
+            await story.init("with-search");
+        });
+
+        test("With search", async ({ story }) => {
+            await test.step("Default", async () => {
```
qroll: instead of steps, these could be individual test cases, especially since they are independent of each other. `With search` can be a describe block instead

e2e/tests/components/input-multi-select/input-multi-select.e2e.spec.ts
```diff
@@ -0,0 +1,485 @@
...
+
+    test.describe(() => {
+        test.beforeEach(async ({ story }) => {
+            await story.init("virtualization");
+        });
+
+        test("Virtualization", async ({ story }) => {
+            await story.openDropdown(story.locators.multiSelect);
+            await story.locators.component.searchInput.fill("Option 4999");
+            await expect(story.getOption("Option 4999")).toBeVisible();
```
qroll: for virtualisation, the main thing to test is the rendering of elements as they are scrolled into view. we can simulate scrolling on the dropdown list and verify that options are rendering correctly

e2e/nextjs-app/src/app/components/input-multi-select/small-variants.e2e.tsx
qroll, line 1: rename to `small-variant`

qroll (changes requested):

src/shared/dropdown-list/dropdown-label.styles.ts
ghazwan-gt: Yes, it is related to `lineClampDynamicCss` @qroll . And all our usage have const assertion

qroll (approved):

PR #1146, opened 2026-05-04
[MISC][GZ] add eslint inline style rule for nextjs app

ghazwan-gt:
**Type of changes**

**Description of changes**
Introduce new rule for preventing adding inline styles inside the Nextjs app for E2E stories. 

I considered using [react/forbid-component-props](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/forbid-component-props.md) but that rule is better suited to restricting props on specific React components. In this case, we want to disallow inline styles across both custom components and native DOM elements, so it is not the right fit.

Instead, this uses [no-restricted-syntax](https://eslint.org/docs/latest/rules/no-restricted-syntax) in the Next.js app ESLint config to enforce a blanket restriction on JSX style attributes.


package.json
```diff
@@ -34,8 +34,8 @@
         "build-storybook": "storybook build",
         "codemod": "ENV=dev npx tsx codemods/run-codemod.ts",
         "lint": "npm run lint:js && npm run lint:css",
-        "lint:js": "eslint '**/*.{ts,tsx,js,jsx}'",
-        "lint:js:fix": "eslint '**/*.{ts,tsx,js,jsx}' --fix",
+        "lint:js": "eslint '**/*.{ts,tsx,js,jsx}' && npm --prefix e2e/nextjs-app run lint:src",
+        "lint:js:fix": "eslint '**/*.{ts,tsx,js,jsx}' --fix && npm --prefix e2e/nextjs-app run lint:src:fix",
```
qroll, line 38: have you also considered configuring the root eslint config to cover the e2e folder?

```
    {
        files: ["e2e/nextjs-app/**/*.{js,jsx,ts,tsx}"]
    },
```

instead of running subcommands

e2e/nextjs-app/src/app/components/breadcrumb/fade-gradients.e2e.tsx
```diff
@@ -30,7 +31,7 @@ export default function Story() {
     const fadeColorRight = ["blue", "green"];
 
     return (
-        <div className="story-column-container" style={{ gap: "2rem" }}>
+        <div className={`story-column-container ${styles["container-gap"]}`}>
```
qroll: we can remove this style since it's on the parent element and the story layout already comes with built in padding

e2e/nextjs-app/src/app/components/[component]/[story]/page.tsx
```diff
@@ -13,10 +13,10 @@ export default async function Page({
         {
             ssr: ![
                 "badge",
+                "box-container",
```
Comment (qroll) [inline e2e/nextjs-app/src/app/components/[component]/[story]/page.tsx]: why does box-container need to be added here?

package.json
ghazwan-gt: Yeah we can do that. In that case, we need to move out the Nextjs config to the root since v9 doesn't merge scattered config anymore

e2e/nextjs-app/src/app/components/breadcrumb/fade-gradients.e2e.tsx
ghazwan-gt: Sure

e2e/nextjs-app/src/app/components/[component]/[story]/page.tsx
Comment (ghazwan-gt) [reply e2e/nextjs-app/src/app/components/[component]/[story]/page.tsx]: Didn't check further on this, but there was a hydration error when opening the story page

package.json
ghazwan-gt: After exploring, we can't bring all the Nextjs ESLint config to the root config without bringing along the Nextjs dependencies. Let's keep it separate for now

ghazwan-gt: Or we can just add this particular rule in the root config

qroll: in that case, it's alright to keep the configs separate. more consistent

e2e/nextjs-app/src/app/components/[component]/[story]/page.tsx
Comment (qroll) [reply e2e/nextjs-app/src/app/components/[component]/[story]/page.tsx]: that's odd since box container shouldn't have any styled component dependencies any longer. seems to be fine locally as well

 can check on this again

qroll (approved):

e2e/nextjs-app/src/app/components/[component]/[story]/page.tsx
Comment (ghazwan-gt) [reply e2e/nextjs-app/src/app/components/[component]/[story]/page.tsx]: Just tried it and it is fine now. Resolving this

PR #1151, opened 2026-05-05
[BOOKINGSG-9220] Migrate pagination

qroll:
**Checklist**

    -   [ ] ~~User style prop is set as CSS variable~~
-   [ ] ~~Updated Storybook documentation~~
-   [x] Added visual tests
-   [ ] ~~Listed breaking changes~~

**Additional information**

- Manually disabling NextJS dev indicator in screenshots
- The id mismatch issue should be fixed by [this PR](https://redirect.github.com/LifeSG/react-design-system/pull/1144), pending merge
- The html mismatch issue is fixed by added a `isMounted` check, based on [this workaround](https://react.dev/reference/react-dom/client/hydrateRoot#handling-different-client-and-server-content) in the React docs
- My thought is to eventually mark a test as failed (using soft assertions) if a hydration error is detected - but only enabling this after any existing hydration issues are resolved, we likely only have bandwidth for this after v4 migration


e2e/tests/components/pagination/pagination.e2e.spec.ts
```diff
@@ -0,0 +1,524 @@
...
+
+                    await compareScreenshot(story, "dropdown-open", {
+                        fullscreen: true,
+                    });
+                });
+            });
+        });
+    });
+
+    test.describe("showFirstAndLastNav=true", () => {
```
zhaoyanxzy, line 366: is it missing a dark mode in mobile test?

src/pagination/pagination.styles.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    svg {
+        height: 1.25rem;
+        width: 1.25rem;
+    }
+
+    &:disabled {
+        color: ${Colour["icon-disabled-subtle"]};
+        cursor: not-allowed;
+    }
```
zhaoyanxzy: Qn: There's no disabled prop passed to the `renderStartEllipsis` but there is this style, is the prop missing or this styling is not required?

src/pagination/pagination.styles.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    &:not(:disabled):hover,
+    &:not(:disabled):focus-visible {
+        background: ${Colour["bg-hover"]};
```
zhaoyanxzy, line 52: nit: ellipsisButton uses `background-color`

src/pagination/pagination.styles.ts
```diff
@@ -0,0 +1,195 @@
...
+        align-self: center;
+    }
+`;
+
+export const paginationMenu = css`
+    display: inline-flex;
+    align-items: center;
+`;
+
+export const ellipsisButton = css`
```
zhaoyanxzy: since there are quite a few repeated codes in ellipsisButton and navigationButton, is it better to have a base style e.g. `iconButton` then use `clsx` library to add the different stylings?

e2e/tests/components/pagination/pagination.e2e.spec.ts
qroll: thinking we can skip this combination since the buttons are expected to behave the same

so we have sufficient coverage via dark mode (desktop) and mobile

src/pagination/pagination.styles.ts
qroll: hmm you're right, was probably copy-pasted from the other styles back then. will clean this up

src/pagination/pagination.styles.ts
qroll: standardised to use `background`, matching the other styles in this file

zhaoyanxzy (approved):

PR #1157, opened 2026-05-06
[BOOKINGSG-9317][GZ] include linaria css in stylelint

ghazwan-gt:
## Summary
- Migrated Stylelint config from JSON to JS config and added Linaria PostCSS custom syntax with Linaria-related rule adjustments.
- Expanded Stylelint execution scope to include `*.styles.ts` files across NPM scripts and lint-staged.
- Updated VS Code Stylelint settings so IDE linting uses the new config.
- Introduced new local ESLint rule to prevent negative interpolation usage.


eslint-local-rules/import-path-preferences.mjs
```diff
@@ -314,9 +314,75 @@ const styleNamespaceImportRule = {
     },
 };
 
+const noNegativeLinariaInterpolationRule = {
```
qroll: extract to a separate file

stylelint.config.mjs
qroll, line 1: since we are using mjs for eslint, could we also use the same extension for stylelint?

stylelint.config.mjs
```diff
@@ -0,0 +1,52 @@
...
+                    "always",
+                    {
+                        except: ["first-nested"],
+                        ignore: ["stylelint-commands"],
+                        ignoreComments: [/pcss-lin/],
+                    },
+                ],
+                "no-invalid-double-slash-comments": true,
+
+                // Ignore Linaria placeholder values like pcss-lin0 and -pcss-lin1.
```
qroll, line 73: not sure how this works. where are these placeholder values introduced?

stylelint.config.mjs
```diff
@@ -0,0 +1,52 @@
...
+        "color-hex-length": "long",
+
+        "@stylistic/color-hex-case": "upper",
+        "@stylistic/indentation": 4,
+        "@stylistic/max-empty-lines": 1,
+    },
+    overrides: [
+        {
+            /*
+            Taken from linaria's stylelint-config-standard-linaria package,
```
qroll, line 40: what's different from the npm linter plugin? https://github.com/callstack/linaria/blob/master/docs/LINTING.md#stylelint-14

eslint-local-rules/import-path-preferences.mjs
ghazwan-gt: Sure, will do that

stylelint.config.mjs
ghazwan-gt: Ah sure

stylelint.config.mjs
ghazwan-gt: It is part of `@linaria/postcss-linaria` where they turn expression into those placeholders during parsing. Will take a look at the flow of having this custom sytax with stylelint

stylelint.config.mjs
ghazwan-gt: It should be the same, but using the one from npm keeps throwing error even when the resulted config is the same

stylelint.config.mjs
ghazwan-gt: Okay after taking a look of how stylelint works and the linaria postcss custom syntax, basically this is how it works
1. File matching phase where we're looking for which files match the one where we want the syntax to be applied
2. Parse phase where stylelint loads the linaria's syntax and parse it into a postcss AST. And during this phase, linaria replaced all string interpolation into those placeholders
3. Validation phase where stylelint checks all the rules applied
4. Fix/Stringify phase where all the fix takes place (with `--fix`) and serialize the output back using the linaria's syntax stringify behavior, which replacing those placeholders back

src/accordion/accordion-item.styles.ts
```diff
@@ -65,16 +56,16 @@ export const titleSmall = css`
 
 export const iconContainer = css`
     transform: rotate(180deg);
-    transition: transform ${Motion["duration-250"]}${Motion["ease-default"]};
+    transition: transform ${Motion["duration-250"]};
```
ziggyzet: uhh why was this part deleted here? is this intentional?

src/alert/alert.styles.ts
```diff
@@ -146,9 +145,9 @@ export const textWrapperContainerShowMoreWithLink = css`
 
 export const textWrapperContainerCollapsed = css`
     ${tokens.maxCollapsedHeight}: initial;
+
     max-height: var(${tokens.maxCollapsedHeight});
     overflow: hidden;
-    -webkit-mask-image: linear-gradient(to bottom, black 50%, transparent 100%);
```
ziggyzet, line 151: this one is removed as well

src/badge/badge.styles.ts
```diff
@@ -31,23 +31,22 @@ export const badgeOverlay = css`
...
+
     position: absolute;
     top: 0;
     right: 0;
-    transform: translate(50%, -25%)
-        translate(
-            var(${tokens.wrapper.offsetX}, 0px),
-            var(${tokens.wrapper.offsetY}, 0px)
-        );
+    transform: translate(50%, -25%) translate(var(${tokens.wrapper.offsetX}), var(${tokens.wrapper.offsetY}));
```
ziggyzet, line 40: is this change to suggest us to reinitialize the variables instead of using fallbacks?

src/box-container/box-container.styles.ts
```diff
@@ -58,24 +50,23 @@ export const headerInteractive = css`
 
 export const labelText = css`
     ${Font["heading-sm-semibold"]}
-    color: ${Colour["text"]};
-    word-wrap: break-word;
```
ziggyzet, line 61: word-wrap is gone here

ziggyzet (commented):
some attributes went missing, will check the rest as well

package.json
```diff
@@ -36,8 +36,8 @@
         "lint": "npm run lint:js && npm run lint:css",
         "lint:js": "eslint '**/*.{ts,tsx,js,jsx}' && npm --prefix e2e/nextjs-app run lint:src",
         "lint:js:fix": "eslint '**/*.{ts,tsx,js,jsx}' --fix && npm --prefix e2e/nextjs-app run lint:src:fix",
-        "lint:css": "stylelint '**/*.css'",
-        "lint:css:fix": "stylelint '**/*.css' --fix",
+        "lint:css": "stylelint --config stylelint.config.mjs '**/*.css' '**/*.styles.ts'",
+        "lint:css:fix": "stylelint --config stylelint.config.mjs '**/*.css' '**/*.styles.ts' --fix",
```
qroll: do we need to specify `--config` explicitly and in the vscode settings? I assumed stylelint would pick up this filename by default

.prettierrc
```diff
@@ -1,16 +1,22 @@
...
+    "semi": true,
+    "singleQuote": false,
+    "bracketSpacing": true,
+    "arrowParens": "always",
+    "overrides": [
+        {
+            "files": "**/*.styles.ts",
+            "options": {
+                "printWidth": 120
+            }
```
qroll, line 13: what is this for?

src/accordion/accordion-item.styles.ts
ghazwan-gt: This one already resolved through [142cdd8](https://github.com/LifeSG/react-design-system/pull/1157/commits/142cdd84cd763f4e9611621eaf1bd044df145e7b)

src/alert/alert.styles.ts
ghazwan-gt: Also resolved through [142cdd8](https://github.com/LifeSG/react-design-system/pull/1157/commits/142cdd84cd763f4e9611621eaf1bd044df145e7b)

package.json
ghazwan-gt: With cjs it needs the flag, but let me try with mjs

.prettierrc
ghazwan-gt: This is to help us preventing the linaria's postcss which wrongly process interpolation when it is on multiple lines

src/box-container/box-container.styles.ts
ghazwan-gt: We have `overflow-wrap: anywhere` just below it which means the `word-wrap` is not being applied at all

PR #1160, opened 2026-05-08
[BOOKINGSG-9324][GZ] eslint negative interpolation rule

ghazwan-gt:
**Type of changes**

**Description of changes**
Using tokens for providing negative value with interpolation resulted in invalid value
```
  // invalid
  padding: -${tokens.padding};
```
To help with this, we are introducing a new local ESLint rule to recognize this and also fix it automatically
```
  // valid
  padding: calc(${tokens.paddding} * -1)
```


ziggyzet (approved):

tests/eslint-rules/import-path-preferences.rule-test.mjs
```diff
@@ -153,13 +153,13 @@ ruleTester.run("style-namespace-import", styleNamespaceImportRule, {
         {
             name: "default import from .styles path is rejected",
             code: 'import styles from "./button.styles";',
-            output: 'import * as styles from "./button.styles";',
+            output: null,
```
qroll: ah is this because the autofix was disabled? let's keep the original output commented out, it can then be restored after v4 migration

tests/eslint-rules/import-path-preferences.rule-test.mjs
qroll: we can also separate this into its own test file to match the new structure

tests/eslint-rules/no-negative-linaria-interpolation.rule-test.mjs
```diff
@@ -0,0 +1,114 @@
...
+`,
+            },
+        ],
+        invalid: [
+            {
+                name: "negative interpolation is rewritten to calc",
+                filename: "/repo/src/button/button.styles.ts",
+                code: `import { css } from "@linaria/core";
+
+const spacing = "8px";
```
qroll, line 85: actually, hardcoded values are alright? the problematic part is only for `var()`. I don't suppose there's a definite way to detect that we are using the DS tokens or variable tokens?

ghazwan-gt: Yeah I don't think we have a way to do that since this is a static tool. And for both cases, it will be invalid since it will be processed by linaria's css

qroll: got it. should be fine since we should be sticking with the Spacing tokens anyway

eslint-local-rules/import-path-preferences.mjs
qroll, line 1: I think this file is meant to be deleted after renaming? along with `eslint-local-rules/style-namespace-import.mjs`

tests/eslint-rules/style-namespace-import.rule-test.mjs
zhaoyanxzy, line 1: hmm why is this empty?

qroll (approved):

PR #1163, opened 2026-05-11
[BOOKINGSG-9156][IT] Migrate input-group component

iant-gds:
**Checklist**

~-   [ ] Added visual tests~


src/input-group/input-group.tsx
```diff
@@ -87,30 +88,45 @@ const Component = <T, V>(
...
+                                    otherProps.disabled &&
+                                        styles.labelAddonContainerDisabled,
+                                    otherProps.readOnly &&
+                                        (position === "right"
+                                            ? styles.labelAddonContainerPositionRightReadOnly
+                                            : styles.labelAddonContainerReadOnly),
+                                    !otherProps.readOnly &&
+                                        position === "right" &&
+                                        styles.labelAddonContainerPositionRight,
+                                    className
```
ghazwan-gt: I don't think we need to apply the parent className here since it could unintentionally override the internal addon styles

src/input-group/input-group.tsx
ghazwan-gt, line 129: The `custom` and `default` case is exactly the same except for the content. Let's make this as a render function instead

ghazwan-gt (approved):

src/input-group/input-group.styles.ts
```diff
@@ -0,0 +1,74 @@
...
+    display: flex;
+    align-items: center;
+    width: 100%;
+    padding: 0 ${Spacing["spacing-16"]};
+
+    &[data-read-only] {
+        padding: 0;
+    }
+
+    &[data-no-border] {
```
qroll: this would still apply if `readOnly={false}` or `noBorder={false}`

will need to explicitly check for `[data-x="true"]`

src/input-group/input-group-list-addon.tsx
```diff
@@ -324,8 +342,21 @@ export const Component = <T, V>(
...
+                        !readOnly &&
+                            noBorder &&
+                            (position === "right"
+                                ? styles.fieldInputNoBorderRight
+                                : styles.fieldInputNoBorderLeft),
+                        !readOnly &&
+                            !noBorder &&
+                            (position === "right"
+                                ? styles.fieldInputPositionRight
+                                : styles.fieldInputPositionLeft)
```
qroll: this is also a candidate for simplification using `data-position`

we could then have something like this for the final clsx
```
readOnly && styles.fieldInputReadOnly,
!readOnly && noBorder && styles.fieldInputNoBorder
```

qroll (changes requested):

src/input-group/input-group-list-addon.tsx
qroll: will be easier to apply a default to the positioning here rather than doing it for each instance below

```suggestion
    const { position = "left" } = addon!;
```

qroll (approved):

PR #1171, opened 2026-05-12
[BOOKINGSG-9244][RYN] migrate navbar component

ryan-nguyen-t:
**Checklist**

-   [x] Added visual tests


src/navbar/drawer.styles.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    ${MediaQuery.MaxWidth.sm} {
+        width: 100%;
+    }
+`;
+
+export const containerShown = css`
+    visibility: visible;
+    right: 0;
+    transition: all 300ms ${Motion["ease-entrance"]};
```
ghazwan-gt, line 45: @qroll is it possible to change this to one of our motion tokens? And is there any design reference we can check for animations?

e2e/nextjs-app/src/proxy.ts
```diff
@@ -22,6 +22,7 @@ export function proxy(request: NextRequest) {
         ${styleSrcDirective}
         img-src 'self' https://assets.life.gov.sg blob: data:;
         font-src 'self' https://assets.life.gov.sg;
+        connect-src 'self' https://cdn.jsdelivr.net/npm/@govtechsg/sgds-web-component@3/components/Masthead/index.umd.js.map;
```
qroll: tbh I was not expecting the need to whitelist the source file. is this causing any errors other than the CSP log?

ghazwan-gt:
image


The secondary brand is overflowing

src/navbar/navbar.tsx
```diff
@@ -252,7 +251,15 @@ const Component = <T,>(
     );
 
     const renderBrand = () => (
-        <NavBrandContainer $compress={compress} data-id="brand-container">
+        <div
+            className={clsx(
+                styles.navBrandContainer,
+                compress
+                    ? styles.navBrandContainerCompressed
+                    : styles.navBrandContainerFull
```
ghazwan-gt: The height of the container is being replaced by the one followed after, this makes us loses the height based on media query

ryan-nguyen-t:
@ghazwan-gt What screen size are you referring to for the overflowing brand
image

https://github.com/LifeSG/react-design-system/pull/1171#issuecomment-4438002563

ziggyzet (approved):

src/navbar/drawer.styles.ts
qroll: drawer component uses duration-800 + entrance/exit but tbh it looks a bit laggy? let me feedback to UX

https://designsystem.life.gov.sg/react/index.html?path=/docs/foundations-motion-default--docs

for now can just leave it

e2e/nextjs-app/src/app/components/navbar/common.ts
```diff
@@ -0,0 +1,66 @@
...
+];
+
+export const downloadActionButtons: NavbarButtonProps[] = [
+    { type: "download" },
+];
+
+export const ASSETS = {
+    LOGO_LIFESG: "https://assets.life.gov.sg/lifesg/logo-lifesg.svg",
+    LOGO_BOOKINGSG: "https://www.booking.gov.sg/logo.svg",
+};
```
qroll: we shouldn't need to specify custom assets; each theme should be pointing to their asset

as discussed during CR, to whitelist the domains in CSP accordingly

e2e/nextjs-app/src/app/components/navbar/common.ts
```diff
@@ -0,0 +1,66 @@
...
+    },
+    {
+        id: "guides",
+        children: "Guides",
+        href: "https://www.life.gov.sg/guides",
+    },
+    {
+        id: "lifesg-app",
+        children: "LifeSG app",
+        href: "https://apps.apple.com/sg/app/lifesg/id1383218758",
```
qroll: just use `https://www.life.gov.sg` all the way for these links, no need to use the actual prod ones

e2e/nextjs-app/src/app/components/navbar/custom-action-button.e2e.tsx
qroll, line 1: rename to `custom-action-button`. the key here is to test the custom component support, avatar is only an example on how the navbar can be used

e2e/tests/components/navbar/navbar.e2e.spec.ts
qroll, line 1: include the following tests:
- navbar with links but no action button
- navbar with no links but with action button
- navbar with custom link
- navbar with uncollapsible items
- all navbar themes - reference footer on how the tests are organised

include dark mode for:
- submenu mobile

src/navbar/brand.tsx
```diff
@@ -44,24 +44,27 @@ export const Brand = ({
...
+            <ImageWithFallback src={resources.logoSrc} alt={logoLabel} />
+        </a>
+    ) : (
+        <div
             data-id={dataId}
             data-testid={testId}
-            $type={type}
-            as={isClickable ? "a" : "div"}
-            {...props}
+            data-type={type}
```
qroll: data-type doesn't seem to be used

src/navbar/brand.tsx
```diff
@@ -44,24 +44,27 @@ export const Brand = ({
...
-        ? {
-              role: "link",
-              tabIndex: 0,
-              onClick: handleClick,
-          }
-        : {};
-
-    return (
-        <styles.Container
+    return isClickable ? (
```
qroll: instead of ternary, we could separate into different return statements instead. will reduce the nesting

```
if (isClickable) {
 return <a>
}

return <div>
```

src/navbar/drawer.styles.ts
qroll: can remain as hardcoded for now, because the navbar waits for a specific timing to ensure the drawer is fully in view. this doesn't cater for themeable design tokens. we can address this separately

src/navbar/navbar-items.styles.ts
```diff
@@ -68,9 +50,19 @@ export const LinkItem = styled.li<ItemStyleProps>`
     }
 `;
 
-const linkCss = css<{ $selected: boolean; weight: TypographyWeight }>`
-    ${(props) => Font[`body-md-${props.weight}`]}
+export const linkItemHiddenBranding = css`
+    &:first-child {
+        margin-left: -0.5rem;
```
qroll, line 55: restore the comment on the negative margin

src/navbar/navbar-items.styles.tsx
```diff
@@ -147,26 +145,30 @@ export const LinkIndicator = styled.div<StyleProps>`
...
+export const chevronIcon = css`
     height: 1.25rem;
     width: 1.25rem;
     color: ${Colour.icon};
-    &:hover {
-        ${(props) =>
-            props.$selected
-                ? Colour["icon-selected-hover"]
-                : Colour["icon-hover"]};
-    }
```
qroll, line 172: missing styles?

src/navbar/navbar-items.styles.ts
```diff
@@ -147,26 +145,30 @@ export const LinkIndicator = styled.div<StyleProps>`
     }
 `;
 
-export const LinkIconContainer = styled.div`
+export const linkIndicatorSelected = css`
+    &:hover {
+        background-color: ${Colour["border-selected-hover"]};
+    }
```
qroll, line 151: lol this is a weird style in hindsight

the hover should also apply when the parent `a` is hovered, right now it's only when user puts the cursor over this element, which is so small that most of the time there's no point

will track this in 9321

src/navbar/navbar.styles.ts
```diff
@@ -11,76 +9,86 @@ import {
...
+export const navElementsContainer = css`
     display: flex;
     height: 100%;
-    margin-left: ${(props) => (props.$hideNavBranding ? "0" : "5rem")};
     flex: 1;
     justify-content: flex-end;
-
     ${MediaQuery.MaxWidth.lg} {
         margin-left: 0rem;
     }
```
qroll: treat as 0 by default, this media query can be removed from this class

src/navbar/navbar.styles.ts
```diff
@@ -11,76 +9,86 @@ import {
...
+    ${MediaQuery.MinWidth.xl} {
+        margin-left: 5rem;
+    }
+`;
+
+export const navElementsContainerWithoutBranding = css`
+    ${MediaQuery.MinWidth.xl} {
+        margin-left: 0;
+    }
+`;
```
qroll: this class won't be needed since it's 0 by default

src/navbar/navbar.styles.ts
```diff
@@ -11,76 +9,86 @@ import {
...
         margin-left: 0rem;
     }
 `;
 
-export const MobileMenuButton = styled(ClickableIcon)`
+export const navElementsContainerWithBranding = css`
+    ${MediaQuery.MinWidth.xl} {
+        margin-left: 5rem;
+    }
+`;
```
qroll: prefer to use `MaxWidth` for consistency (desktop first approach)

qroll (changes requested):

src/navbar/navbar.styles.ts
```diff
@@ -0,0 +1,160 @@
...
+    height: ${ComponentToken.Navbar["full-height"]};
+`;
+
+export const navDarkCompressed = css`
+    height: calc(${ComponentToken.Navbar["compressed-height"]} - 1px);
+`;
+
+export const navDarkFull = css`
+    height: calc(${ComponentToken.Navbar["full-height"]} - 1px);
+`;
```
qroll: as discussed in CR, we can apply the height as a variable to reduce the number of combinations

src/navbar/navbar.tsx
```diff
@@ -252,7 +251,16 @@ const Component = <T,>(
...
 
     const renderBrand = () => (
-        <NavBrandContainer $compress={compress} data-id="brand-container">
+        <div
+            className={clsx(
+                styles.navBrandContainer,
+                styles.navBrandContainerResponsive,
+                compress
+                    ? styles.navBrandContainerCompressed
+                    : styles.navBrandContainerFull
```
qroll: put navBrandContainerResponsive last, following its source order, to reduce confusion. same for navSeparator

e2e/nextjs-app/src/proxy.ts
ryan-nguyen-t: yeah only the CSP error console log. I have removed

ghazwan-gt (approved):

e2e/tests/components/navbar/navbar.e2e.spec.ts
ryan-nguyen-t: - "navbar with no links but with action button" : I alr created "single-action-button", so I renamed it to align with the test cases.
The rest are updated!

src/navbar/navbar.styles.ts
qroll: there's no need to use `useApplyStyle` for this, you should be able to apply the variable via css as well

```
export const navCompressed = css`
    ${token.nav.height}: ${ComponentToken.Navbar["compressed-height"]};
`;

export const navDark = css`
   // ...
`
```

src/navbar/navbar.tsx
```diff
@@ -346,9 +346,9 @@ const Component = <T,>(
                         <div
                             className={clsx(
                                 styles.navElementsContainer,
-                                hideNavBranding
-                                    ? styles.navElementsContainerWithoutBranding
-                                    : styles.navElementsContainerWithBranding
+                                !hideNavBranding
+                                    ? styles.navElementsContainerWithBranding
+                                    : null
```
qroll:
```
!hideNavBranding && styles.navElementsContainerWithBranding
```

e2e/tests/components/navbar/__screenshots__/chromium/Navbar-Default-supportgowhere-theme--mount.png
qroll, line 1: hmmm why is the masthead missing in some cases?

src/navbar/navbar.styles.ts
qroll: was this updated?

ryan-nguyen-t: paiseh, updated now!

e2e/tests/components/navbar/__screenshots__/chromium/Navbar-Default-supportgowhere-theme--mount.png
ryan-nguyen-t: I think it was the issue with running parallel tests. Rerunning the test can gen the masthead. Sometimes the snapshots are not consistent...

ryan-nguyen-t: updated: rerunning can only capture masthead for some but not all missing test cases, so I added a function to wait for the masthead to render first

src/navbar/navbar.styles.ts
qroll: the margin-left can be totally removed as the div is not expected to have margin by default

e2e/tests/components/navbar/navbar.e2e.spec.ts
```diff
@@ -44,13 +44,22 @@ test.describe("Navbar", () => {
...
     // =======================================================================
 
+    const waitForMasthead = async (page: Page) => {
+        await page.waitForFunction(
+            () =>
+                !!customElements.get("sgds-masthead") &&
+                !!document.querySelector("sgds-masthead")?.shadowRoot
+                    ?.firstElementChild
+        );
+    };
```
qroll: you can move this to the StoryPage class

qroll (approved):

PR #1183, opened 2026-05-15
[MOL-20186][TIEN] Simplify modal viewport handling by removing offset…

tientranngoc-alt:
**Type of changes**

**Description of changes**

-   Link to [ticket](https://sgtechstack.atlassian.net/browse/MOL-20186)
-   Refactored modal viewport-stability orchestration into a reusable `useStableViewport` hook.
-   `useStableViewport(onStable, enabled)` now owns rAF polling + timeout fallback and triggers `onStable` once after viewport stabilizes (or fallback timeout), with full cleanup on close/unmount.
-   Simplified `Modal` logic by removing inline viewport polling/timer refs and using a lean effect for keyboard dismiss + animation reset.
-   Preserved the existing behavior goal: delay modal entrance animation until viewport height settles (to reduce top cut/jank around soft keyboard transitions).
-   Updated `e-signature` tests to wait for modal visibility asynchronously, aligning with animation-gated modal behavior.

**Checklist**

-   [x] Added/updated tests

### DIFF
--- a/src/modal-v2/modal-v2.styles.tsx
+++ b/src/modal-v2/modal-v2.styles.tsx
@@ -3,10 +3,8 @@ import { MediaQuery, Motion } from "../theme";
 import { ModalAnimationDirection } from "./types";
 
 interface Props {
-    $show: boolean;
     $animationFrom?: ModalAnimationDirection;
     $verticalHeight?: number;
-    $offsetTop?: number;
 }
 
 export const Container = styled.div<Props>`
@@ -17,17 +15,16 @@ export const Container = styled.div<Props>`
     overflow: auto;
 
     ${MediaQuery.MaxWidth.sm} {
-        ${(props) => {
-            return css`
-                height: calc(
-                    ${props.$verticalHeight
-                            ? `${props.$verticalHeight}px`
-                            : "1vh"} * 100
-                );
-            `;
-        }}
+        height: calc(
+            ${(props) =>
+                    props.$verticalHeight
+                        ? `${props.$verticalHeight}px`
+                        : "1vh"} * 100
+        );
+    }
 
-        top: ${(props) => props.$offsetTop || 0}px;
+    @supports (height: 100dvh) {
+        height: 100dvh;
     }
 
     ${(props) => css`
--- a/src/modal-v2/modal-v2.tsx
+++ b/src/modal-v2/modal-v2.tsx
@@ -5,7 +5,7 @@ import {
     useInteractions,
     useTransitionStatus,
 } from "@floating-ui/react";
-import React, { useEffect, useRef } from "react";
+import React, { useRef } from "react";
 import { Overlay } from "../overlay/overlay";
 import { useViewport } from "../shared/hooks";
 import { useEvent } from "../util";
@@ -34,7 +34,15 @@ export const ModalV2 = ({
     // =========================================================================
     // CONST, STATE, REF
     // =========================================================================
-    const { verticalHeight, offsetTop } = useViewport();
+    const dismissKeyboard = useEvent(() => {
+        if (dismissKeyboardOnShow) {
+            (document.activeElement as HTMLElement)?.blur?.();
+        }
+    });
+    const { verticalHeight } = useViewport({
+        enabled: show,
+        onBeforeStart: dismissKeyboard,
+    });
     const childRef = useRef<HTMLDivElement>(null);
     const childWithRef =
         children && React.cloneElement(children, { ref: childRef });
@@ -61,27 +69,12 @@ export const ModalV2 = ({
     const { getFloatingProps } = useInteractions([dismiss]);
 
     // =========================================================================
-    // EFFECTS
-    // =========================================================================
-    const dismissKeyboard = useEvent(() => {
-        if (dismissKeyboardOnShow) {
-            (document.activeElement as HTMLElement)?.blur?.();
-        }
-    });
-
-    useEffect(() => {
-        if (show) {
-            dismissKeyboard();
-        }
-    }, [show, dismissKeyboard]);
-
-    // =========================================================================
     // RENDER FUNCTIONS
     // =========================================================================
     return (
         <Overlay
             data-testid={`${testId}-overlay`}
-            show={isMounted}
+            show={show}
             enableOverlayClick={enableOverlayClick}
             onOverlayClick={onOverlayClick}
             id={id}
@@ -90,11 +83,9 @@ export const ModalV2 = ({
             zIndex={zIndex}
         >
             <Container
-                $show={isMounted}
                 $animationFrom={animationFrom}
                 data-testid={testId}
                 $verticalHeight={verticalHeight}
-                $offsetTop={offsetTop}
                 data-status={status}
                 {...otherProps}
             >
--- a/src/modal/modal.styles.tsx
+++ b/src/modal/modal.styles.tsx
@@ -4,15 +4,24 @@ import { MediaQuery } from "../theme";
 
 interface Props {
     $show: boolean;
+    $ready?: boolean;
     $animationFrom?: ModalAnimationDirection;
     $verticalHeight?: number;
-    $offsetTop?: number;
 }
 
 const visibilityStyle = (
     show: boolean,
-    animationFrom: ModalAnimationDirection
+    animationFrom: ModalAnimationDirection,
+    ready?: boolean
 ) => {
+    if (show && !ready) {
+        return `
+			${animationFrom}: -3%;
+			opacity: 0;
+			transition: none;
+		`;
+    }
+
     if (show) {
         return `
 			${animationFrom}: 0;
@@ -37,7 +46,13 @@ export const Container = styled.div<Props>`
     height: 100%;
     width: 100%;
     overflow: hidden;
-    ${(props) => visibilityStyle(props.$show, props.$animationFrom || "bottom")}
+
+    ${(props) =>
+        visibilityStyle(
+            props.$show,
+            props.$animationFrom || "bottom",
+            props.$ready
+        )}
 
     ${MediaQuery.MaxWidth.sm} {
         height: calc(
@@ -46,7 +61,5 @@ export const Container = styled.div<Props>`
                         ? `${props.$verticalHeight}px`
                         : "1vh"} * 100
         );
-
-        top: ${(props) => props.$offsetTop || 0}px;
     }
 `;
--- a/src/shared/hooks/useViewport.tsx
+++ b/src/shared/hooks/useViewport.tsx
@@ -1,13 +1,30 @@
-import throttle from "lodash/throttle";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 
-export const useViewport = () => {
+const STABLE_FRAMES = 2;
+const HEIGHT_OFFSET = 0.5;
+// This 180 ms duration is chosen to align with the mobile Chrome/keyboard animation (150–300 ms range) and the modal’s entrance `transition-delay` (200 ms) before we force it into the `ready` state.
+const FALLBACK_TIME = 180;
+
+interface UseViewportProps {
+    enabled?: boolean;
+    onBeforeStart?: () => void;
+}
+
+export const useViewport = ({
+    enabled = false,
+    onBeforeStart,
+}: UseViewportProps = {}) => {
     const [verticalHeight, setVerticalHeight] = useState<number>();
-    const [offsetTop, setOffsetTop] = useState<number>();
+    const [ready, setReady] = useState(false);
+    const pendingFramesRef = useRef<number[]>([]);
+    const timeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+    const hasResolvedRef = useRef(false);
+    const onBeforeStartRef = useRef(onBeforeStart);
 
-    // =============================================================================
-    // EVENT HANDLERS
-    // =============================================================================
+    useEffect(() => {
+        onBeforeStartRef.current = onBeforeStart;
+    }, [onBeforeStart]);
+
     const handleWindowResize = useCallback(() => {
         const newVerticalHeight = window.innerHeight * 0.01;
         setVerticalHeight(newVerticalHeight);
@@ -17,7 +34,6 @@ export const useViewport = () => {
         if (window.visualViewport) {
             const newVerticalHeight = window.visualViewport.height * 0.01;
             setVerticalHeight(newVerticalHeight);
-            setOffsetTop(window.visualViewport.offsetTop);
         }
     }, []);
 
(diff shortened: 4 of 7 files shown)


src/modal/modal.styles.tsx
```diff
@@ -18,7 +17,7 @@ const visibilityStyle = (
 			${animationFrom}: 0;
 			opacity: 1;
 			transition: all 300ms cubic-bezier(0.21, 0.79, 0.53, 1);
-			transition-delay: 200ms;
+			transition-delay: 50ms;
```
qroll: why are we changing the transition delay here?

tientranngoc-alt: I changed this to shorten the time before the modal appears. With a 200ms delay, there is still a visible tick/frame where the modal briefly stutters before settling into its correct position. Lowering it to 50ms avoids that.

qroll: but the overlay itself takes 200ms to animate in

setting this means the user is seeing 50ms of animation on the modal only. that doesn't sound right

maybe look into tweaking this animation instead. or see how we can resolve the root cause of the repositioning

tientranngoc-alt:
updated PR

src/modal/modal.styles.tsx
```diff
@@ -4,15 +4,24 @@ import { MediaQuery } from "../theme";
...
 
 const visibilityStyle = (
     show: boolean,
-    animationFrom: ModalAnimationDirection
+    animationFrom: ModalAnimationDirection,
+    shouldAnimate?: boolean
 ) => {
+    if (show && !shouldAnimate) {
+        return `
+			${animationFrom}: 0;
```
qroll: why 0? this basically removes the positioning animation

now the final effect looks different without keyboard. the modal is supposed to slide in from the bottom

https://github.com/user-attachments/assets/e2f820fe-63a7-464e-a910-226ad33ff8e8

src/modal/modal.styles.tsx
qroll, line 1: also apply the changes to modal-v2

src/shared/hooks/useStableViewport.ts
```diff
@@ -0,0 +1,87 @@
+import { useEffect, useRef } from "react";
+
+const STABLE_FRAMES = 2;
+const HEIGHT_OFFSET = 0.5;
+const FALLBACK_TIME = 180;
+
+export const useStableViewport = (
+    onStable: () => void,
+    enabled: boolean
```
qroll: prefer to provide arguments as an object instead

src/shared/hooks/useStableViewport.ts
```diff
@@ -0,0 +1,87 @@
+import { useEffect, useRef } from "react";
+
+const STABLE_FRAMES = 2;
+const HEIGHT_OFFSET = 0.5;
+const FALLBACK_TIME = 180;
+
+export const useStableViewport = (
```
qroll: can this be merged into `useViewport`?

src/shared/hooks/useStableViewport.ts
```diff
@@ -0,0 +1,87 @@
+import { useEffect, useRef } from "react";
+
+const STABLE_FRAMES = 2;
+const HEIGHT_OFFSET = 0.5;
```
qroll: why 0.5? that's a sub-pixel value

src/shared/hooks/useStableViewport.ts
```diff
@@ -0,0 +1,87 @@
...
+        const waitForStableViewport = () => {
+            let lastHeight =
+                window.visualViewport?.height ?? window.innerHeight;
+            let stableFrames = 0;
+
+            const checkViewportStability = () => {
+                const currentHeight =
+                    window.visualViewport?.height ?? window.innerHeight;
+
+                if (Math.abs(currentHeight - lastHeight) < HEIGHT_OFFSET) {
```
qroll: add a comment on this code to explain why we need to check the diff against HEIGHT_OFFSET

src/shared/hooks/useStableViewport.ts
```diff
@@ -0,0 +1,87 @@
...
+            return clearPending;
+        }
+
+        if (!window.visualViewport) {
+            scheduleFrame(resolve);
+            return clearPending;
+        }
+
+        waitForStableViewport();
+        timeoutRef.current = window.setTimeout(resolve, FALLBACK_TIME);
```
qroll:
```suggestion
        timeoutRef.current = setTimeout(resolve, FALLBACK_TIME);
```

src/shared/hooks/useStableViewport.ts
```diff
@@ -0,0 +1,87 @@
+import { useEffect, useRef } from "react";
+
+const STABLE_FRAMES = 2;
+const HEIGHT_OFFSET = 0.5;
+const FALLBACK_TIME = 180;
```
qroll: why 180ms only?

src/shared/hooks/useStableViewport.ts
qroll, line 1: rename file to `use-stable-viewport.ts`

qroll (changes requested):

src/shared/hooks/useStableViewport.ts
tientranngoc-alt: 0.5px was chosen because `visualViewport` height can fluctuate a bit (0.1px - 0.4px) between frames even when the viewport hasn’t actually changed; 0.5 is enough to filter out that noise.

src/shared/hooks/useStableViewport.ts
tientranngoc-alt: we will apply the changes in modal-v2 so i will merge and refactor the `useViewport`

src/modal/modal.styles.tsx
tientranngoc-alt: i saw the issue, decrease this to -5% can resolve this.

tientranngoc-alt: https://github.com/user-attachments/assets/a3da351d-c2d1-439e-8c7a-f738c5676268

src/shared/hooks/useStableViewport.ts
tientranngoc-alt: if stability detection doesn’t resolve after ~2 stable frames, we still enable animation so the modal doesn’t get stuck; 180ms is close to typical mobile keyboard settle time without delaying UX too much.

tientranngoc-alt:
addressed comment !

src/shared/hooks/useStableViewport.ts
qroll: how did you derive this value? do add a comment in the code to explain

src/shared/hooks/useViewport.tsx
```diff
@@ -17,48 +33,108 @@ export const useViewport = () => {
...
+            let stableFrames = 0;
+
+            const checkViewportStability = () => {
+                const currentHeight =
+                    window.visualViewport?.height ?? window.innerHeight;
+                /*
+                    Compare the viewport height across consecutive frames; 
+                    once it remains stable for at least 2 frames, trigger the modal animation.               
+                */
+                if (Math.abs(currentHeight - lastHeight) < HEIGHT_OFFSET) {
```
qroll, line 101:
```suggestion
        // Compare the viewport height across consecutive frames;
        // once it remains stable for at least 2 frames, trigger the modal animation
        const waitForStableViewport = () => {
            let lastHeight =
                window.visualViewport?.height ?? window.innerHeight;
            let stableFrames = 0;

            const checkViewportStability = () => {
                const currentHeight =
                    window.visualViewport?.height ?? window.innerHeight;

                // Use a small offset to account for fluctuations in reported height
                // even when the viewport hasn’t actually changed
                if (Math.abs(currentHeight - lastHeight) < HEIGHT_OFFSET) {
```

src/modal-v2/modal-v2.styles.tsx
```diff
@@ -1,54 +1,64 @@
...
-            ${props.$animationFrom}: 0;
-            transition: all ${Motion["duration-250"]} ${Motion["ease-entrance"]};
-            transition-delay: ${Motion["duration-150"]};
-        }
-
-        &[data-status="close"] {
-            opacity: 0;
-            ${props.$animationFrom}: -3%;
-            transition: all ${Motion["duration-250"]} ${Motion["ease-exit"]};
-        }
```
qroll, line 50: we shouldn't be replacing the `data-status` handling for modal v2 wholesale. this is still needed to handle the animation transition from Floating UI properly

you'll need to reconcile this - you could try to synchronize `shouldAnimate` with a local state and use that for `useFloating` `open` value instead

src/shared/hooks/useViewport.tsx
```diff
@@ -17,48 +33,108 @@ export const useViewport = () => {
...
+        waitForStableViewport();
+        timeoutRef.current = setTimeout(resolve, FALLBACK_TIME);
+
+        return clearPending;
+    }, [enabled]);
 
     return {
         verticalHeight,
-        offsetTop,
+        shouldAnimate,
```
qroll:
```suggestion
        /** Viewport height is stable */
        ready,
```

src/shared/hooks/useViewport.tsx
```diff
@@ -1,13 +1,29 @@
...
+import { useCallback, useEffect, useRef, useState } from "react";
 
-export const useViewport = () => {
+const STABLE_FRAMES = 2;
+const HEIGHT_OFFSET = 0.5;
+const FALLBACK_TIME = 180;
+
+interface UseViewportProps {
+    enabled?: boolean;
+    onBeforeGate?: () => void;
```
qroll:
```suggestion
    onBeforeStart?: () => void;
```

qroll (changes requested):

src/modal-v2/modal-v2.tsx
```diff
@@ -90,16 +99,16 @@ export const ModalV2 = ({
...
                 $animationFrom={animationFrom}
                 data-testid={testId}
                 $verticalHeight={verticalHeight}
-                $offsetTop={offsetTop}
                 data-status={status}
                 {...otherProps}
             >
                 <ModalContext.Provider value={{ onClose }}>
-                    {isMounted && (
+                    {shouldRender && (
```
qroll: this should still follow `isMounted` because it depends on floating ui

src/modal-v2/modal-v2.tsx
```diff
@@ -60,28 +82,15 @@ export const ModalV2 = ({
...
+    const shouldRender = show || isMounted;
 
     // =========================================================================
     // RENDER FUNCTIONS
     // =========================================================================
     return (
         <Overlay
             data-testid={`${testId}-overlay`}
-            show={isMounted}
+            show={shouldRender}
```
qroll: we can inline this for now

```suggestion
            show={show || isMounted}
```

src/modal-v2/modal-v2.styles.tsx
```diff
@@ -17,19 +31,21 @@ export const Container = styled.div<Props>`
...
+        height: calc(
+            ${(props) =>
+                    props.$verticalHeight
+                        ? `${props.$verticalHeight}px`
+                        : "1vh"} * 100
+        );
     }
 
+    ${(props) =>
+        visibilityStyle(
```
qroll: can remove this, as we can fully rely on the transition status now

src/modal-v2/modal-v2.styles.tsx
```diff
@@ -4,11 +4,25 @@ import { ModalAnimationDirection } from "./types";
 
 interface Props {
     $show: boolean;
```
qroll: the `$show` prop is actually unused, so to reduce confusion we can remove it

src/modal/modal.styles.tsx
```diff
@@ -4,15 +4,24 @@ import { MediaQuery } from "../theme";
...
 
 const visibilityStyle = (
     show: boolean,
-    animationFrom: ModalAnimationDirection
+    animationFrom: ModalAnimationDirection,
+    ready?: boolean
 ) => {
+    if (show && !ready) {
+        return `
+			${animationFrom}: -5%;
```
qroll: checking if there's any difference using `-3%` vs `-5%`?

src/shared/hooks/useViewport.tsx
```diff
@@ -17,48 +33,107 @@ export const useViewport = () => {
...
             const newVerticalHeight = window.visualViewport.height * 0.01;
             setVerticalHeight(newVerticalHeight);
-            setOffsetTop(window.visualViewport.offsetTop);
         }
     }, []);
 
     useEffect(() => {
-        // set initial vh
-
-        // use VisualViewport API if available, it gives more accurate dimensions when iOS software keyboard is active
```
qroll, line 27: restore this comment

qroll (changes requested):
this is the story I'm using to test in simulator. but I'm still seeing the jumpy modal in storybook

do the new changes have any effect on the app?

```
export const Default: StoryObj<Component> = {
    render: () => {
        const [show, setShow] = useState(false);
        const openModal = () => setShow(true);
        const closeModal = () => setShow(false);
        return (
            <div style={{ height: "200vh" }}>
                <div style={{ height: "20vh" }} />
                <Button.Default onClick={openModal}>
                    Click to open
                </Button.Default>
                <input style={{ fontSize: 18, padding: 16 }} />
                <ModalV2
                    show={show}
                    onOverlayClick={closeModal}
                    onClose={closeModal}
                >
                    <ModalV2.Card>
                        <ModalV2.CloseButton />
                        <ModalV2.Content>
                            <Typography.BodyBL>
                                Lorem ipsum dolor sit amet consectetur
                                adipisicing elit. Totam debitis explicabo rerum
                                velit quod recusandae, cum odio inventore
                                repellendus non quas quis praesentium suscipit,
                                soluta incidunt officiis omnis, quae error!
                            </Typography.BodyBL>
                        </ModalV2.Content>
                        <ModalV2.Footer
                            primaryButton={
                                <Button.Default>Log out</Button.Default>
                            }
                            secondaryButton={
                                <Button.Default styleType="secondary">
                                    Continue
                                </Button.Default>
                            }
                        />
                    </ModalV2.Card>
                </ModalV2>
            </div>
        );
    },
};
```

src/modal-v2/modal-v2.tsx
qroll: resolving as we found a different approach for v2

src/modal-v2/modal-v2.tsx
qroll: resolving as we found a different approach for modal v2

src/modal-v2/modal-v2.styles.tsx
qroll: resolving due to different approach for v2

src/modal/modal.styles.tsx
tientranngoc-alt: hmm this maybe my mismatch, should be -3%

src/modal-v2/modal-v2.tsx
```diff
@@ -60,28 +72,13 @@ export const ModalV2 = ({
...
-    }, [show, dismissKeyboard]);
-
     // =========================================================================
     // RENDER FUNCTIONS
     // =========================================================================
     return (
         <Overlay
             data-testid={`${testId}-overlay`}
-            show={isMounted}
+            show={show || isMounted}
```
qroll:
```suggestion
            show={show}
```

src/modal-v2/modal-v2.tsx
```diff
@@ -34,11 +34,23 @@ export const ModalV2 = ({
...
+        onBeforeStart: dismissKeyboard,
+    });
     const childRef = useRef<HTMLDivElement>(null);
     const childWithRef =
         children && React.cloneElement(children, { ref: childRef });
 
+    // =========================================================================
+    // EFFECTS
+    // =========================================================================
+
```
qroll: remove this unused section header

qroll (changes requested):

qroll (approved):

PR #1196, opened 2026-05-20
[BOOKINGSG-9334][GZ] migrate time slot bar

ghazwan-gt:
**Checklist**

-   [X] Added visual tests

**Additional information**
- Moved out `TimeSlotItem` and `TimeSlotDivider` to be on its own file
- Simplified arrow rendering function


src/file-upload/file-upload.styles.tsx
```diff
@@ -19,7 +19,7 @@ export const TextContainer = styled.div`
 export const Title = styled(Typography.BodyBL)``;
 
 export const TitleContainer = styled(Markup)`
-    color: ${V3_Colour.text};
+    color: ${V3_Colour.text({} as any)};
```
ziggyzet: was this change blocking the PR to be commited?

ghazwan-gt: My bad, I used it for an experiment and commit it by mistake

src/time-slot-bar/time-slot-divider.tsx
```diff
@@ -0,0 +1,40 @@
...
+const TimeSlotDivider = ({
+    variant,
+    left,
+    dataTestId,
+}: TimeSlotDividerProps) => {
+    const dividerRef = useRef<HTMLDivElement>(null);
+
+    useApplyStyle(dividerRef, {
+        [styles.tokens.border.left]:
+            left === undefined ? undefined : `${left}px`,
```
qroll: can use the `formatUnitValue` util here

src/time-slot-bar/time-slot-item.tsx
```diff
@@ -0,0 +1,52 @@
...
+    slotOffset,
+    slotWidth,
+    variant,
+}: TimeSlotItemProps) => {
+    const itemRef = useRef<HTMLDivElement>(null);
+
+    useApplyStyle(itemRef, {
+        [styles.tokens.item.width]: `${slotWidth}px`,
+        [styles.tokens.item.left]: `${slotOffset}px`,
+        [styles.tokens.item.color]: color || Colour.text,
```
qroll, line 32: i initially had a hard time tracing this, it's strange that we set the token on the parent but only use it in the child

in another PR you used `color: inherit`, maaybe this method can be applied here as well?

src/time-slot-bar/time-slot-bar.tsx
```diff
@@ -376,115 +375,111 @@ const Component = (props: TimeSlotBarProps, ref: React.Ref<TimeSlotBarRef>) => {
...
-                    focusOutline="none"
-                    onClick={() => {
-                        handleArrowButtonClick("right");
-                    }}
-                >
-                    <ArrowIconRight />
-                </ArrowButton>
-            );
-        }
+        if (!shouldShowArrow) return null;
```
qroll: let's move this check up, before the direction-specific styling is applied

src/time-slot-bar/time-slot-bar.styles.ts
```diff
@@ -0,0 +1,181 @@
...
+    ${tokens.item.width}: initial;
+
+    position: absolute;
+    left: var(${tokens.item.left});
+    width: var(${tokens.item.width});
+
+    > * {
+        width: 100%;
+        height: 100%;
+    }
```
qroll: hmm why are we using the universal selector here? would prefer to target the elements specifically

qroll (changes requested):

src/time-slot-bar/time-slot-divider.tsx
ghazwan-gt: Right, forgot about that one

src/time-slot-bar/time-slot-item.tsx
ghazwan-gt: Oh, nice catch!! Why didn't think of this lol

src/time-slot-bar/time-slot-bar.tsx
ghazwan-gt: Sure

src/time-slot-bar/time-slot-bar.styles.ts
ghazwan-gt: To be honest, not sure about this since it was from the previous code. But let me provide pass a classname to the children instead

src/time-slot-bar/time-slot-divider.tsx
qroll: update divider too!

src/time-slot-bar/time-slot-bar.styles.ts
```diff
@@ -0,0 +1,183 @@
...
+`;
+
+export const arrowIconRight = css`
+    width: 1rem;
+    height: 1rem;
+    color: ${Colour["icon-primary"]};
+`;
+
+export const arrowIconLeft = css`
+    font-size: 1rem;
```
zhaoyanxzy: should align this with `arrowIronRight` in sizing the svg

ghazwan-gt: Let me take a look at the figma for this

zhaoyanxzy (approved):

PR #1209, opened 2026-05-22
[BOOKINGSG-9265][GZ] improve theme breakpoint utils

ghazwan-gt:
**Type of changes**

**Description of changes**

[Improve breakpoint token utilities](https://sgtechstack.atlassian.net/browse/BOOKINGSG-9265)
-  Improve the breakpoint hooks to be SSR-safe
- Add storybook documentation on how to use the hooks
- Add e2e test to assert the hooks works on SSR
- Combine the previous media query specificity e2e test to be with the new test
- Update some import path on the theme to prevent circular dependency 

**Additional Info**
The next follow-up to this changes is to:
- Update all `react-responsive` & `useSafeMaxWidthMediaQuery` usage to the new hooks
- Remove `react-responsive` package
- Delete `useSafeMaxWidthMediaQuery` utils

**Checklist**

-   [X] Updated documentation
-   [X] Added/updated tests


qroll:
landscape MediaQuery tokens for use in styles - will they be included in this PR?

ghazwan-gt:
> landscape MediaQuery tokens for use in styles - will they be included in this PR?

Oh right, forgot to add this one. On it

qroll:
> > landscape MediaQuery tokens for use in styles - will they be included in this PR?
> 
> Oh right, forgot to add this one. On it

actually, it may not be needed after we changed how breakpoint detection works

ESignature needed `(orientation: landscape) and (max-height: 480px)` because on mobile devices, the device height is still treated as the screen height rather than the screen width when rotated to landscape

but in theme provider we are now checking against `window.innerWidth` - I do not know if that rotates accordingly

src/theme/utils/use-media-query.ts
```diff
@@ -0,0 +1,147 @@
...
+        minWidth: `${breakpoint}-min`,
+        orientation,
+    });
+
+export const useMaxWidthMediaQuery = (
+    breakpoint: BreakpointName,
+    orientation?: MediaQueryOrientation
+) =>
+    useMediaQuery({
+        maxWidth: `${breakpoint}-max`,
```
qroll: there's no xxl max, so we need to restrict the typing here

src/theme/utils/use-media-query.ts
```diff
@@ -0,0 +1,147 @@
...
+    const resolvedMaxWidth = useResolveBreakpointToken(
+        getWidthCandidate(maxWidth)
+    );
+    const normalizedMinWidth = normalizeCssLengthValue(resolvedMinWidth);
+    const normalizedMaxWidth = normalizeCssLengthValue(resolvedMaxWidth);
+    const clauses = [
+        getMediaQueryClause("min-width", normalizedMinWidth),
+        getMediaQueryClause("max-width", normalizedMaxWidth),
+        orientation ? `(orientation: ${orientation})` : undefined,
+    ].filter((value): value is string => Boolean(value));
```
qroll:
```suggestion
    ].filter((value): value is string => !!value);
```

src/theme/tokens/media-query.ts
qroll, line 1: the landscape media query token was originally intended for esignature's custom media query, so we don't actually need this

src/theme/utils/use-resolve-breakpoint-token.ts
```diff
@@ -0,0 +1,71 @@
...
+        : undefined;
+    const resolvedTokenValue = useResolvedTokenValue<
+        BreakpointCSSVariableString,
+        string
+    >({
+        value: cssVariableToken,
+        fallback: fallbackLength ?? breakpointOrLength,
+        isToken: isBreakpointToken,
+        normalizeCustom: (value) => value,
+    });
```
qroll: actually why can't we use `useResolvedTokenValue` directly? that was done previously for the dashed border utils to handle radius/width so this seems pretty similar

```
useResolvedTokenValue({
    value: breakpointOrLength,
    fallback: fallbackLength ?? breakpointOrLength,
    isToken: isBreakpointToken,
    normalizeCustom: (value) => value,
})
```

stories/theme/breakpoint/a-breakpoint-introduction.mdx
```diff
@@ -21,6 +21,7 @@ and **media query** tokens. For grid-based page layouts, **column number** and
 ## Collections
 
 [Default](/docs/foundations-breakpoint-default--docs)
+[Media Query](/docs/foundations-breakpoint-media-query--docs)
```
qroll: collections refer to the theme-specific sets, merge the docs into this file under the `Usage` section instead

we can split this page into two sub-sections: CSS-based style declarations and runtime utilities

src/theme/utils/use-media-query.ts
```diff
@@ -0,0 +1,145 @@
...
+    const queryString = clauses.join(" and ");
+    const hasMinWidth = Boolean(normalizedMinWidth);
+    const hasMaxWidth = Boolean(normalizedMaxWidth);
+    const defaultMatch = getDefaultMatch(hasMinWidth, hasMaxWidth);
+    const [matches, setMatches] = useState(() =>
+        getCurrentMatch(queryString, defaultMatch)
+    );
+
+    useEffect(() => {
+        setMatches(getCurrentMatch(queryString, defaultMatch));
```
qroll: we're calling `setMatches` twice in the main effect, could we only limit to 1?

e2e/tests/components/theme/__screenshots__/chromium/Theme-media-query-Specificity-Enable-override-media-query-specificity--layout-container-flex-column-specificity.png
qroll, line 1: since the tests have been moved to a different file, remove the previous screenshots generated under `theme.e2e.spec.ts`

e2e/nextjs-app/src/app/components/theme/media-query.e2e.tsx
```diff
@@ -0,0 +1,104 @@
...
+    useOrientation,
+} from "@lifesg/react-design-system/theme";
+import { useEffect, useState } from "react";
+
+const boolToText = (value: boolean) => String(value);
+
+export default function Story() {
+    const [dimensions, setDimensions] = useState("0 x 0");
+
+    const minXs = useMinWidthMediaQuery("xs");
```
qroll, line 17: include xxs as well (edge case)

e2e/tests/components/theme/media-query.e2e.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+            await expect(story.locators.orientationPortraitStatus).toHaveText(
+                "false"
+            );
+        });
+
+        test("portrait orientation detection", async ({ story }) => {
+            await story.init("media-query", {
+                size: "mobile",
+                orientation: "portrait",
+            });
```
qroll: I think the same `size` should be used here to ensure the values actually change

qroll (changes requested):

src/theme/utils/use-resolve-breakpoint-token.ts
ghazwan-gt: Well, we can do that, but doing so will restrict the value to be a css variable name. Now that I think about it more, using the library means using the defined breakpoint when accessing the theme. Will update this

qroll: I was under the impression that `useResolvedTokenValue` still lets you pass in a non-token value

src/theme/tokens/media-query.ts
qroll: I see that `useOrientation` was removed but this is still here

I thought it would be the other way round, since we intend to do runtime detection for esignature

e2e/tests/components/theme/media-query.e2e.spec.ts
qroll: resolving for now since `useOrientation` was removed

src/theme/utils/use-resolve-breakpoint-token.ts
qroll: on this note, instead of `BreakpointCSSVariableString` we should also limit to the `<size>-min` and `<size-max`

because the column/gutter/margins are not relevant to media query

src/theme/tokens/media-query.ts
ghazwan-gt: I don't think`useOrientation` still has it use since it will be used only by esignature, hence the removal. It can be handled in the component itself later.

I missed deleting the one in this file, will remove

qroll: it feels awkward for the component to have to register its own matchMedia listener when there's already a hook. I would suggest to extend `useMediaQuery` to support custom clauses in that case then, when working on the esignature component

qroll:
Layout ColDiv lg test is failing due to the change in orientation. should we revert the changes to the e2e utils as well?

src/theme/utils/use-resolve-breakpoint-token.ts
ghazwan-gt: Ah right, I forgot that it include those as well. Will be updating the type

qroll (approved):

PR #1211, opened 2026-05-22
[BOOKINGSG-9327][IT] Migrate masked input component

iant-gds:
**Checklist**

-   [x] Added visual tests


src/masked-input/masked-input.style.tsx
```diff
@@ -41,63 +40,56 @@ export const IconContainer = styled.div`
...
     svg {
         height: 1em;
         width: 1em;
     }
 `;
 
-// -----------------------------------------------------------------------------
-// LOADING DISPLAY
-// -----------------------------------------------------------------------------
-export const LoadingWrapper = styled.div`
```
qroll, line 59: keep the original headers to organise the classes

src/masked-input/masked-input.styles.ts
```diff
@@ -108,19 +100,19 @@ export const ClickableErrorWrapper = styled(BasicButton)`
     cursor: pointer;
 
     ${Font["body-baseline-regular"]}
+`;
 
-    &:hover, &:active, &:focus {
-        ${TryAgainLabel} {
+export const clickableErrorWrapperHover = css`
```
qroll: why does this need to be separated from `clickableErrorWrapper`?

src/masked-input/masked-input.tsx
```diff
@@ -280,8 +297,17 @@ const Component = (
         }
 
         return (
-            <InputGroupWrapper
+            <InputGroup
                 ref={inputRef}
+                className={clsx(
+                    styles.inputGroupWrapper,
+                    styles.inputGroupWrapperInput,
```
qroll: if there's no conditional display for this this we can merge the two classes together as well. or check if `inputGroupWrapperInput` is even needed in the first place

tests/masked-input/masked-input.spec.tsx
```diff
@@ -179,4 +179,58 @@ describe("MaskedInput", () => {
...
+    });
+
+    describe("Disable masking", () => {
+        it("should not toggle masking when icon is clicked and disableMask is true", () => {
+            render(<MaskedInput value="S1234567D" disableMask />);
+
+            expect(screen.getByDisplayValue("S1234567D")).toBeInTheDocument();
+
+            const icon = screen.queryByTestId("icon-masked");
+            if (icon) {
```
qroll: we should know if icon exists ahead of time since this is a static test with known props

avoid conditionals in unit tests because they are prone to fail silently

tests/masked-input/masked-input.spec.tsx
```diff
@@ -179,4 +179,58 @@ describe("MaskedInput", () => {
...
+                    value=""
+                    transformInput="uppercase"
+                    onChange={onChangeFn}
+                />
+            );
+
+            const input = screen.getByTestId("input");
+            fireEvent.change(input, { target: { value: "Test input" } });
+
+            expect(onChangeFn).toHaveBeenCalled();
```
qroll: can assert directly that `onChangeFn` was called with the expected arguments

e2e/nextjs-app/src/app/components/masked-input/custom-mask-icons.e2e.tsx
qroll, line 1: should be more explicit on what kind of variants these are

suggest to rename this one to `custom-mask-icon`

e2e/nextjs-app/src/app/components/masked-input/form-variants.e2e.tsx
qroll, line 1: can we include the placeholder text as well?

e2e/tests/components/masked-input/masked-input.e2e.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+        test("Grid layout", async ({ story }) => {
+            await compareScreenshot(story, "mount");
+        });
+    });
+
+    test.describe(() => {
+        const now = new Date(fixedTimestamp).getTime();
+
+        test.beforeEach(async ({ story }) => {
+            await story.page.clock.install({ time: now - 30 * 1000 });
```
qroll: hmm why is the timer needed here?

e2e/tests/components/masked-input/masked-input.e2e.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+        test("Other variants", async ({ story }) => {
+            await compareScreenshot(story, "mount");
+        });
+
+        test("Focus variants", async ({ story }) => {
+            await test.step("Focus custom icon input", async () => {
+                await story.locators.customIconFormMasked
+                    .getByTestId("input")
+                    .focus();
+                await compareScreenshot(story, "disabled-masking", {
```
qroll: this is labelled `disabled-masking` but the component doesn't use the `disableMask` prop or have the disabled state. what is being tested here?

e2e/tests/components/masked-input/__screenshots__/chromium/Masked-Input-Loading-and-error-display--mount.png
qroll, line 1: ahhh I see that the auto focus is causing the blue focus ring to show by default

could we include a duplicate error display so we can see the error display without the focus ring?

qroll (changes requested):

e2e/tests/components/masked-input/masked-input.e2e.spec.ts
iant-gds: ahh.. forgot to remove it after adding the disabled masking variant to unit tests. Have removed it

e2e/tests/components/masked-input/masked-input.e2e.spec.ts
iant-gds: i migrated another component with the spinner animation and had a comment to add in the timer there (maybe to prevent flakiness..?), i can remove it if really not needed

src/masked-input/masked-input.style.tsx
qroll: there are more headers other than loading display, restore them as well

e2e/tests/components/masked-input/masked-input.e2e.spec.ts
qroll: since the loading/error state are specified by static props, there's nothing async running so manipulating the clock is not needed here

the spinner is achieved via css animations which are paused by Playwright so we also don't need to specifically pause timers for it

e2e/tests/components/masked-input/masked-input.e2e.spec.ts
qroll: o I see. hmm `disableMask` has a slightly different styling for the mask icon, so we could still include this variant as part of the visual tests (no need for interaction test)

e2e/tests/components/masked-input/__screenshots__/chromium/Masked-Input-Focus-variants--disabled-masking.png
qroll, line 1: to remove if no longer in use

src/masked-input/masked-input.tsx
```diff
@@ -280,8 +294,16 @@ const Component = (
...
         return (
-            <InputGroupWrapper
+            <InputGroup
                 ref={inputRef}
+                className={clsx(
+                    styles.inputGroupWrapper,
+                    readOnly && styles.inputGroupWrapperReadOnly,
+                    readOnly &&
+                        !shouldDisableMasking() &&
+                        styles.inputGroupWrapperInputPointer,
```
ghazwan-gt: I'm not sure if I'm understanding this correctly, but I don't think we need these readonly-based styling as readonly will only render button looking at the condition at line 257

iant-gds: ahh.. you're right

qroll: good catch, I think that was missed out when readonly was switched to be button-based

`inputGroupWrapperReadOnly` and `inputGroupWrapperInputPointer` can be cleaned up

qroll (approved):

PR #1213, opened 2026-05-24
[BOOKINGSG-9331][IT] Migrate unit number component

iant-gds:
**Checklist**

    ~-   [ ] User style prop is set as CSS variable~
-   [x] Added visual tests


src/unit-number/unit-number-input.tsx
```diff
@@ -405,13 +409,11 @@ export const UnitNumberInput = ({
             readOnly={readOnly}
             tabIndex={-1}
             onBlur={handleNodeBlur}
+            className={className}
         >
             <HashContainer
                 data-testid="addon"
-                className={clsx(
-                    inputGroupStyles.labelAddonContainer,
-                    className
```
qroll, line 413: good catch!

src/unit-number/unit-number-input.styles.ts
```diff
@@ -1,23 +1,21 @@
...
 import { Colour } from "../theme";
-import { Typography } from "../typography";
 
-export const HashContainer = styled.div`
+export const hashContainer = css`
     margin-right: 0.5rem;
 `;
 
-export const FloorInput = styled(Input)`
+export const floorInput = css`
```
qroll: hmm we can split this into a generic `input` and `floorInput` with the custom width, since this is partially shared by both components

src/unit-number/unit-number-input.tsx
qroll, line 1: can we update the existing unit tests to cover the following behaviour?

- `onChange` and `onChangeRaw` callback behaviour
    - when only floor input is changed
    - when only unit number input is changed
- `onBlur`/`onBlurRaw` behaviour with keyboard navigation

qroll (changes requested):

ghazwan-gt (approved):

qroll (approved):

PR #1232, opened 2026-06-02
[BOOKINGSG-9335][IT] Migrate time-slot-bar-week component

iant-gds:
**Checklist**

-   [x] Added visual tests


e2e/tests/components/time-slot-bar-week/time-slot-bar-week.e2e.spec.ts
qroll, line 1: include the following test cases:
- disabled date in header
- min/max date range in header
- hovered, selected, selected + hover states in header date
- half slot cell display (top half only, bottom half only, both (spanning three cells))

e2e/tests/components/time-slot-bar-week/time-slot-bar-week.e2e.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+
+class StoryPage extends AbstractStoryPage {
+    protected readonly component = "time-slot-bar-week";
+
+    public readonly locators: {
+        navigationHeader: {
+            monthButton: Locator;
+            yearButton: Locator;
+        };
+        expandCollapseButton: Locator;
```
qroll: put under an `internals` object

e2e/tests/components/time-slot-bar-week/time-slot-bar-week.e2e.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+    });
+
+    test.describe(() => {
+        test.beforeEach(async ({ story }) => {
+            await story.init("max-height", { mockedTimestamp: fixedTimestamp });
+        });
+
+        test("Max height", async ({ story }) => {
+            await test.step("Initial render", async () => {
+                await compareScreenshot(story, "mount");
```
qroll:
```suggestion
                await compareScreenshot(story, "collapsed");
```

src/time-slot-bar-week/time-slot-bar-week.styles.ts
```diff
@@ -0,0 +1,13 @@
+import { css } from "@linaria/core";
+
+import { MediaQuery } from "../theme";
+
+export const wrapper = css`
+    --header-bottom-spacing: 0.5rem;
```
qroll: this is supposed to refer to calendar manager's variable which I think was refactored to a token constant. can help to update this?

src/time-slot-bar-week/time-slot-bar-week-days.styles.ts
```diff
@@ -0,0 +1,189 @@
...
+    margin: 0;
+    border-radius: ${Radius["xs"]};
+
+    &:focus-within {
+        outline: 2px solid ${Colour["focus-ring"]};
+        outline-offset: -2px;
+    }
+`;
+
+const halfFillStripesRepeatingGradient = `repeating-linear-gradient(
```
qroll: look into a way to refactor the styles here to fully reuse TimeSlot component instead of repeating the striped styling style

src/time-slot-bar-week/time-slot-bar-week-days.styles.ts
```diff
@@ -0,0 +1,189 @@
+import { css } from "@linaria/core";
+
+import { Colour, Font, Motion, Radius, Spacing } from "../theme";
+
+export const tokens = {
+    timeColumn: {
+        height: "--fds-internal-timeSlotBarWeek-timeColumn-height",
+    },
+    timeSlotComponent: {
```
qroll:
```suggestion
    timeSlot: {
```

src/time-slot-bar-week/time-slot-bar-week-days.tsx
```diff
@@ -598,19 +607,36 @@ export const TimeSlotBarWeekDays = ({
             !expandAll &&
             (slot.rowIndex ?? 0) >= visibleRowCount;
 
+        const slotHeight =
+            variant === "fixed"
+                ? cellLength * SLOT_HEIGHT + (cellLength - 1) * SLOT_GAP
+                : SLOT_HEIGHT;
+
+        const slotRef = useRef<HTMLDivElement>(null);
```
qroll: hmm we shouldn't be using hooks directly in loops

extract this to a standalone React function component instead

qroll (changes requested):

src/time-slot-bar-week/time-slot-bar-week-days.tsx
```diff
@@ -639,62 +652,64 @@ export const TimeSlotBarWeekDays = ({
...
-            </TimeSlotComponent>
+            </styles.TimeSlotComponent>
         );
     };
 
     const renderTimeSlotBarCells = () => {
         return (
-            <Expandable style={expandableStyles} role="row">
-                <ColumnWeekCell
+            <styles.Expandable style={expandableStyles} role="row">
```
qroll: there is a CSP warning. we can add a `isMounted` flag to only apply the style when the component is rendered on client

src/time-slot-bar-week/time-slot-bar-week-days.styles.ts
iant-gds: not sure if my changes was what you had in mind

e2e/tests/components/time-slot-bar-week/time-slot-bar-week.e2e.spec.ts
iant-gds: Coupled some of the test cases into the same screenshot:
- disabled date + half slot cell display: `TimeSlotBarWeek-Default-prefilled--mount.png`

src/time-slot-bar-week/time-slot-bar-week-days.styles.ts
qroll: not quite, looking for the following
1. to follow the TimeSlot stripe width which is 6px
2. use the TimeSlot component directly without having to specify additional CSS for the colour styling (so ideally we just need to specify the `bgColor` etc. props)

my suggestion is to enhance `TimeSlot` to take in a `fill?: "default" | "top" | "bottom"` prop and render the `background`  using the same method in `timeSlotComponentHalfFillTop` and `timeSlotComponentHalfFillBottom`. it's possible to chain the background like this:

```
background: var(--half-fill-top), var(--half-fill-bottom), linear-gradient(...),
```

src/time-slot-bar-week/time-slot-bar-week-days.tsx
```diff
@@ -567,18 +593,36 @@ export const TimeSlotBarWeekDays = ({
...
+                            </div>
+                        </div>
                     ))}
-            </TimeColumn>
+            </div>
         );
     };
-    const renderSlotCell = (formattedDate: string, slot: TimeSlotCell) => {
+
+    const WeekTimeSlotCell = ({
```
qroll: move this declaration to be at the top-level instead of nesting within another component. handle the ref forwarding accordingly

src/time-slot-bar-week/time-slot-bar-week-days.tsx
```diff
@@ -598,22 +642,31 @@ export const TimeSlotBarWeekDays = ({
...
+                : SLOT_HEIGHT;
+
+        const slotRef = useRef<HTMLDivElement>(null);
+        useApplyStyle(slotRef, {
+            [styles.tokens.timeSlot.height]: formatUnitValue(slotHeight, "px"),
+        });
+
         return (
             <TimeSlotComponent
                 key={id}
```
qroll: there is a unique key warning in the console. since this is now a standalone react component the key should be set by the parent iterator, not inside of the component

e2e/tests/components/time-slot-bar-week/time-slot-bar-week.e2e.spec.ts
qroll: let's separate out the disabled date in header to a standalone story instead

e2e/tests/components/time-slot-bar-week/__screenshots__/chromium/TimeSlotBarWeek-Min-and-Max-Date--dropdown.png
qroll, line 1: hmm my original intention was to cover the `DayCell` styling in the header and to ensure it appears greyed out when the date is out of range

this month/year display is a shared component and the behaviour is well-covered by the date inputs and calendar, so we don't need to check these specifically

qroll (changes requested):

src/time-slot-bar-week/time-slot-bar-week.tsx
ghazwan-gt, line 13: Not part of the changes, but it would be nice to adjust the comment on the types to match the date form

src/time-slot-bar-week/time-slot-bar-week-days.tsx
```diff
@@ -94,7 +186,9 @@ export const TimeSlotBarWeekDays = ({
     const dateFormat = "YYYY-MM-DD";
     const [expandAll, setExpandAll] = useState<boolean>(false);
     const [hoverDay, setHoverDay] = useState<Dayjs>();
+    const [isMounted, setIsMounted] = useState<boolean>(false);
```
qroll: can use the `useIsMounted` hook from the utils

src/time-slot-bar-week/time-slot-bar-week-days.tsx
```diff
@@ -68,10 +57,113 @@ interface FocusableSlotMeta {
...
+    slotButtonRefs: React.MutableRefObject<
+        Record<string, HTMLButtonElement | null>
+    >;
+}
+
 const SLOT_HEIGHT = 12; // px
 const SLOT_GAP = 4; // px
 const ROW_HEIGHT = SLOT_HEIGHT + SLOT_GAP;
 
+const WeekTimeSlotCell = ({
```
qroll: this file is getting quite long, we can extract this component to a `time-slot-cell.tsx`

the common consts can be moved into a `consts.ts`

src/shared/time-slot/time-slot.styles.ts
```diff
@@ -60,3 +62,41 @@ export const timeSlot = css`
         }
     }
 `;
+
+export const timeSlotHalfFillTop = css`
```
qroll, line 66: this won't work with the hover style but we can enhance this separately since it's currently not needed for the TimeSlotBarWeek component

also noticed the grey background is actually 1 shade darker than the empty cells as well. probably should expose a "base" cell background colour as well

will create a ticket to cover both of these enhancements

src/time-slot-bar-week/time-slot-bar-week-days.styles.ts
```diff
@@ -0,0 +1,125 @@
...
+export const chevronIcon = css`
+    transform: rotate(180deg);
+    transition: transform ${Motion["duration-250"]} ${Motion["ease-default"]};
+`;
+
+export const chevronIconExpanded = css`
+    transform: rotate(0deg);
+`;
+
+export const timeSlotComponent = css`
```
qroll, line 119: to restore the xs `border-radius` and `&:focus-within` styles

qroll (changes requested):

qroll (approved):

PR #1282, opened 2026-06-19
[CCUBE-2158][IT] Suppress useTheme error internally

iant-gds:
**Type of changes**

**Description of changes**

-   Link to [ticket](https://sgtechstack.atlassian.net/browse/CCUBE-2158)
-   Allow components to render when not wrapped in `ThemeProvider` by returning default values

**Checklist**

-   [ ] ~Looks good on mobile and tablet~
-   [ ] ~Updated documentation~
-   [ ] ~Added/updated tests~

### DIFF
--- a/src/theme/theme-provider/hooks.tsx
+++ b/src/theme/theme-provider/hooks.tsx
@@ -3,6 +3,7 @@ import { useContext, useMemo } from "react";
 
 import { getInheritedInlineCssVariables } from "../utils/css-variable";
 import { ThemeContext } from "./context";
+import type { ThemeContextValue } from "./types";
 
 type InheritedThemeScope = {
     themeProps: {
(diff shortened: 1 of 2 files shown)


src/theme/theme-provider/hooks.tsx
```diff
@@ -16,7 +16,11 @@ export const useTheme = () => {
     const context = useContext(ThemeContext);
 
     if (!context) {
-        throw new Error("useTheme must be used within ThemeProvider");
+        return {
+            theme: "lifesg" as const,
+            mode: "light" as const,
```
ghazwan-gt: Instead of doing this, how about adding the return type to the hooks?

src/theme/utils/use-design-token.ts
```diff
@@ -16,9 +16,12 @@ export const useDesignToken = (
     const [value, setValue] = useState<string | undefined>(undefined);
 
     useEffect(() => {
-        if (!themeElement) return;
+        const element =
+            themeElement ??
+            (typeof document !== "undefined" ? document.documentElement : null);
```
qroll: thinking there's no need to fallback in this case (documentElement does not receive the styles); if the element is not available then the design token can remain undefined

qroll (approved):

PR #1290, opened 2026-06-22
[CCUBE-2102] Resolve to fallback if token is not available

qroll:
**Type of changes**

**Description of changes**

-   Resolves a scenario where `useResolvedTokenValue({ value: "var(--fds-x)" })` returns `"var(--fds-x)"` instead of the fallback when the variable is not successfully evaluated
- This allows media queries to be resolved using the hardcoded defaults

**Checklist**

-   [ ] ~~Added/updated E2E tests~~

### DIFF
--- a/src/theme/utils/use-design-token.ts
+++ b/src/theme/utils/use-design-token.ts
@@ -3,7 +3,6 @@ import { useEffect, useState } from "react";
 import { useTheme } from "../theme-provider/hooks";
 import type { CSSVariableString } from "../types";
 import { parseCSSVariableValue } from "./css-variable";
-import { getResolvedValue } from "./token-resolver";
 
 const isEmptyValue = (value: unknown) => {
     return value == null || value === "";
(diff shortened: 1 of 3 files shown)


src/theme/utils/use-design-token.ts
```diff
@@ -53,13 +52,24 @@ export const useResolvedTokenValue = <
...
 
-    if (tokenString) {
-        return getResolvedValue(resolvedTokenValue, tokenString);
+    if (isToken(value)) {
+        if (!isEmptyValue(resolvedToken)) {
+            return resolvedToken;
+        }
+    } else if (!isEmptyValue(value)) {
+        return normalizeCustom(value);
     }
```
ghazwan-gt: Lets use `tokenString` to avoid repeating the same computation. Also we can flatten the nested condition to be a single line with early return

```suggestion
    if (tokenString && !isEmptyValue(resolvedToken)) {
        return resolvedToken;
    }

    if (!tokenString && !isEmptyValue(value)) {
        return normalizeCustom(value);
    }
```

src/theme/utils/use-design-token.ts
```diff
@@ -53,13 +52,24 @@ export const useResolvedTokenValue = <
...
 
-    return normalizeCustom(effectiveValue as TCustom);
+    if (isToken(fallback)) {
+        if (!isEmptyValue(resolvedFallback)) {
+            return resolvedFallback;
+        }
+    }
```
ghazwan-gt:
```suggestion
    if (fallbackString && !isEmptyValue(resolvedFallback)) {
        return resolvedFallback;
    }
```

tests/theme/utils/use-design-token.spec.tsx
```diff
@@ -210,10 +215,13 @@ describe("useResolvedTokenValue", () => {
         });
     });
 
-    it("returns fallback when value is empty", async () => {
+    it("resolves fallback values when value is empty", async () => {
```
ghazwan-gt: Lets also add a case where `value` is a token, but the resolved value is empty

ghazwan-gt: Lets also add a case where both `value` and `fallback` is empty. In this case, it should be `undefined`

src/theme/utils/use-design-token.ts
qroll: would prefer to use `isToken` directly as it's more meaningful (if the provided value is a token, then ..., else ...)

I could do something like this instead

```
    const isAToken = isToken(value);
    const tokenName = isAToken ? value : undefined;
    const resolvedToken = useDesignToken(tokenName);

    if (isAToken && !isEmptyValue(resolvedToken)) {
        return resolvedToken;
    }
    
    if (!isAToken && !isEmptyValue(value)) {
        return normalizeCustom(value);
    }
```

ghazwan-gt: Yeah, that could works

ghazwan-gt (approved):

PR #1291, opened 2026-06-23
[CCUBE-2164][GZ] handle custom large font size on calendar

ghazwan-gt:
**Type of changes**

-   [X] Tests (improvements to unit tests or E2E tests)

**Description of changes**

[[A11y] Improve Calendar display with large font](https://sgtechstack.atlassian.net/browse/CCUBE-2164)
Handle the case when the some content of the calendar is not accessible when setting the browser font size to large. The fix is to make the calendar content scrollable when it is overflowing.

**Checklist**

-   [X] Looks good on mobile and tablet
-   [ ] Updated documentation
-   [X] Added/updated E2E tests


e2e/nextjs-app/src/app/components/date-input/custom-font-size.e2e.tsx
```diff
@@ -0,0 +1,21 @@
+"use client";
```
ziggyzet, line 1: i think if there isn't any difference as to how we create the component, we don't need to make a new file here and instead can just use the default date-inputs

ghazwan-gt: Oooh that makes sense. Will use the `default` one

ziggyzet (approved):

e2e/tests/components/date-input/__screenshots__/chromium/DateInput-With-Custom-Browser-Font-Size--after.png
qroll, line 1: we can remove the `Custom-Browser-Font-Size` files since the test cases have been renamed

qroll (changes requested):

e2e/nextjs-app/src/app/components/date-input/custom-font-size.e2e.tsx
```diff
@@ -0,0 +1,35 @@
...
+                data-testid="date-input"
+                value={selectedDate}
+                onChange={setSelectedDate}
+            />
+            <p
+                data-testid="selected-value"
+                className={styles["selected-value"]}
+            >
+                {selectedDate || "none"}
+            </p>
```
qroll: can remove this selected value display since it is unused by the test, same for the parent `div`. `useState` is not needed as well. we can trim this story to be a simple JSX component

qroll (approved):

PR #1293, opened 2026-06-23
[MISC][GZ] silence wyw-in-js warnings in storybook

ghazwan-gt:
**Type of changes**

## Changes
[[Storybook] wyw-in-js storybook warnings](https://sgtechstack.atlassian.net/browse/CCUBE-2166)

### Root Cause
When wyw-in-js statically evaluates CSS-in-JS at build time, it walks the import chain and encounters:
- `dayjs` and its plugins which are CJS-only modules that can't be statically analyzed
- Internal ts files (`src/theme/tokens/*` and `src/theme/utils/font.ts`) that don't export evaluated values needed by CSS-in-JS

### Solution
Added `importOverrides` [configuration](https://wyw-in-js.dev/configuration#options) to the `@wyw-in-js/webpack-loader` options in `.storybook/main.ts`, allowing (skip) these unresolvable imports with `{ unknown: "allow" }`.

**Checklist**

-   [ ] Updated documentation

**Screenshots**
The final output
image

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -40,6 +40,45 @@ const config: StorybookConfig = {
                     loader: "@wyw-in-js/webpack-loader",
                     options: {
                         sourceMap: process.env.NODE_ENV !== "production",
+                        importOverrides: {
+                            dayjs: { unknown: "allow" },
+                            "dayjs/plugin/customParseFormat": {
+                                unknown: "allow",
+                            },
+                            "dayjs/plugin/isBetween": { unknown: "allow" },
+                            "dayjs/plugin/isSameOrAfter": { unknown: "allow" },
+                            "dayjs/plugin/isSameOrBefore": { unknown: "allow" },
+                            "dayjs/plugin/timezone": { unknown: "allow" },
+                            "./src/theme/tokens/spacing.ts": {
+                                unknown: "allow",
+                            },
+                            "./src/theme/tokens/border.ts": {
+                                unknown: "allow",
+                            },
+                            "./src/theme/tokens/breakpoint.ts": {
+                                unknown: "allow",
+                            },
+                            "./src/theme/tokens/colour.ts": {
+                                unknown: "allow",
+                            },
+                            "./src/theme/tokens/component.ts": {
+                                unknown: "allow",
+                            },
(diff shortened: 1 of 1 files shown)


ziggyzet (approved):
LGTM!

.storybook/main.ts
```diff
@@ -40,6 +40,45 @@ const config: StorybookConfig = {
                     loader: "@wyw-in-js/webpack-loader",
                     options: {
                         sourceMap: process.env.NODE_ENV !== "production",
+                        importOverrides: {
```
iant-gds, line 43: `importOverrides` seems to be able to support wildcards, not sure if we wanna do `dayjs/**` and `./src/theme/tokens/*` instead?

ghazwan-gt: Oh nice catch! Will use wildcard instead

qroll:
hmm still getting warnings for both `dayjs` and `src/theme`. not sure if I'm missing anything, have tried clearing the storybook cache.

the first commit works for `dayjs` but still emits warnings for `src/theme`

Screenshot 2026-06-24 at 10 58 24 AM

also some clarifying questions regarding these:

> When wyw-in-js statically evaluates CSS-in-JS at build time, it walks the import chain and encounters:
> 
> dayjs and its plugins which are CJS-only modules that can't be statically analyzed
> Internal ts files (src/theme/tokens/* and src/theme/utils/font.ts) that don't export evaluated values needed by CSS-in-JS

1. does this mean wyw-in-js can't handle CJS modules in general?
2. what does "evaluated values needed" mean exactly? `./src/theme/tokens/motion.ts` is used in at least one component's styles, for example, and it still shows up in the warnings

iant-gds (approved):

ghazwan-gt:
@qroll 
> hmm still getting warnings 

my bad, was not including the nested dir under `src/theme`

> does this mean wyw-in-js can't handle CJS modules in general?


Yes, wyw-in-js can't handle CJS modules because it's ESM-only and relies on static analysis of import statements. CJS uses dynamic require() which breaks that static analysis.

> what does "evaluated values needed" mean exactly? 
This is referring to the string/number value to be extracted to be put in the css file. The issue lies in the import chain actually, `dayjs` and its plugin are cjs. And we also need to allow the files that depends on it transitively

PR #1304, opened 2026-07-01
[CCUBE-2180][GZ] check and report e2e stale snapshots

ghazwan-gt:
**Type of changes**

**Description of changes**
 Link to [ticket](https://sgtechstack.atlassian.net/browse/CCUBE-2180)

Adding a script to check for any stale snapshots. The approach is:
- Add annotation for each snapshot
- Add a custom reporter to generate a manifest based on annotation
- Add a script to check for any snapshot not present in the manifest, marked those as stale
- Include e2e run for changed only on pre-push for local guard

When any stale snapshots is present, the pre-push and CI pipeline would be stopped, and marked as failed.
> [!CAUTION]
> Pre-push check only works when the branch has an upstream branch reference. When it is not present, the checker won't run. This is due to playwright need the branch to compare the changes diff, otherwise it will run the whole suite which is not ideal.

**Checklist**

-   [ ] Updated documentation

**Screenshots** 

Pipeline https://github.com/LifeSG/react-design-system/actions/runs/30613329864/job/91102159301?pr=1304. 

---

Prepush Guard
image

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -73,6 +73,7 @@ jobs:
         with:
           name: screenshot-manifest-${{ matrix.shard }}
           path: .screenshot-manifest-${{ matrix.shard }}.txt
+          include-hidden-files: true
           retention-days: 1
 
   merge-reports:
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -102,11 +102,11 @@ jobs:
           merge-multiple: true
       - name: Merge reports
         run: npx playwright merge-reports --reporter=github,html ./blob-reports
-      - name: Check for stale screenshots
-        if: ${{ needs.test.result == 'success' }}
-        run: npm run test-e2e:check-stale-screenshots -- --manifest='screenshot-manifests/.screenshot-manifest-*.txt'
       - uses: actions/upload-artifact@v7
         with:
           name: playwright-report
           path: playwright-report/
           retention-days: 14
+      - name: Check for stale screenshots
+        if: ${{ needs.test.result == 'success' }}
+        run: npm run test-e2e:check-stale-screenshots -- --manifest='screenshot-manifests/.screenshot-manifest-*.txt'
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -104,7 +104,7 @@ jobs:
         run: npx playwright merge-reports --reporter=github,html ./blob-reports
       - name: Check for stale screenshots
         if: ${{ needs.test.result == 'success' }}
-        run: npx tsx scripts/check-stale-screenshots.ts --manifest='screenshot-manifests/.screenshot-manifest-*.txt'
+        run: npm run check-stale-screenshots -- --manifest='screenshot-manifests/.screenshot-manifest-*.txt'
       - uses: actions/upload-artifact@v7
         with:
           name: playwright-report
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -72,7 +72,7 @@ jobs:
         if: ${{ !cancelled() }}
         with:
           name: screenshot-manifest-${{ matrix.shard }}
-          path: .screenshot-manifest-*.txt
+          path: .screenshot-manifest-${{ matrix.shard }}.txt
           retention-days: 1
 
   merge-reports:
@@ -102,6 +102,7 @@ jobs:
       - name: Merge reports
         run: npx playwright merge-reports --reporter=github,html ./blob-reports
       - name: Check for stale screenshots
+        if: ${{ needs.test.result == 'success' }}
         run: npx tsx scripts/check-stale-screenshots.ts --manifest='screenshot-manifests/.screenshot-manifest-*.txt'
       - uses: actions/upload-artifact@v7
         with:
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -104,7 +104,7 @@ jobs:
         run: npx playwright merge-reports --reporter=github,html ./blob-reports
       - name: Check for stale screenshots
         if: ${{ needs.test.result == 'success' }}
-        run: npm run check-stale-screenshots -- --manifest='screenshot-manifests/.screenshot-manifest-*.txt'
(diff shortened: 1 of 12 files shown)


.github/workflows/playwright.yml
```diff
@@ -84,8 +94,17 @@ jobs:
...
+        with:
+          pattern: screenshot-manifest-*
+          path: screenshot-manifests/
+          merge-multiple: true
       - name: Merge reports
         run: npx playwright merge-reports --reporter=github,html ./blob-reports
+      - name: Check for stale screenshots
+        if: ${{ needs.test.result == 'success' }}
+        run: npx tsx scripts/check-stale-screenshots.ts --manifest='screenshot-manifests/.screenshot-manifest-*.txt'
       - uses: actions/upload-artifact@v7
```
qroll: would this still be able to run if the `Check for stale screenshots` step failed?

.husky/pre-push
```diff
@@ -1,4 +1,9 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-npm run test && npm run build
+npx concurrently --kill-others-on-fail --group --timings \
+  --names "test+build,e2e" \
+  --prefix "[{name}]" \
+  --prefix-colors "blue,magenta" \
+  "jest --reporters=summary --coverageReporters=none && npm run build" \
```
qroll: hmm why disable coverage here?

e2e/tests/utils.ts
```diff
@@ -182,20 +196,20 @@ export const compareScreenshot = async (
...
-        threshold: 0.01, // Strict colour matching
-        maxDiffPixelRatio: 0.01, // Allow a small percentage of pixels to differ
-        maxDiffPixels: 50,
-        mask: options?.mask,
+    await expect
+        .soft(target)
+        .toHaveScreenshot(`${name}.png`, screenshotOptions);
+
+    test.info().annotations.push({
+        type: "screenshot-used",
```
qroll: this annotation shows up in the final report, is that intended?

[this implementation](https://github.com/matrix-org/matrix-react-sdk/pull/12743/changes#diff-fa202c5ca8bf17b4afd7827765221390319d109bd3cb83cb7aa1c1d6d2a4c389R375) used a `_` prefix to hide it from the final report:

e2e/screenshot-manifest-reporter.ts
```diff
@@ -0,0 +1,187 @@
...
+    const end = length - middle.length - start;
+    return s.substring(0, start) + middle + s.slice(-end);
+}
+
+/**
+ * Mirrors Playwright's internal sanitizeFilePathBeforeExtension utility.
+ * Sanitizes only the base name portion of a filename, preserving the extension
+ * so that e.g. "non-sticky open.png" becomes "non-sticky-open.png".
+ */
+function sanitizeFilePathBeforeExtension(
```
qroll: also seeing from that example repo that playwright has a `snapshotPath` [util](https://playwright.dev/docs/api/class-testinfo#test-info-snapshot-path) that returns the formatted file name

scripts/e2e/pre-push.sh
```diff
@@ -0,0 +1,33 @@
+# scripts/e2e-prepush.sh
```
qroll: unnecessary comment

scripts/e2e/pre-push.sh
```diff
@@ -0,0 +1,33 @@
...
+}
+
+if ! REMOTE_REF=$(resolve_changed_ref); then
+	echo "[e2e] No upstream ref found — skipping E2E. Set an upstream with: git branch --set-upstream-to=<remote>/<branch>"
+	exit 0
+fi
+
+echo "[e2e] Running changed-file E2E between HEAD and $REMOTE_REF"
+npm run test-e2e:setup
+CI=true npx playwright test --only-changed="$REMOTE_REF"
```
qroll: can run in non-CI mode, else this would need to build lib + install + build app again

scripts/e2e/pre-push.sh
```diff
@@ -0,0 +1,33 @@
...
+npx tsx scripts/check-stale-screenshots.ts
```
qroll: could pull this into an npm script

qroll (commented):
initial comments

.github/workflows/playwright.yml
ghazwan-gt: The pipeline would stop right after checking stale screenshots in that case. I don't think there's any value in uploading the artifacts

.husky/pre-push
ghazwan-gt: Oops, forgot to remove it after testing

e2e/tests/utils.ts
ghazwan-gt: Hmm running it on UI mode doesn't really matters since the annotations has its own tab. But will take a look at the other forms of the report

e2e/screenshot-manifest-reporter.ts
ghazwan-gt: Oh nice! We can use this one instead

e2e/tests/utils.ts
ghazwan-gt: image
Yeah, having annotations like this is awkward, will hide it instead

ghazwan-gt: image
It is now hidden

.github/workflows/playwright.yml
qroll: if there's failure, then we would still need to be able to download the tests artifacts

package.json
```diff
@@ -28,6 +28,8 @@
         "test-e2e:docker": "docker compose up --build",
         "test-e2e:setup": "./scripts/e2e-setup.sh",
         "test-e2e-ci": "npm run test-e2e:setup && ./scripts/e2e-ci-build.sh && ./scripts/e2e-ci-test.sh",
+        "test-e2e:changed": "npm run test-e2e:setup && playwright test --only-changed",
+        "check-stale-screenshots": "tsx scripts/check-stale-screenshots.ts",
```
qroll: we can prefix this command too

```suggestion
        "test-e2e:check-stale-screenshots": "tsx scripts/check-stale-screenshots.ts",
```

scripts/e2e/pre-push.sh
qroll: hmm let's use `#!/bin/bash` here instead

qroll (approved):

PR #1306, opened 2026-07-01
[CCUBE-2157][RYN] add CSS presets for token reuse

ryan-nguyen-t:


qroll (commented):
1. is it intended to have both `<theme>.source.css` files and `<theme>.css` in the source code? would prefer to have just `<theme>.css` and process it during the build
2. could we arrange the presets into sub folders? for `component-tokens`, `fonts`, `semantic-colour`
3. instead of running postcss twice in the `copy` plugin, could we use `libStylePlugin` directly with postcss plugins?

src/theme/styles/presets/component-tokens/default.css
```diff
@@ -0,0 +1,38 @@
+@define-mixin component-tokens-standard {
+    --fds-button-radius: var(--fds-radius-sm);
```
qroll, line 3:
```suggestion
    /* Button */  
    --fds-button-radius: var(--fds-radius-sm);
```

src/theme/styles/source/bookingsg.css
```diff
@@ -0,0 +1,142 @@
...
+
+    @mixin semantic-colours-light-standard;
+
+    /* 
+    ========================================================================
+    COMPONENT TOKENS
+    ========================================================================
+    */
+
+    /* Button */
```
qroll: this should go in the mixin declaration instead

src/theme/styles/source/a11y-playground.source.css
```diff
@@ -0,0 +1,184 @@
+@import url("../default.css");
+@import url("../presets/font-scale-standard.css");
+@import url("../presets/semantic-colours-dark.css");
+@import url("../presets/semantic-colours-light-standard.css");
+@import url("../presets/shadow.css");
```
qroll: have you tried the postcssMixins option for the plugin? would that work to auto apply the mixins instead of having to manually include the imports for each file?

https://github.com/postcss/postcss-mixins#mixinsdir

src/theme/styles/ccube.css
qroll, line 1: let's introduce a single `font-open-sans.css` for lifesg/rbs/ccube to use directly, and remove from `default.css`

src/theme/styles/presets/component-tokens/default.css
qroll, line 1: rename to `component-tokens/default.css`

src/theme/styles/presets/fonts/scale-default.css
qroll, line 1: rename to `fonts/default-scale.css`

src/theme/styles/source/a11y-playground.source.css
qroll: might not work with this, so can stick to naming the file according to the mixin name

> could we arrange the presets into sub folders?

src/theme/styles/source/lifesg.css
```diff
@@ -0,0 +1,207 @@
...
+
+    /* Footer: */
+    --fds-footer-colour-bg: var(--fds-colour-bg-strong);
+    --fds-footer-colour-text: var(--fds-colour-text);
+    --fds-footer-link-colour-text: var(--fds-colour-text);
+    --fds-footer-link-colour-text-hover: var(--fds-colour-text-hover);
+    --fds-footer-disclaimer-link-colour-text: var(--fds-colour-text);
+    --fds-footer-disclaimer-link-colour-text-hover: var(
+        --fds-colour-text-subtler
+    );
```
qroll: seems to be identical to the default semantic tokens, can be removed

src/theme/styles/source/bookingsg.css
qroll: to remove this `/* Button */` header from all the files using `component-tokens-standard`

.storybook/main.ts
qroll, line 1: was discussed to include postcss-loader for storybook webpack so that the css stylesheets can be processed at runtime

are the changes going to be included?

qroll:
nextjs app is also not able to support postcss in local dev

will create a long lived branch and split the build-related issues into a separate task

PR #1309, opened 2026-07-03
[CCUBE-2182][GZ] component jsdocs batch 3

ghazwan-gt:
**Type of changes**

**Description of changes**
[Annotate components and interfaces with JSDocs](urlhttps://sgtechstack.atlassian.net/browse/CCUBE-2182)
- Added JSDocs to these components
  - countdown-timer
  - dashed-border
  - data-table
  - date-input
  - date-navigator

**Checklist**

-   [X] Updated documentation

~**Screenshots**~

### DIFF
--- a/src/countdown-timer/countdown-timer.tsx
+++ b/src/countdown-timer/countdown-timer.tsx
@@ -16,6 +16,12 @@ import * as styles from "./countdown-timer.styles";
 import type { CountdownTimerProps } from "./types";
 import { useTimer } from "./use-timer";
 
+/**
+ * Displays a live countdown of time remaining until a deadline.
+ *
+ * Use when a time-limited interaction — such as a session timeout or a timed
+ * form — needs a persistent visual reminder.
+ */
 export const CountdownTimer = ({
     className,
     align = "right",
--- a/src/countdown-timer/types.ts
+++ b/src/countdown-timer/types.ts
@@ -12,39 +12,83 @@ interface MobileOffset {
     top?: number | undefined;
 }
 
+/** Base props shared by both the `timer`-based and `timestamp`-based variants. */
 interface CountdownTimerBaseProps extends React.HTMLAttributes<HTMLDivElement> {
     "data-testid"?: string | undefined;
-    /** To show/play the countdown timer */
+    /** Controls whether the timer is visible and actively counting down. */
     show: boolean;
+    /**
+     * When `true`, the timer switches to a fixed viewport overlay once its
+     * inline position scrolls out of view, keeping the countdown visible at
+     * all times.
+     *
+     * @default true
+     */
     fixed?: boolean | undefined;
-    /** Specifies a timer (in seconds) for notifications */
+    /**
+     * Remaining-seconds threshold at which the warning state activates.
+     * Once the countdown reaches this value, `onNotify` fires once and
+     * `onTick` fires every second until the timer finishes.
+     */
     notifyTimer?: number | undefined;
-    /** Allows customization of the sticky position in tablet/desktop view */
+    /**
+     * Fine-tunes the fixed overlay position on tablet and desktop viewports.
+     *
+     * @remarks `top` defaults to 168 when not provided; `left`/`right`
+     * fall back to the bounding-rect of the inline wrapper.
+     */
     offset?: Offset | undefined;
-    /** Allows customization of the sticky position in mobile only */
+    /**
+     * Fine-tunes the fixed overlay position on mobile viewports.
+     *
+     * @default { top: 80 }
+     */
     mobileOffset?: MobileOffset | undefined;
-    /** Specifies if the component is aligned to its left or right in its sticky position */
+    /**
+     * Side of the inline wrapper the fixed overlay is pinned to when
+     * scrolled into sticky mode.
+     *
+     * @default "right"
+     */
     align?: AlignPosition | undefined;
-    /** Called every second with the remaining time left when the timer value falls below the `notifyTimer` threshold */
+    /**
+     * Called every second with the remaining seconds once the countdown falls
+     * at or below `notifyTimer`.
+     */
     onTick?: ((seconds: number) => void) | undefined;
-    /** Called when the timer value falls below the specified `notifyTimer` threshold */
+    /**
+     * Called once when the remaining seconds first reach the `notifyTimer`
+     * threshold.
+     *
+     * @remarks Fires at most once per timer session; resets when `show` or
+     * `timer`/`timestamp` changes.
+     */
     onNotify?: (() => void) | undefined;
-    /** Called when countdown reaches 0 */
+    /** Called when the countdown reaches zero. */
     onFinish?: (() => void) | undefined;
-    /** Specifies the interval (in seconds) between polite announcements of the remaining time left */
+    /**
+     * Interval in seconds between polite screen-reader announcement of the
+     * remaining time while the timer is above the `notifyTimer`
+     * threshold.
+     *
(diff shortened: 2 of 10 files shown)


src/countdown-timer/types.ts
```diff
@@ -12,39 +12,83 @@ interface MobileOffset {
...
+     */
     onNotify?: (() => void) | undefined;
-    /** Called when countdown reaches 0 */
+    /** Called when the countdown reaches zero. */
     onFinish?: (() => void) | undefined;
-    /** Specifies the interval (in seconds) between polite announcements of the remaining time left */
+    /**
+     * The interval (in seconds) of polite screen-reader announcement of the
+     * remaining time is made while the timer is above the `notifyTimer`
+     * threshold.
```
zhaoyanxzy: took me a while to understand this part "... of ... of the... ". Is this better:
```suggestion
     * Interval in seconds between polite screen-reader announcement of the
     * remaining time while the timer is above the `notifyTimer`
     * threshold.
```

src/countdown-timer/countdown-timer.tsx
```diff
@@ -16,14 +16,20 @@ import * as styles from "./countdown-timer.styles";
...
+ */
 export const CountdownTimer = ({
     className,
     align = "right",
     timer,
     timestamp,
     notifyTimer,
     offset,
-    mobileOffset,
+    mobileOffset = { top: 80 },
```
qroll: not needed? the fallback is done in `getOffsetY`

src/countdown-timer/types.ts
```diff
@@ -12,39 +12,83 @@ interface MobileOffset {
...
+     * Remaining-seconds threshold at which the warning state activates.
+     * Once the countdown reaches this value, `onNotify` fires once and
+     * `onTick` fires every second until the timer finishes.
+     */
     notifyTimer?: number | undefined;
-    /** Allows customization of the sticky position in tablet/desktop view */
+    /**
+     * Fine-tunes the fixed overlay position on tablet and desktop viewports.
+     *
+     * @remarks `top` defaults to 168 when not provided; `left`/`right`
```
qroll, line 37: since fallbacks are mentioned here, coudl we use `@default` directly?

src/dashed-border/dashed-border.tsx
```diff
@@ -98,4 +98,11 @@ const Component = (
     );
 };
 
+/**
+ * A container that renders a dashed border using an SVG overlay.
+ *
+ * Use when you need a dashed visual boundary around content — for example, drop zones,
+ * placeholder regions, or empty-state containers. The border is drawn as an `aria-hidden`
+ * SVG positioned overlay so it does not participate in the accessibility tree.
```
qroll: this line is an implementation detail, we could omit this

src/dashed-border/types.ts
```diff
@@ -8,16 +8,38 @@ import type { ValueOf } from "../util/utility-types";
...
 
+/** Props for the `DashedBorder` component. */
 export interface DashedBorderProps
     extends Omit<React.HTMLAttributes<HTMLDivElement>, "color" | "style"> {
-    /** Optional background fill behind the dashed overlay. */
+    /**
+     * Optional background fill applied inside the container, behind the dashed overlay.
+     *
+     * @remarks The SVG overlay is rendered even when `enabled` is `false` if this prop
+     * is provided, so the background colour is still visible.
```
qroll:
```suggestion
     * @remarks If this prop is provided, the background colour remains visible even when `enabled` is `false`
```

src/data-table/types.ts
```diff
@@ -2,44 +2,74 @@ import type { CSSProperties, ReactNode } from "react";
...
     headers: HeaderProps[];
+    /** Row data to render. When omitted or empty the empty state is shown. */
     rows?: RowProps[] | undefined;
     className?: string | undefined;
+    /**
+     * Controlled set of row IDs that are currently selected.
+     *
+     * @remarks Must be provided alongside `enableMultiSelect` for selection to work. Pass an
+     * empty array (not `undefined`) to enable the select-all checkbox while having no rows
+     * selected.
```
qroll, line 19:
```
Pass an empty array (not `undefined`) to enable the select-all checkbox while having no rows selected.
```

hmm do we need to specifically call this out? if this behaviour isn't intuitive, then the implementation could be tweaked instead

src/date-input/date-input.tsx
```diff
@@ -16,6 +16,13 @@ import { DateInputHelper } from "../util";
 import * as styles from "./date-input.styles";
 import type { DateInputProps } from "./types";
 
+/**
+ * A single-date input field with an attached calendar dropdown.
+ *
+ * Use this to enter or pick a single date. The field
+ * accepts keyboard entry via segmented day/month/year inputs and opens a
+ * calendar dropdown on focus.
```
qroll: could be omitted

src/date-navigator/types.ts
```diff
@@ -1,28 +1,60 @@
...
     onCalendarDateSelect?: ((currentDate: string) => void) | undefined;
     /**
-     * The root element that contains the dropdown element. Defaults to the document body.
+     * The root element that contains the dropdown element.
      *
-     * If the parent that contains the trigger element has a higher z-index than the dropdown,
-     * the dropdown may not be visible. Specify the parent element here instead
+     * @remarks
+     * Defaults to the document body. If a parent of the trigger has a higher
+     * `z-index` than the dropdown, the dropdown may be obscured.
```
qroll: I think the storybook props are worded closer to the intended use of this prop.

proposing this:

```
The root element that contains the dropdown element.

@remarks Specify this if you need to change the parent of the dropdown in the HTML DOM. Possible use case: sharing a stacking context. Note: if a parent of the trigger has a higher `z-index` than the dropdown, the dropdown may be obscured.

@default `document.body`
```

qroll (changes requested):

src/date-input/types.ts
qroll, line 25: to standardise with `date-navigator`

src/countdown-timer/countdown-timer.tsx
ghazwan-gt: Oops my bad, thinking of removing the one in the `getOffsetY` as we only reference it there

src/countdown-timer/types.ts
ghazwan-gt: Hmm as both `left` and `right` are dynamic values calculated at runtime, I prefer to use `@remarks`. Or do you think it would be better to not mention of the fallback of the property?

src/data-table/types.ts
ghazwan-gt: I'd argue that we should mention this since I've come across it during checking which is confusing at first. For now, we can keep the comment and have this as part of future enhancement

src/date-input/types.ts
ghazwan-gt: Let me handle this in a separate PR

src/countdown-timer/types.ts
qroll: as discussed, since `offset` has more complex fallback behaviour, can leave in `@remarks`

src/data-table/types.ts
qroll: created a ticket to update the implementation

src/countdown-timer/countdown-timer.tsx
qroll: would prefer to leave it in `getOffsetY` for now so that the fallback is referenced in the same place as desktop

qroll (approved):

PR #1347, opened 2026-07-17
[CCUBE-2183][GZ] automated storybook component api documentation part 1

ghazwan-gt:
## Type of changes

## Description of changes

[Autogenerate Storybook ArgsTable](https://sgtechstack.atlassian.net/browse/CCUBE-2183)
### Automated component API
Cover these stories:
- Core
- Content
- Navigation
  
####  Commands
```
npm run storybook:argtypes          # Generate all component argTypes files
npm run storybook:argtypes -- --watch  # Watch mode — regenerate on file changes
```

#### Generation mode
default:
  - Scans all `src/**/types.ts` files
  - Extracts interfaces, type aliases, and `Component<>` variable statements
  - Generates per-component `.storybook/generated/[name].argtypes.generated.ts` files
  - Generates a global registry `.storybook/generated/storybook-argtypes.generated.ts` mapping story titles to their `argTypes`

Watch mode (--watch):
  - Uses `chokidar` to watch src/ and stories/ directories
  - On `types.ts` change, regenerates that compone
  - On `types.ts` deletion, removes generated file + updates registry

####  Type extraction features
  - `@storybookSection` assigns props to named tabs in the Storybook UI
  - `@storybookSkipProps` skips a specific interface/type from generation
  - `@storybookSkipFile` skips an entire `types.ts` file
  - JSDoc tags: `@deprecated`, `@default`, `@remarks`, `@example`
  - Expands literal unions (e.g. `Exclude<FontSize, "form-label">` to become individual values)
  - Resolves inherited HTML attributes into a descriptive row
  - Handles generic interfaces
  - Multi-tab duplication for interfaces with multiple `@storybookSection` entries, used by shared component types

### Storybook
- New storybook component, `AutoArgTypesTabs`,  to render `argTypes` content from the global generated registry
- Enable component API table to inherit the theme applied by the Storybook decorator. Dark mode now affect the component API also
- `Tabs` storybook component renders the tab content with full width
- Simplified the webpack progress log to have cleaner log when running storybook

##Checklist##

-   [X] Updated documentation


.storybook/decorators/theme-docs-container.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const resolvedThemeType =
+        THEME_KEY_TO_TYPE_MAPPING[resolvedThemeName] ?? "lifesg";
+    const mode = isDark ? "dark" : "light";
+
+    useEffect(() => {
+        const handleDarkMode = (value: boolean) => {
+            setIsDark(value);
+        };
+
+        context.channel.on("DARK_MODE", handleDarkMode);
```
qroll: can import `DARK_MODE_EVENT_NAME` from `@storybook-community/storybook-dark-mode`

scripts/generate-storybook-argtypes.ts
```diff
@@ -0,0 +1,1180 @@
...
+        ignoreInitial: true,
+        persistent: true,
+        ignored: (filePath, stats) => {
+            if (!stats?.isFile()) {
+                return false;
+            }
+
+            if (filePath.endsWith(".argtypes.generated.ts")) {
+                return true;
+            }
```
qroll: the generated files sit in the `.storybook` folder which does not fall dunder the watchRoots, is this needed?

scripts/generate-storybook-argtypes.ts
```diff
@@ -0,0 +1,1180 @@
...
+                    name: "",
+                    tabGroup: jsDocMeta.tabGroup,
+                    typeSummary: "",
+                }),
+            ];
+        });
+}
+
+/** Generate argTypes for a single component's types.ts file. */
+async function generateForSourceFile(project: Project, sourceFilePath: string) {
```
qroll, line 904: we should extract `generateForSourceFile` and `generateStorybookArgTypesRegistry` to separate files eventually, but this refactoring can be done at the end

src/link-list/types.ts
```diff
@@ -1,11 +1,104 @@
...
+    /**
+     * Additional content rendered below the description.
+     * Accepts a JSX element for rich secondary content.
+     */
+    secondaryDescription?: JSX.Element | undefined;
+    "data-testid"?: string | undefined;
+    /** Arbitrary custom data passed through to the `onItemClick` callback. */
+    options?: T | undefined;
+    /** Called when an item is clicked. This overrides the `onItemClick` event handler. */
+    onClick: (event: React.MouseEvent<HTMLAnchorElement>) => void;
```
qroll: the item-level `onClick` is optional for consumers

src/typography/types.ts
```diff
@@ -1,3 +1,5 @@
+// @storybookSkipFile
```
qroll: why was typography skipped?

qroll (commented):
- for default values
```
@default `...`
```

let's remove the backtick, so it'll just be
```
@default Colour["border"]
@default document.body
@default 0
```

- inert should be excluded as it's part of HTML attributes
- expanding union types (to confirm on the behaviour for internal vs cross module)
- style the section headers to retain the original casing
- empty string `""` for ErrorDisplay `additionalProps`

Image

- Remove the "This component also inherits..." remark from `card.mdx`
- DataTable, Footer are not migrated
- FullscreenImageCarousel `items` is replaced with `FullscreenImageCarouselItemProps` and also receiving an empty string `""`
- Inheritance info declared twice for Table

Screenshot 2026-07-20 at 12 22 47 PM

src/table/types.ts
```diff
@@ -0,0 +1,46 @@
+// @storybookSection Table.Head
+/**
+ * Inherits props from [HTMLTableSectionElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLTableSectionElement)
+ */
+export interface HeadProps
```
qroll: we should uniquely prefix the types to avoid naming collisions with other component

```suggestion
export interface TableHeadProps
```

src/link-list/types.ts
```diff
@@ -1,11 +1,104 @@
...
+}
+
+// @storybookSkipProps
+/**
+ * Base props for `LinkList`.
+ */
+export type BaseProps<T> = {
```
qroll: should not be exported

qroll: or rename to `LinkListBaseProps`

src/navbar/types.ts
qroll, line 162: to be skipped, as it's extended by `NavbarProps`

src/modal-v2/types.ts
```diff
@@ -5,6 +5,8 @@ export type ModalAnimationDirection = "top" | "bottom" | "left" | "right";
 
 /**
  * Props for the `ModalV2.Card` sub-component.
+ *
+ * @storybookSection Sub Components
```
qroll: this doesn't seem to be updated yet in SB, but I assume it's meant to reflect the individual sub components eventually?

src/modal/types.ts
qroll, line 1: same for this file

src/uneditable-section/types.ts
```diff
@@ -96,17 +102,17 @@ export interface UneditableSectionProps {
      *
      * @param item The item whose mask control was activated.
      */
-    onMask?: ((item: UneditableSectionItemProps) => void) | undefined;
+    onMask?: (item: UneditableSectionItemProps) => void | undefined;
```
qroll: the surrounding `()` are needed, as otherwise Typescript interprets as a function that returns `void | undefined`

stories/storybook-common/arg-types/table.tsx
```diff
@@ -0,0 +1,233 @@
...
+ * directly (via `PureArgsTable`) so each tab reliably renders only its own
+ * tab-group entries.
+ */
+export const AutoArgTypesTabs = ({
+    of,
+    storyTitle,
+    sort = "alpha",
+    defaultTabTitle = "Component",
+}: AutoArgTypesTabsProps): JSX.Element => {
+    const resolvedStoryTitle = storyTitle ?? getStoryTitleFromOf(of);
```
qroll: how is `storyTitle` intended to be used? as I noticed all usages have been passing `of` so far

scripts/generate-storybook-argtypes.ts
ghazwan-gt: Nice catch, we can remove this one

src/typography/types.ts
ghazwan-gt: It was used during experimenting skip tag, would include it

src/modal-v2/types.ts
ghazwan-gt: Yes, this is being included by accident. Was meant to include the change to provide default value of `show` props

stories/storybook-common/arg-types/table.tsx
ghazwan-gt: This should be deleted. Was part of the original code to utilize `include` and `exlude` props

scripts/generate-storybook-argtypes.ts
```diff
@@ -0,0 +1,1180 @@
...
+    return sourceFile
+        .getVariableStatements()
+        .filter((statement) => statement.isExported())
+        .filter((statement) => {
+            const declaration = statement.getDeclarations()[0];
+
+            return declaration
+                ?.getInitializer()
+                ?.getText()
+                .includes("Component<");
```
qroll: is this a placeholder? what `VariableStatement`s are in the types? what is using `Component<` at the moment?

scripts/generate-storybook-argtypes.ts
```diff
@@ -0,0 +1,1180 @@
...
+        ...getInheritedComponentArgTypes(sourceFile),
+    ];
+
+    const sortedRows = rows.toSorted((a, b) => a.key.localeCompare(b.key));
+
+    const generatedArgTypes = Object.fromEntries(
+        sortedRows.map((row) => [row.key, row.value])
+    );
+
+    const generated = `// This file is generated. Do not edit manually.
```
qroll, line 937: run prettier on this to prevent whitespace/formatting differences before the linter runs

scripts/generate-storybook-argtypes.ts
```diff
@@ -0,0 +1,1180 @@
...
+        ...sourceFile
+            .getTypeAliases()
+            .map((declaration) => declaration.getName()),
+    ];
+
+    const rows = [
+        ...typeSections.flatMap((typeName) => {
+            const interfaceDeclaration = sourceFile.getInterface(typeName);
+
+            if (interfaceDeclaration) {
```
qroll: we are merging `typeSections`, then processing them separately with `getInterfaceArgTypes` and `getTypeAliasArgTypes`. why not separate them from the start?

scripts/generate-storybook-argtypes.ts
```diff
@@ -0,0 +1,1180 @@
...
+
+    if (!docs.length) {
+        return {};
+    }
+
+    const description =
+        docs
+            .map((doc) => doc.getCommentText()?.trim())
+            .filter((comment): comment is string => Boolean(comment))
+            .join("\n\n") || undefined;
```
qroll, line 168: why do we need to trim the node?

scripts/generate-storybook-argtypes.ts
```diff
@@ -0,0 +1,1180 @@
...
+
+function getInterfaceArgTypes(
+    sourceFile: SourceFile,
+    interfaceName: string
+): GeneratedArgType[] {
+    const interfaceDeclaration = sourceFile.getInterfaceOrThrow(interfaceName);
+    const interfaceJsDocMeta = getJsDocMeta(interfaceDeclaration);
+    const declaredTabGroups = interfaceJsDocMeta.tabGroups ?? [
+        interfaceJsDocMeta.tabGroup,
+    ];
```
qroll: wondering why `tabGroups` and `tabGroup` have to be distinguished if we are normalising to an array again

scripts/generate-storybook-argtypes.ts
```diff
@@ -0,0 +1,1180 @@
...
+        if (/@?storybookSkipProps\b/.test(comment)) {
+            return true;
+        }
+    }
+
+    return false;
+}
+
+/**
+ * Collect JSDoc metadata from a node. Supported tags: @deprecated, @default, @remarks, @example.
```
qroll: nit but this function is showing up as deprecated due to the presence of the deprecated tag, but wrapping the tags in `"` seems to prevent them from applying

src/table/types.ts
qroll, line 1: `table.tsx` has typescript errors as it is still referencing the previous interface name

src/link-list/types.ts
qroll, line 1: typescript errors in link list as it is trying to export `BaseProps`

scripts/generate-storybook-argtypes.ts
ghazwan-gt: This was used during the interface expanding to get the inherited argtypes. We can remove this as it is not used

scripts/generate-storybook-argtypes.ts
ghazwan-gt: Nice catch! That was unnecessary, will remove it

scripts/generate-storybook-argtypes.ts
ghazwan-gt: There's unnecessary leading space at the front

scripts/generate-storybook-argtypes.ts
ghazwan-gt: Right, the `tabGroup` was the initial design, but since we need to provide a type be part of multiple tabs (shared), the `tabGroups` was introduced. And I forgot to cleanup this up

scripts/generate-storybook-argtypes.ts
qroll: re: prettier, would prefer it to run during the generation itself, so that formatting diffs do not show. else it may result in an empty commit after

src/link-list/types.ts
qroll: back to this, to rename to `LinkListBaseProps` if not moving to an internal types file https://github.com/LifeSG/react-design-system/pull/1347#discussion_r3612046240

qroll (approved):

PR #1357, opened 2026-07-23
[CCUBE-2183][GZ] automated storybook component api documentation part 6

ghazwan-gt:
## Type of changes

## Description of changes

[Autogenerate Storybook ArgsTable](https://sgtechstack.atlassian.net/browse/CCUBE-2183)

- Refactor the script to exclude the duplicated props from interface which extends through `Omit` or `Pick`
- Generate argTypes for the rest of Overlays stories
  - Renamed `PopoverInline` dir to `Inline` to support the current script. Lmk if we want to keep it with the same name 
 

## Checklist

-   [X] Updated documentation


scripts/generate-storybook-argtypes.ts
```diff
@@ -50,7 +50,11 @@ type StorybookTaggedDeclarationNode =
     | TypeAliasDeclaration
     | VariableStatement;
 
-const sourceFileGlobs = ["src/*/types.ts", "src/*/addons/types.ts"];
+const sourceFileGlobs = [
+    "src/*/types.ts",
+    "src/filter/addons/types.ts",
+    "src/popover/inline/types.ts",
```
qroll, line 56: should we just look at `types.ts` files at all levels? would those unused in storybook (e.g. `src/shared`) be ignored?

scripts/generate-storybook-argtypes.ts
```diff
@@ -1304,7 +1308,7 @@ function shouldSkipImportedTypeRows(
         const extendsImportedType = interfaceDeclaration
             .getExtends()
             .some((extendNode) => {
-                const extendText = extendNode.getExpression().getText();
+                const extendText = extendNode.getText();
```
qroll, line 1311: what's the difference here?

src/menu/types.ts
qroll, line 1: missed it out earlier, but for menu, the default `position` is "bottom-start" instead

stories/modal-v2/modal-v2.mdx
```diff
@@ -102,4 +102,4 @@ scrollable.
 All components also inherit the
```
qroll: can remove this line now

stories/modal/modal.mdx
```diff
@@ -65,7 +65,7 @@ You're recommended to apply this spacing around your content to avoid overlap wi
 
 Both components also inherit the [HTMLDivElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDivElement).
```
qroll: can remove this line

scripts/generate-storybook-argtypes.ts
ghazwan-gt: We have this at start, but turned out with this we would also generate files for those non-storybook related files. Well we can do this, but this means polluting the `.storybook/generated` dir with unrelated files

scripts/generate-storybook-argtypes.ts
ghazwan-gt: This one is to handle `extends` & `implements` with `Pick` or `Omit` utility. With `getExpression`, it will only return `Omit` without the type inside the bracket.

Lets take this for example
```
export interface MenuProps extends Omit<PopoverTriggerProps, "popoverContent"> {
    /**
     * The `Menu.Content` element to display inside the popover panel.
     */
    menuContent: FunctionComponentElement<MenuContentProps>;
}
```
With `getExpression`, we won't be able to recognize that `MenuProps` is extending `PopoverTriggerProps`, which resulted with us having a new category listing the `PopoverTriggerProps` props. This not the result we expect as we already expand all of it to be part of `MenuProps`

qroll (approved):

scripts/generate-storybook-argtypes.ts
qroll: more of a good to have, but off the top of my head, we could
1. enforce a `types.ts` file to have a `@storybookX` tag, could reuse `@storybookSection` even (with no tab)
2. blacklist instead of whitelist, like in our rollup config with `dirsToIgnore`

just think it's not ideal to have to remember to explicitly include the path in the script

PR #1414, opened 2026-08-12
[CCUBE-2258][RYN] refactor Form.Textarea and Textarea error display

ryan-nguyen-t:
**Type of changes**

-   [x] Breaking change (fix or feature that would cause existing apis or functionality to change)

**Description of changes**

-   Link to [ticket](https://sgtechstack.atlassian.net/browse/CCUBE-2258)
- Refactor `Textarea` to own error message + character counter rendering; `Form.Textarea` becomes a stateless passthrough wrapper (no state, effects, or handlers) mirroring the `Form.Input` pattern.

**Checklist**

-   [ ] Updated documentation


src/input-textarea/textarea.tsx
```diff
@@ -1,6 +1,7 @@
 import clsx from "clsx";
 import React, { useEffect, useState } from "react";
 
+import { FormErrorMessage } from "../form/form-label";
```
ziggyzet, line 4: hmmm don't think it's wise to reference error message label from the base textarea here. might be better to just make an error message specific to text area, but wdyt?

ghazwan-gt: Agree with @ziggyzet here. But since there are other places with this implementation, I'd suggest to have a ticket to handle this. I'm thinking of having a shared `ErrorLabel` component instead of importing from `Form` as the component could be independent. What do you guys think?

src/input-textarea/textarea.tsx
```diff
@@ -155,6 +156,8 @@ const TextareaComponent = (
         prefix,
         maxLength,
         renderCustomCounter,
+        errorMessage,
```
ghazwan-gt, line 159: As we have a new props here, lets add this case to the e2e. Can include it in the standalone stories

src/input-textarea/textarea.styles.ts
```diff
@@ -79,3 +79,19 @@ export const elementError = css`
...
+export const labelContainer = css`
+    display: flex;
+    justify-content: flex-end;
+`;
+
+export const errorMessageContainer = css`
+    display: flex;
+    flex: 1;
+    margin-right: ${Spacing["spacing-12"]};
+    gap: ${Spacing["spacing-4"]};
```
ghazwan-gt: From what I see, `gap` is not used since we only have a sole children, and we can move `margin-right` to the parent by using `gap` instead

src/input-textarea/textarea.tsx
```diff
@@ -196,15 +224,10 @@ const TextareaComponent = (
                 prefix={prefix}
                 transformValue={transformValue}
                 maxLength={maxLength}
+                error={!!errorMessage || otherProps.error}
                 {...otherProps}
```
ghazwan-gt: With this, `error` will be overwritten by the one from `...otherProps`, ignoring `errorMessage`. Lets just destructure the `error` at the top

e2e/nextjs-app/src/app/components/input-textarea/standalone-error-message-mobile.e2e.tsx
ghazwan-gt, line 1: We can have the long text to be part of `standalone-error-message` story instead

e2e/tests/components/input-textarea/input-textarea.e2e.spec.ts
```diff
@@ -13,6 +13,8 @@ class StoryPage extends AbstractStoryPage {
         disabledTextarea: Locator;
         readonlyTextarea: Locator;
         errorTextarea: Locator;
+        errorMessageTextarea: Locator;
+        errorMessageCounterTextarea: Locator;
```
ghazwan-gt: These two are not being used

src/input-textarea/textarea.styles.ts
ghazwan-gt: It seems some of the stylings are removed by mistake, making the e2e fails

ghazwan-gt (dismissed):

e2e/tests/components/input-textarea/__screenshots__/chromium/InputTextarea-Standalone-errorMessage-mobile---mount.png
qroll, line 1: behaviour of text overflow is different from production, expect the icon to be aligned to the top

Image

e2e/tests/components/input-textarea/__screenshots__/chromium/InputTextarea-Standalone-errorMessage--mount.png
qroll, line 1: since desktop can't produce the overflow behaviour, how about we merge the tests and just use mobile viewport by default?

qroll (changes requested):

src/input-textarea/textarea.tsx
qroll: as discussed, to reuse `FormErrorMessage` for now

in V5, the plan is to merge the label and error message into the individual components, so the Form module is no longer relevant.

we can introduce these standalone components:
- ErrorLabel
- Label
- FormControl

and mark the Form module as deprecated

e2e/tests/components/input-textarea/input-textarea.e2e.spec.ts
```diff
@@ -250,4 +250,22 @@ test.describe("InputTextarea", () => {
...
+            await compareScreenshot(story, "mount");
+        });
+
+        test("Standalone errorMessage (dark mode)", async ({ story }) => {
+            await story.init("standalone-error-message", {
+                size: "mobile",
+                mode: "dark",
+            });
+            await compareScreenshot(story, "mount");
+        });
```
qroll: can remove, as dark mode behaviour is mostly captured by the other tests

(and we shouldn't nest story.init)

qroll (changes requested):

qroll (approved):

PR #1426, opened 2026-08-17
[CCUBE-2188][GZ] refactor file upload

ghazwan-gt:
**Type of changes**

**Description of changes**
-   Link to [ticket](https://sgtechstack.atlassian.net/browse/CCUBE-2188)
-   Link to [Figma](https://www.figma.com/design/tHmHQNszVzvIamtZWixjgR/%F0%9F%8C%B1-Flagship-V3-Component-Kit?node-id=2162-23496&p=f&m=dev)
-  Merged `FileListItem` (display/error) and `FileItemEdit` (edit) into a single unified component that manages its own mode internally
  - Reduced code complexity by extracting shared logic into reusable pieces
  - Simplified the parent `FileList` by removing mode-tracking state and replacing it with a simple predicate
  - Consolidated two separate style files into one

**Checklist**

-   [X] Looks good on mobile and tablet
-   [ ] Updated documentation
-   [X] Added/updated E2E tests

**Screenshots**

| Before | After |
|--------|--------|
| image | image |


src/file-upload/file-list/file-list.tsx
```diff
@@ -452,16 +405,19 @@ function Component(
             }
 
             return (
-                <FileItemEdit
+                <FileListItem
```
qroll, line 364: seeing that the props for edit mode and non-edit mode FileListItem are mostly identical, could the rendering logic be merged to a `renderItem`?

src/file-upload/file-list/file-list.tsx
```diff
@@ -314,62 +298,28 @@ function Component(
...
     /**
      * Due to a UI requirement, we will render the items
      * with edit modes as a group
      */
-    const getArrangedItems = (
-        fileItems: FileItemProps[] | undefined,
-        renderModes: FileItemRenderModes
-    ) => {
-        if (!fileItems || fileItems.length === 0) return [];
+    const getArrangedItems = (fileItems: FileItemProps[]) => {
```
qroll: on further consideration, this grouping logic is no longer needed. previously it was used to collect editable items that were siblings to be passed to `file-item-edit` for rendering the borders

now that we don't have the separate layout for editable items, we can flatten the structure and hopefully simplify the iteration logic

src/file-upload/file-list-item/file-list-item.styles.ts
```diff
@@ -10,6 +10,102 @@ import {
...
+        margin-right: 0;
+        margin-bottom: ${Spacing["spacing-8"]};
+    }
+`;
+
+export const editFileSizeText = css`
+    color: ${Colour["text-subtler"]};
+`;
+
+export const actionButtonsSection = css`
```
qroll: propose to move the action button styles to a standalone `file-item-actions.styles.ts` if they are not used by other elements

src/file-upload/file-list-item/file-item-details.tsx
```diff
@@ -0,0 +1,114 @@
...
+}: Props) => {
+    if (mode === "error") {
+        return (
+            <div ref={detailSectionRef} className={styles.nameSection}>
+                <Typography.BodyMD
+                    weight={description ? "semibold" : "regular"}
+                >
+                    {formattedName}
+                </Typography.BodyMD>
+                {description && (
```
qroll: does error need to cater for description?

src/file-upload/file-list-item/file-item-details.tsx
```diff
@@ -0,0 +1,114 @@
...
+            </div>
+        );
+    }
+
+    return (
+        <div ref={detailSectionRef} className={styles.nameSection}>
+            <Typography.BodyMD weight={description ? "semibold" : "regular"}>
+                {formattedName}
+            </Typography.BodyMD>
+            {description && (
```
qroll, line 63: just to check, in what scenario does this fall under where `description` is still defined?

src/file-upload/file-list-item/file-list-item.tsx
```diff
@@ -120,345 +146,213 @@ const Component = ({
...
-            </div>
+    const thumbnail = shouldShowThumbnail ? (
+        <FileListItemThumbnail
+            thumbnailImageDataUrl={thumbnailImageDataUrl || ""}
+            fileType={fileItem.type}
+            data-testid={`${id}-thumbnail`}
+        />
+    ) : null;
+
+    const renderInlineActions = () =>
```
qroll: any particular reason for having a render function rather than inlining for this mode?

src/file-upload/file-list-item/file-list-item.tsx
```diff
@@ -120,345 +146,213 @@ const Component = ({
...
-        }
+        setCurrentMode("edit");
+        onModeChange?.("edit");
     };
 
     const handleKeyDown = (event: React.KeyboardEvent<HTMLButtonElement>) => {
-        /**
-         * Circumvent issue of keydown action activating the sort mechanism
-         * rather than the actual action
-         */
```
qroll, line 132: restore this comment

src/file-upload/file-list-item/file-list-item.tsx
qroll, line 1: finding the split of `renderModes` into `editingIds` on the parent file list and `currentMode` on the child to be awkward, but will go with this

e2e/tests/components/file-upload/__screenshots__/chromium/FileUpload-Long-description-text--mount.png
qroll, line 1: was thinking desktop is not needed since the text doesn't wrap in this viewport, and mobile has the same layout anyway

src/file-upload/file-list/file-list.tsx
```diff
@@ -314,77 +270,18 @@ function Component(
...
-    const shouldRenderEditMode = (item: FileItemProps) => {
+    const isInNonDisplayState = (item: FileItemProps) => {
+        if (item.errorMessage) return true;
         return (
-            !item.errorMessage &&
             !readOnly &&
-            !(typeof item.progress === "number" && item.progress < 1) &&
             checkEditable(item) &&
-            !item.description
+            !item.description &&
```
qroll: I think this needs to check `descriptionRequired && !item.description`, plus the whether the item itself is in edit mode

src/file-upload/file-list-item/file-list-item.tsx
qroll: did `editingIds` get removed? that's still needed to disable the sorting when at least 1 item is in edit mode

src/file-upload/file-list-item/file-list-item.tsx
```diff
@@ -135,330 +152,196 @@ const Component = ({
...
-
-        return undefined;
+    const handleTextareaBlur = (
+        event: React.FocusEvent<HTMLTextAreaElement>
+    ) => {
+        onBlur?.(event.target.value);
     };
 
+    const disableSave =
+        descriptionRequired && currentDescription.trim().length === 0;
```
qroll: nit: move below the section header

src/file-upload/file-list-item/file-list-item.tsx
```diff
@@ -135,330 +152,196 @@ const Component = ({
...
             >
-                {renderContents()}
-                {!readOnly && !hasInlineActions && renderActions()}
+                <div
+                    className={styles.contentSection}
+                    data-has-thumbnail={
+                        currentMode === "display" && shouldShowThumbnail
+                    }
+                >
+                    {currentMode === "display" && thumbnail}
```
qroll, line 311: the thumbnail needs to be centered vertically even when actions are not inlined

Image

qroll (changes requested):

src/file-upload/file-list/file-list.tsx
```diff
@@ -401,24 +411,20 @@ function Component(
...
 
-    if (disabled || readOnly || !shouldEnableSort()) {
-        return renderItemsWithWrapper();
-    } else {
-        return (
-            <DndContext
-                sensors={sensors}
-                onDragEnd={handleDragEnd}
-                onDragStart={handleDragStart}
+    return (
```
qroll, line 418: I think the original intention for separate rendering was to avoid initialising the dnd context and its listeners when it's not needed

src/file-upload/file-list/file-list.tsx
```diff
@@ -223,6 +225,12 @@ function Component(
         }
     };
 
+    const handleModeChange = (mode: FileItemMode) => {
+        setEditingCount((prev) =>
+            mode === "edit" ? prev + 1 : Math.max(0, prev - 1)
+        );
```
qroll, line 231: this might not work if the `fileItems` prop updates separately; the count won't be removed?

src/file-upload/file-list-item/file-list-item.tsx
qroll: can check why this happens?

Screenshot 2026-09-04 at 2 46 15 PM

1. upload image
2. upload non-image
3. result: sort indicator is visible on the non-image
4. upload another file
5. result: sort indicator is not visible

for step 3, the sort indicator should not be visible either

src/file-upload/file-list/file-list.tsx
qroll: got it, since the `key` is set to the file id, the file item will unmount and clear the count

src/file-upload/file-list/file-list.tsx
qroll: oh but doesn't seem to affect the functionality, resolving

src/file-upload/file-list-item/file-list-item.tsx
ghazwan-gt: Ooh nice finding, didn't think of this one before. Will take a look at this

e2e/tests/components/file-upload/file-upload.e2e.spec.ts
qroll, line 1: could this be part of the unit tests instead?

src/file-upload/file-list/file-list.tsx
```diff
@@ -280,15 +280,11 @@ function Component(
 
     const isInNonDisplayState = (item: FileItemProps) => {
         if (item.errorMessage) return true;
-
-        const isUploaded = (item.progress ?? 1) >= 1;
-        const isDescriptionRequiredAndEmpty =
-            descriptionRequired && !item.description;
```
qroll, line 286: hmm I don't understand the change here

the original `isDescriptionRequiredAndEmpty` is still needed for the case where `descriptionRequired={false}`

src/file-upload/file-list/file-list.tsx
```diff
@@ -280,15 +280,11 @@ function Component(
...
         return (
             !readOnly &&
```
qroll, line 284: I realise readOnly is a shared state and can be moved to `shouldEnableSort`

src/file-upload/file-list/file-list.tsx
qroll, line 292: back to this, I think the initial editingCount is incorrect. will need to fire `"edit"` on mount when the image file is uploaded

e2e/tests/components/file-upload/file-upload.e2e.spec.ts
ghazwan-gt: Moved to unit test

src/file-upload/file-list/file-list.tsx
ghazwan-gt: My bad, fixed it

src/file-upload/file-list/file-list.tsx
ghazwan-gt: Refactored, now `readOnly` is now checked in `shouldEnableSort` instead of per item

src/file-upload/file-list/file-list.tsx
ghazwan-gt: Updated, the child now fires `onModeChange("edit")` on mount when descriptionRequired is true and `onModeChange("display")` on unmount cleanup

src/file-upload/file-list-item/file-list-item.tsx
```diff
@@ -54,75 +40,136 @@ const Component = ({
...
     // =========================================================================
     // EFFECTS
     // =========================================================================
+    const currentModeRef = useRef(currentMode);
+    currentModeRef.current = currentMode;
+    const onModeChangeRef = useRef(onModeChange);
+    onModeChangeRef.current = onModeChange;
+
+    useEffect(() => {
+        if (currentMode === "edit" && descriptionRequired) {
```
qroll: should this just be currentMode check? because for case where description is optional, it still mounts with the textarea open by default

specifically this refutes the `should not disable sort for auto-edit items when descriptionRequired is false` test

```suggestion
        if (currentMode === "edit") {
```

qroll (approved):

PR #1428, opened 2026-08-18
[CCUBE-2221][ZZ] feat: migrate ESignature to Modal v2

ziggyzet:
**Type of changes**

-   [x] New feature (non-breaking change which adds functionality)
-   [x] Tests (improvements to unit tests or E2E tests)

**Description of changes**

-   Link to [ticket](https://sgtechstack.atlassian.net/browse/CCUBE-2221)
-   Migrating Esignature to use modal v2. There are some screenshot updates after these changes.

**Checklist**

-   [ ] Updated documentation
-   [x] Added/updated E2E tests


src/e-signature/e-signature.styles.ts
```diff
@@ -71,56 +75,117 @@ export const instructions = css`
...
-    ${MediaQuery.MaxWidth.sm}, &[data-mobile-landscape="true"] {
+    & > [data-has-close-button] {
+        flex: 1;
         display: flex;
         flex-direction: column;
-        max-width: none;
-        height: 100%;
-        border-radius: 0;
-        padding: 0;
+        min-height: 0;
```
ghazwan-gt, line 93: Lets remove this one

src/e-signature/e-signature.styles.ts
```diff
@@ -180,30 +249,39 @@ export const signatureLine = css`
...
 
 export const modalButtons = css`
-    display: flex;
-    justify-content: space-between;
-    margin-top: ${Spacing["spacing-16"]};
+    && {
+        flex-direction: row-reverse;
+        margin: ${Spacing["spacing-16"]} 0 0;
+        margin-right: 0;
+        margin-left: 0;
```
ghazwan-gt: Lets remove these

src/e-signature/e-signature.styles.ts
```diff
@@ -71,56 +75,117 @@ export const instructions = css`
...
     }
-`;
 
-export const growContainer = css`
-    width: 100%;
-    margin: auto;
-    padding: ${Spacing["layout-xxl"]} ${Spacing["layout-sm"]};
+    ${MediaQuery.MaxWidth.sm} {
+        &[data-mobile-landscape][data-mobile-landscape] {
+            max-width: none;
```
ghazwan-gt: We can keep the max-width as it is since we're using `MaxWidth.sm`

src/e-signature/e-signature.styles.ts
```diff
@@ -71,56 +75,117 @@ export const instructions = css`
...
+            height: 100vh;
+            margin: 0;
```
ghazwan-gt: We can also keep the margin as it is

ghazwan-gt (dismissed):

src/e-signature/e-signature.styles.ts
ziggyzet: this one specifically idt we can remove or else the content of the card will overflow

ghazwan-gt (dismissed):

tests/e-signature/e-signature.spec.tsx
```diff
@@ -85,20 +91,23 @@ describe("ESignature", () => {
...
         fireEvent.click(getAddSignatureButton());
         drawSignature();
         fireEvent.click(screen.getByRole("button", { name: "Save" }));
 
-        await waitFor(() => {
-            expect(getSignatureModal()).not.toBeVisible();
+        await act(async () => {
+            jest.advanceTimersByTime(500);
         });
+
```
qroll, line 105: wondering why the assertion on signature modal was removed, and we need to manually handle the timers here?

src/e-signature/e-signature.styles.ts
```diff
@@ -180,30 +247,37 @@ export const signatureLine = css`
...
 export const modalActionButton = css`
-    width: 8.5rem;
-
-    ${MediaQuery.MaxWidth.sm} {
-        width: 100%;
-    }
-
     &[data-mobile-landscape="true"] {
         height: 2.5rem;
     }
```
qroll: we can clean this up because the button default size is already 2.5rem

src/e-signature/e-signature.styles.ts
```diff
@@ -180,30 +247,37 @@ export const signatureLine = css`
 `;
 
 export const modalButtons = css`
-    display: flex;
-    justify-content: space-between;
-    margin-top: ${Spacing["spacing-16"]};
+    && {
+        flex-direction: row-reverse;
```
qroll: this style seems unnecessary

src/e-signature/e-signature.styles.ts
```diff
@@ -180,30 +247,37 @@ export const signatureLine = css`
 `;
 
 export const modalButtons = css`
```
qroll: thinking that if custom footer handling is required, we might as well just render a custom container as part of `ModalV2.Content`

src/e-signature/e-signature.styles.ts
```diff
@@ -180,30 +247,37 @@ export const signatureLine = css`
...
-    justify-content: space-between;
-    margin-top: ${Spacing["spacing-16"]};
+    && {
+        flex-direction: row-reverse;
+        margin: ${Spacing["spacing-16"]} 0 0;
+        justify-content: space-between;
+        column-gap: 0;
+    }
+
+    && > button {
```
qroll: should we reduce the specificity of the footer button styles?

src/e-signature/e-signature.styles.ts
```diff
@@ -180,30 +247,37 @@ export const signatureLine = css`
 `;
 
 export const modalButtons = css`
-    display: flex;
-    justify-content: space-between;
-    margin-top: ${Spacing["spacing-16"]};
+    && {
```
qroll: thinking that if custom footer handling is required, we might as well just render a custom container as part of ModalV2.Content

src/e-signature/e-signature.styles.ts
```diff
@@ -71,56 +75,115 @@ export const instructions = css`
...
+        box-shadow: ${Shadow["lg-strong"]};
     }
-`;
 
-export const growContainer = css`
-    width: 100%;
-    margin: auto;
-    padding: ${Spacing["layout-xxl"]} ${Spacing["layout-sm"]};
+    ${MediaQuery.MaxWidth.sm} {
+        &[data-mobile-landscape][data-mobile-landscape] {
```
qroll: why is the data-attribute specified twice?

src/e-signature/e-signature.styles.ts
```diff
@@ -71,56 +75,115 @@ export const instructions = css`
...
+            height: 100vh;
+            border-radius: 0;
+            padding: 0;
```
qroll: proposing to enhance `ModalV2.Card` with a `fullscreen` prop that disables the border-radius/height/width on mobile/mobile-landscape, could simplify the override behaviour for esignature

tests/e-signature/e-signature.spec.tsx
ziggyzet: the assertion was to check whether or not it's visible, but we're mount/unmounting for modal v2
modal v2 has a transition animation with 300ms, so that's just to skip the animation

src/e-signature/e-signature.styles.ts
ziggyzet: to increase the specificity because we're overriding what's inside the modal card

src/e-signature/e-signature.styles.ts
ziggyzet: okay noted, will change the implementation here!

tests/e-signature/e-signature.spec.tsx
qroll: why is it not needed for the other tests?

src/e-signature/e-signature.styles.ts
```diff
@@ -71,56 +75,95 @@ export const instructions = css`
...
     && {
-        height: 100%;
-        overflow-y: auto;
+        width: 100%;
+        height: 29rem;
+        max-width: 672px;
+        max-height: none;
+        margin: 0 auto;
+        padding: ${Spacing["spacing-16"]};
+        box-shadow: ${Shadow["lg-strong"]};
```
qroll, line 86: what does the box shadow do?

src/e-signature/e-signature.styles.ts
```diff
@@ -71,56 +75,95 @@ export const instructions = css`
 // SIGNATURE MODAL
 // -----------------------------------------------------------------------------
 
-export const scrollableModal = css`
-    /* increase specificity as the styles are overwritten */
+export const modalCard = css`
     && {
```
qroll, line 79: do we still need the increased specificity?

src/e-signature/e-signature.styles.ts
```diff
@@ -71,56 +75,95 @@ export const instructions = css`
...
+        width: 100%;
+        height: 29rem;
+        max-width: 672px;
+        max-height: none;
+        margin: 0 auto;
+        padding: ${Spacing["spacing-16"]};
+        box-shadow: ${Shadow["lg-strong"]};
+    }
+
+    & > [data-has-close-button] {
```
qroll, line 89: is there another way we can style the content without using this `data-has-close-button` selector? as that's an internal implementation for the modal. it also feels like this could be something that ModalV2 is supposed to handle properly with the fullscreen prop

src/e-signature/e-signature.styles.ts
```diff
@@ -71,56 +75,95 @@ export const instructions = css`
...
+
+        ${slotSpacerTokens.contentLastChildMarginBottom}: 0;
     }
 `;
 
-export const growContainer = css`
-    width: 100%;
-    margin: auto;
-    padding: ${Spacing["layout-xxl"]} ${Spacing["layout-sm"]};
+export const modalHeader = css`
```
qroll, line 99: ideally with the new `ModalV2.Header` we no longer need to specify so many custom styles. what are the changes here and why are they needed?

src/e-signature/e-signature.styles.ts
```diff
@@ -71,56 +75,95 @@ export const instructions = css`
...
     }
 `;
 
-export const modalTitle = css`
-    ${Font["body-baseline-semibold"]}
-    color: ${Colour["text"]};
-    margin-bottom: ${Spacing["spacing-16"]};
-    text-align: center;
+export const modalContent = css`
+    && {
```
qroll, line 152: ideally we should avoid using the `&&` selector

src/e-signature/e-signature.styles.ts
```diff
@@ -71,56 +75,95 @@ export const instructions = css`
...
+    }
+    display: flex;
+    flex-direction: column;
+    min-height: 0;
 
-    ${MediaQuery.MaxWidth.sm}, &[data-mobile-landscape="true"] {
-        ${Font["body-md-semibold"]}
-        margin: ${Spacing["spacing-12"]} 0;
+    ${MediaQuery.MaxWidth.sm} {
+        &[data-mobile-landscape] {
```
qroll, line 160: hmm why are these selectors nested?

qroll (changes requested):

src/e-signature/e-signature.styles.ts
ziggyzet: `Modal.Box` had this styling before but `ModalV2.Card` has xs-strong. Just making sure the styling is similar as the old one as much as possbile

src/e-signature/e-signature.styles.ts
ziggyzet: yes still needed here to override the `ModalV2.Card` styles

src/e-signature/e-signature.styles.ts
ziggyzet: to increase specificity as well

src/e-signature/e-signature.styles.ts
ziggyzet: will find a workaround then

src/e-signature/e-signature.styles.ts
ziggyzet: this was because the header fonts and spacing somehow after changing to using `ModalV2` is way different to what we were using when it's still `Modal`

src/e-signature/e-signature.styles.ts
ziggyzet: can find work arounds for this then. will comment back if no solution is found

src/e-signature/e-signature.styles.ts
```diff
@@ -71,56 +67,69 @@ export const instructions = css`
...
+    }
 
-        ${modalStyles.tokens.closeButton.topInset}: ${Spacing["spacing-8"]};
-        ${modalStyles.tokens.closeButton.rightInset}: ${Spacing["spacing-20"]};
+    & > div[aria-hidden] {
+        display: none;
+    }
+
+    ${MediaQuery.MaxWidth.sm} {
+        &[data-mobile-landscape] {
```
qroll, line 97: are the selectors supposed to be nested? this means the screen width is sm AND the screen height is sm

in the previous selectors, they were comma separated which should have a OR logic

src/modal-v2/types.ts
```diff
@@ -40,16 +46,18 @@ export interface ModalCloseButtonProps
...
      */
     "data-testid"?: string | undefined;
     /**
-     * Heading text rendered in the header. Always centered regardless of
+     * Heading rendered in the header. Accepts a plain string, or a node when
+     * you need to control the typography. Always centered regardless of
      * `closeButtonPosition`. Omit for a close-button-only header.
      */
-    title?: string | undefined;
+    title?: React.ReactNode | undefined;
```
qroll, line 60: Figma shows a standard heading xs now so the customisation is not needed

src/e-signature/e-signature.styles.ts
qroll: checking Figma, it is indeed meant to be shadow lg-strong. will create a separate ticket to align ModalV2

src/e-signature/e-signature.styles.ts
qroll: not in latest code, resolving

qroll (commented):
still somewhat confused as to why some new styles are needed

testing against the latest master, I'm able to strip down most of the styles https://github.com/LifeSG/react-design-system/tree/RL/CCUBE-2221, with the main discrepancy being the header spacing

PR #1450, opened 2026-08-31
Fix breakpoint detection on mobile with content overflow

qroll:
**Type of changes**

**Description of changes**

-   Use the ~~visual viewport~~ `window.matchMedia` api to calculate the actual screen width instead of `window.innerWidth`, which could grow beyond the visible viewport when there is overflow content (without overflow hidden)
- Original issue reported:
  - a responsive element was rendered: on md, it's a flex row, on sm, it's a flex column
  - on page load, it renders as the flex row as expected
  - but when resizing the window to sm, it remains as flex row instead of switching to flex column. on inspection, the breakpoint class remains on `fds-breakpoint-md`
  - this was demonstrated with Chrome's Responsive Viewport Mode

**Checklist**

-   [ ] ~~Added/updated unit tests~~ - can't be verified in unit tests
-   [x] Added/updated E2E tests

### DIFF
--- a/src/theme/theme-provider/breakpoint.ts
+++ b/src/theme/theme-provider/breakpoint.ts
@@ -57,7 +57,10 @@ const BREAKPOINT_CLASS_PREFIX = "fds-breakpoint-";
 
 function applyBreakpointClasses(sourceElement: HTMLElement) {
     const body = document.body;
-    const width = window.innerWidth;
+
+    const width = window.visualViewport
+        ? window.visualViewport.width * window.visualViewport.scale
+        : window.innerWidth;
 
     [...body.classList].forEach((cls) => {
         if (cls.startsWith(BREAKPOINT_CLASS_PREFIX)) {
--- a/src/theme/theme-provider/breakpoint.ts
+++ b/src/theme/theme-provider/breakpoint.ts
@@ -1,6 +1,6 @@
 import { Breakpoint } from "../tokens/breakpoint";
 import type { CSSVariableString } from "../types";
-import { parseCSSVariableValue, parsePxOrRemValue } from "../utils";
+import { parseCSSVariableValue } from "../utils";
 
 function getBreakpointRanges(sourceElement: HTMLElement) {
     const getBreakpointValue = (token: CSSVariableString) => {
@@ -58,10 +58,6 @@ const BREAKPOINT_CLASS_PREFIX = "fds-breakpoint-";
 function applyBreakpointClasses(sourceElement: HTMLElement) {
     const body = document.body;
 
-    const width = window.visualViewport
-        ? window.visualViewport.width * window.visualViewport.scale
-        : window.innerWidth;
-
     [...body.classList].forEach((cls) => {
         if (cls.startsWith(BREAKPOINT_CLASS_PREFIX)) {
             body.classList.remove(cls);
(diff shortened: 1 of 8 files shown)


qroll:
not including in e2e tests, as I'm not able to replicate the issue without the fix in the chromium browser. the browser continues to report `window.innerWidth` with the fixed value in `page.setViewportSize()`

for reference:

```
// e2e/nextjs-app/src/app/components/theme/media-query-content-overflow.e2e.tsx
"use client";

import { useMinWidthMediaQuery } from "@lifesg/react-design-system";
import { useEffect, useState } from "react";

import styles from "./media-query-specificity.module.css";

export default function Story() {
    const [isMounted, setIsMounted] = useState(false);
    const [breakpoint, setBreakpoint] = useState<string>("");
    const isSm = useMinWidthMediaQuery("sm");
    const isMd = useMinWidthMediaQuery("md");

    useEffect(() => {
        setIsMounted(true);
    }, []);

    useEffect(() => {
        const observer = new MutationObserver(() => {
            const body = document.body;
            const breakpointClass = [...body.classList].find((cls) =>
                /^fds-breakpoint-[a-z]+$/.test(cls)
            );
            if (breakpointClass) {
                setBreakpoint(breakpointClass);
            }
        });
        observer.observe(document.body, {
            attributes: true,
            attributeFilter: ["class"],
        });

        return () => {
            observer.disconnect();
        };
    }, []);

    return (
        <div className="story-column-container">
            {isMounted && (
                <>
                    <div>isSm: {isSm ? "true" : "false"}</div>
                    <div>isMd: {isMd ? "true" : "false"}</div>
                    <div>breakpoint: {breakpoint}</div>
                </>
            )}
            <div className={styles["overflow-content"]}>Overflow content</div>
        </div>
    );
}
```

```
// e2e/nextjs-app/src/app/components/theme/media-query-specificity.module.css
.overflow-content {
    width: 720px;
    padding: 16px;
    background: #BFDBFE;
}
```

```
// e2e/tests/components/theme/media-query.e2e.spec.ts
    test.describe(() => {
        test.beforeEach(async ({ story }) => {
            await story.init("media-query-content-overflow");
        });

        test("Viewport width calculation", async ({ story }) => {
            await story.page.setViewportSize({ width: 700, height: 720 });

            await story.page.waitForTimeout(1000);

            await story.page.setViewportSize({ width: 400, height: 720 });

            await story.page.waitForTimeout(1000);

            await expect(story.layout).toContainText("isSm: true");
            await expect(story.layout).toContainText("isMd: false");
            await expect(story.layout).toContainText(
                "breakpoint: fds-breakpoint-sm"
            );
        });
    });
```

src/theme/theme-provider/breakpoint.ts
```diff
@@ -57,7 +57,10 @@ const BREAKPOINT_CLASS_PREFIX = "fds-breakpoint-";
 
 function applyBreakpointClasses(sourceElement: HTMLElement) {
     const body = document.body;
-    const width = window.innerWidth;
+
+    const width = window.visualViewport
+        ? window.visualViewport.width * window.visualViewport.scale
```
ghazwan-gt, line 62: Should we do `Math.round` instead here?

ghazwan-gt (dismissed):

src/theme/theme-provider/breakpoint.ts
qroll: do you mean to round the final value of `width`? or to round `window.visualViewport`?

ghazwan-gt: I mean for the calculation result
```
Math.round(window.visualViewport.width * window.visualViewport.scale)
```
Just in case the value has fractional result e.g. x.999, could cause a 1px mismatch

qroll: got it, let me look into the rounding behaviour of media queries to remain consistent with our javascript checks

qroll: hmm yeah, there seems to be some rounding involved

**Chrome**



**Safari**
Screenshot 2026-09-02 at 11 40 08 PM

**Safari iOS** has some bizarre behaviour with the media query range syntax, but my sim is only on 26.2:

Screenshot 2026-09-02 at 11 50 11 PM

old [article](https://dev.to/sabberworm/what-to-do-about-sub-pixel-media-queries-6ic) for reference, which matches with the findings here

qroll: perhaps switching over to use `window.matchMedia` would give us more consistent results

ghazwan-gt (approved):

PR #1468, opened 2026-09-14
[MOL-22452] Remediate findings

thanhpt0105-gt:
**Type of changes**

-   [x] Other (technical improvements, refactoring, or changes that don't fall into the above categories)

**Description of changes**
 Link to [ticket] https://sgtechstack.atlassian.net/browse/MOL-22452
 Description of changes:
- Prevent GitHub Actions script injection in trigger-gitlab-pipeline.yml by moving github.event.head_commit.message and github.ref_name from direct ${{ }} interpolation to env: block
- Replace execSync with execFileSync in codemod CLI to prevent command injection via shell metacharacters in target path
- Sanitize href in withNotificationBanner link attributes using DOMPurify.isValidAttribute to block javascript: and data: URLs

**Checklist**

-   [ ] Updated documentation

**Screenshots**


.github/workflows/trigger-gitlab-pipeline.yml
```diff
@@ -14,10 +14,13 @@ jobs:
...
       - name: Print Configs
+        env:
+          HEAD_COMMIT_MSG: ${{ github.event.head_commit.message }}
+          REF_NAME: ${{ github.ref_name }}
         run: |
-          BRANCH_NAME="${{ github.ref_name }}"
+          BRANCH_NAME="$REF_NAME"
 
-          COMMIT_MSG=$(echo -e "${{ github.event.head_commit.message }}" | head -n 1)
+          COMMIT_MSG=$(printf '%s' "$HEAD_COMMIT_MSG" | head -n 1)
```
qroll, line 23: this doesn't seem to trim the second line because printf escapes the newline

instead of head, can we try [awk](https://stackoverflow.com/questions/43924863/awk-split-on-n) to grab only the first line?

src/notification-banner/notification-banner-hoc.tsx
```diff
@@ -9,6 +9,26 @@ import type {
...
 
+const SAFE_URL_SCHEMES = /^(https?:|mailto:|tel:)/i;
+
+function sanitizeAttributes(
+    attrs: Record<string, unknown>
+): Record<string, unknown> {
+    const sanitized: Record<string, unknown> = {};
+
+    for (const [key, value] of Object.entries(attrs)) {
+        if (/^on[A-Z]/i.test(key)) continue;
```
qroll, line 20: we can sanitize `href` on behalf of the consumer, but filtering out event handlers doesn't make sense here

as this is a generic component, there could be legitimate use cases where the consumer may need to specify custom `onClick`s. it's the consumer responsibility to avoid passing untrusted input here

qroll (changes requested):

src/notification-banner/notification-banner-hoc.tsx
```diff
@@ -9,6 +9,21 @@ import type {
     NotificationContentAttributes,
 } from "./types";
 
+function sanitizeLinkAttributes(
+    attrs: ContentLinkAttributes
+): ContentLinkAttributes {
+    const { href, ...rest } = attrs;
+
+    if (
+        typeof href === "string" &&
```
qroll: if `href` is not a string, it would bypass this condition

we should reject it as well

qroll (changes requested):

.github/workflows/trigger-gitlab-pipeline.yml
thanhpt0105-gt: Updated

src/notification-banner/notification-banner-hoc.tsx
thanhpt0105-gt: sanitize href only

src/notification-banner/notification-banner-hoc.tsx
thanhpt0105-gt: Updated to reject if href is not string.

qroll (approved):

